PR #1, opened 2025-07-29
add instruction to readme to run build before running example

kandros:
otherwise 

```
❯ bun run src/examples/index.ts
38 |   const targetLibPath = join(libDir, target, `${libraryName}.${suffix}`)
39 |   if (existsSync(targetLibPath)) {
40 |     return targetLibPath
41 |   }
42 | 
43 |   throw new Error(`Could not find opentui library for platform: ${target}`)
             ^
error: Could not find opentui library for platform: aarch64-macos
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ This creates platform-specific libraries in `src/zig/lib/` that are automaticall
 
 ## Examples
 
+Requires running a build script first. (`build`, `build:dev`, `build:prod`)
+
 ```bash
 bun run src/examples/index.ts
 ```


kommander:
Hey, thanks for your interest! There is a build section directly above the example section though? I think more focus will be put on documentation as soon as we have something semi stable, then I see the build docs more with developer docs as the lib package should then work out of the box.

ZiedYousfi (changes requested):
Maybe just a 

```bash
bun run build && bun run src/examples/index.ts
```

PR #5, opened 2025-07-30
Add Prebuilt Binaries

lazuee:
Publish prebuilt binaries instead of building it manually

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,13 +1,14 @@
 {
   "name": "@opentui/core",
   "module": "src/index.ts",
-  "main": "src/index.js",
+  "main": "dist/index.js",
   "version": "0.1.0",
   "type": "module",
   "scripts": {
-    "build": "cd src/zig && zig build",
-    "build:dev": "cd src/zig && zig build -Doptimize=Debug",
-    "build:prod": "cd src/zig && zig build -Doptimize=ReleaseFast",
+    "build": "bun run build:lib && bun run build:native",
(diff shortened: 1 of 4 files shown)


kommander:
Hey! Thanks for your eagerness to contribute. I wonder why this needs to generate additional package.json files? Can you explain how this provides the prebuild libs?

The publishing I would put in a GitHub workflow and trigger with release tags.

I would take a similar route as I did with https://github.com/kommander/bun-webgpu, while this build would be simpler though.

lazuee:
@kommander 

`build.mjs --native (?:--dev)` - compiles the Zig libraries and creates a separate `package.json` in each platform-arch folder and let npm treat each prebuilt binary as its own (optional) package and only install the one matching your platform-arch.

`build.mjs --lib` - bundles the main library into `dist/` and generates a `package.json` there will add native-binary packages as `optionalDependencies`, so only the packages matching the user platform-arch will be installed.

`publish.mjs` - prompts if the main package version was bumped and if there are any new changes pushed. It then uses the `dist/` as the main package and the `node_modules/@opentui/core-[platform]-[arch]` folders as the prebuilt binaries. Before publishing, it checks that the package versions match to avoid mismatches, then proceeds to publish all packages.

kommander:
Ohhhh I totally misread that. That makes sense. @thdxr do we want to publish some alpha versions already? I still would opt for publishing in a workflow.

kommander:
Took this into #10, thanks!

PR #7, opened 2025-07-31
refactor(renderer): Separate render loop from single frame    rendering

SyedaAnshrahGillani:
This pull request refactors the CliRenderer to improve code clarity and
  maintainability.

  The core rendering logic has been extracted from the loop() method into a
  new private method called _renderFrame().

  This change allows renderOnce() to directly call the rendering logic without
  invoking the entire loop, which is more efficient and semantically correct.
  The loop() method now focuses on its primary responsibility: managing the
  rendering loop and scheduling frames.

  This refactoring improves the overall structure of the CliRenderer and makes
  the code easier to understand and maintain. No functional changes are
  introduced.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -745,7 +745,9 @@ export class CliRenderer extends Renderable {
     const frameRequests = this.animationRequest.values()
     this.animationRequest.clear()
     const animationRequestStart = performance.now()
-    frameRequests.forEach((callback) => callback(deltaTime))
+    for (const callback of frameRequests) {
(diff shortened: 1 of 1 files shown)


kommander:
Hey, thanks for contributing! I see your concerns. I need to test this. 

For me the loop() method is basically the render method, just running one cycle. Extracting that would still need to take care of the deltaTime though, should not just be 0.

simonklee:
stale

PR #8, opened 2025-08-04
feat(🫐): Use TypeGPU to strengthen the I/O layer between the CPU and GPU (TGSL Version)

iwoplaza:
This is an updated version of the [original PR](https://github.com/sst/opentui/pull/6), but instead of functions being implemented in WGSL, they're implemented in TypeScript. This allows type-safety to flow from host code directly to shader code.

The inclusion of `unplugin-typegpu` allows the implementations of `tgpu.fn` functions to be compiled to WGSL at bundle time


nikelborm:
Oh, Wow! That's exactly as I would have intuitively expected such an API to look like! 😉

nikelborm:
Were any of the tools I mentioned in the previous PR helpful in the process of creating this implementation?

iwoplaza:
> Were any of the tools I mentioned in the previous PR helpful in the process of creating this implementation?

Not this particular one, as we've already had plugins for other bundlers, so porting it over was relatively straightforward. The projects you mentioned are going to be really helpful when tackling the WGSL-literal inference and linting.

PR #9, opened 2025-08-04
docs: "Caustics Shader" and "Lava Lamp Shader" demos

iwoplaza:
These examples showcase how to use OpenTUI & TypeGPU

https://github.com/user-attachments/assets/2cae73e3-a1bb-421a-8ad9-e6571874a308
https://github.com/user-attachments/assets/cab29ffe-4216-4013-89cb-ab0a31b17374

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -2,11 +2,14 @@
   "lockfileVersion": 1,
   "workspaces": {
     "": {
-      "name": "opentui",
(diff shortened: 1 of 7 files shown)


simonklee:
Thanks for the suggestion, but we prefer not adding that dependency for an example.

PR #17, opened 2025-08-15
Add insert before function

fezproof:
Remove some duplicate logic to replace parents
Add insert before, follow web convention

You can now call
```typescript
parent.insertBefore(childObj, anchorObj);

// OR

parent.insertBefore(childObj); // this will act just like parent.add(childObj)
```

### DIFF
--- a/src/Renderable.ts
+++ b/src/Renderable.ts
@@ -587,19 +587,21 @@ export abstract class Renderable extends EventEmitter {
     this.needsUpdate()
   }
 
-  public add(obj: Renderable): void {
+  private replaceParent(obj: Renderable) {
     if (this.renderableMap.has(obj.id)) {
       this.remove(obj.id)
     }
-
     if (obj.parent) {
       obj.parent.remove(obj.id)
     }
-
     obj.parent = this
     if (this.ctx) {
       obj.propagateContext(this.ctx)
     }
+  }
+
+  public add(obj: Renderable): void {
+    this.replaceParent(obj)
 
     this.renderableArray.push(obj)
     this.needsZIndexSort = true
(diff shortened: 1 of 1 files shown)


Adictya:
Hey @fezproof looks like we are working on the same thing. I was also porting over my solid implementation from https://github.com/Adictya/mold to open tui.

And I ended up implementing the same thing but with a slightly different api here: https://github.com/sst/opentui/pull/18

Imo having an index keeps it more open-ended. What do you think?

kommander:
As the child index is not obvious we can have both in the API? Imo insertBefore is nicely browser API "compatible".

src/Renderable.ts
```diff
@@ -612,6 +614,33 @@ export abstract class Renderable extends EventEmitter {
...
+    if (anchorIndex === -1) {
+      throw new Error("Anchor does not exist")
+    }
+
+    this.replaceParent(obj)
+
+    this.renderableArray.splice(anchorIndex, 0, obj)
+    this.renderableMap.set(obj.id, obj)
+    const childLayoutNode = obj.getLayoutNode()
+    this.layoutNode.insertChild(childLayoutNode, anchorIndex)
```
kommander, line 637: I like the replaceParent method. Maybe this can just use @Adictya's `add(renderable, index?)` method?

Adictya:
We can merge this then if everything looks good, Ill modify my pr by pulling the changes.

src/Renderable.ts
```diff
@@ -612,6 +614,33 @@ export abstract class Renderable extends EventEmitter {
...
+
+    this.needsZIndexSort = true
+    this.requestLayout()
+    this.emit("child:added", obj)
```
kommander, line 641: btw. do these events make sense to anyone, can't remember why I added these.

fezproof:
Awesome glad it all worked!

@Adictya I have a simple working version of a solidjs terminal with bun, but I am in a much earlier stage than you.

I will be looking on how you did HMR with a bun server as that will be a big win. Happy to chat/collaberate if you are interested. You can find me on the SST discord

Adictya:
> I will be looking on how you did HMR with a bun server as that will be a big win

@fezproof Yeah, unfortunately the ecosystem isn't there yet I faced issues with both bun and vite-node. But got a working version with vite-node finally by working around some bugs. Will be raising prs to address these issues, after which it should work pretty much out of the box.

PR #18, opened 2025-08-15
Core changes for Solid reconciler implementation

Adictya:
Porting my solidjs implementation from https://github.com/Adictya/mold onto Opentui with HMR

### DIFF
--- a/src/renderables/ASCIIFont.ts
+++ b/src/renderables/ASCIIFont.ts
@@ -28,13 +28,13 @@ export class ASCIIFontRenderable extends FrameBufferRenderable {
 
   constructor(id: string, options: ASCIIFontOptions) {
     const font = options.font || "tiny"
-    const text = options.text || "known"
+    const text = options.text || ""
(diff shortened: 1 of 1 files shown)


src/renderables/ASCIIFont.ts
```diff
@@ -28,7 +28,8 @@ export class ASCIIFontRenderable extends FrameBufferRenderable {
 
   constructor(id: string, options: ASCIIFontOptions) {
     const font = options.font || "tiny"
-    const measurements = measureText({ text: options.text, font })
+    const text = options.text || "known"
```
kommander: Rather empty string then?

Adictya: Causes a crash if its just an empty string

kommander: If it needs to be optional, call `super` with this:
```ts
      width: measurements.width || 1,
      height: measurements.height || 1,
```
Need to work out how the FrameBufferRenderable should work best.

PR #23, opened 2025-08-15
build libs dynamic and fail build if no libs copied

kommander:


Vaelatern:
I was hoping for more context when I found this line. This breaks the resulting binaries on musl: https://github.com/sst/opencode/issues/3876

PR #27, opened 2025-08-15
chore(box): update bg prop to backgroundColor

msmps:
## Description

- bring box `bg` color property inline with other renderables -> `backgroundColor`


kommander (approved):
👍

PR #31, opened 2025-08-16
Add solid reconciler with example

Adictya:
```
bun install
cd packages/solid
bun --conditions=browser run examples/index.tsx
```

### DIFF
--- a/packages/core/src/renderables/Text.ts
+++ b/packages/core/src/renderables/Text.ts
@@ -68,9 +68,6 @@ export class TextRenderable extends Renderable {
   set content(value: StyledText | string) {
     this._text = typeof value === "string" ? stringToStyledText(value) : value
     this.updateTextInfo()
-    this.needsUpdate()
-    this.syncSelectionToTextBuffer()
-    this.needsUpdate()
   }
 
   get fg(): RGBA {
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -7,7 +7,7 @@
   },
   "types": "src/index.ts",
   "module": "src/index.ts",
-  "main": "dist/index.js",
+  "main": "src/index.ts",
   "version": "0.1.5",
   "type": "module",
   "scripts": {
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -451,8 +451,8 @@ export abstract class Renderable extends EventEmitter {
     }
 
     this._positionType = options.position ?? "relative"
-    if (this._positionType === "absolute") {
-      node.setPositionType(PositionType.Absolute)
+    if (this._positionType !== "relative") {
+      node.setPositionType(parsePositionType(this._positionType))
     }
 
     // TODO: flatten position properties internally as well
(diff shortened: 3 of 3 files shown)


packages/solid/src/elements/index.ts
```diff
@@ -0,0 +1,142 @@
...
+      }
+    }
+
+    return element as JSX.Element;
+  };
+};
+
+type ContianerProps = { children?: JSX.Element };
+
+export const Box = createCustomElement<ElementProps<BoxOptions, BoxRenderable> & ContianerProps>("opentui_box", true);
```
fezproof: Is there a reason to make custom capitalised elements as opposed to just having `<box>stuff</box>`?
Having to have to specifically import custom elements from the renderer doesn't really feel natural when you don't really need to. This is one of the biggest limitations of react ink, and we shouldn't need to make the same choice

Adictya: Hmm, yeah hadn't considered. No reason they have to be will try and see

Adictya:
@fezproof  Gave the importless thing a quick stab, seems to work fine with one key drawback being the Text and AsciiFont flows. I don't see an easy way of being able to pass text content as children to them without significant reworks to core. Maybe you can take a look?

Personally I am not too opinionated on this. Imports are automatic and painless in most IDE's and Solid itself does not shy away in recommending Importable components like Show etc.

Curious to hear @kommander 's opinion on this.

kommander:
@Adictya I actually don't know anything about the importless stuff, but what rework would you need in the core to render text as children?

packages/solid/src/utils/id-counter.ts
```diff
@@ -0,0 +1,11 @@
+const idCounter = new Map<string, number>();
```
fezproof, line 1: Do we need this? Can we just use the createUniqueId provided by solid?

packages/core/src/renderables/Text.ts
```diff
@@ -68,7 +68,8 @@ export class TextRenderable extends Renderable {
   set content(value: StyledText | string) {
     this._text = typeof value === "string" ? stringToStyledText(value) : value
     this.updateTextInfo()
-    this.setupMeasureFunc()
+    this.needsUpdate()
```
fezproof:
```suggestion
```
Don’t need double call of needsUpdate

packages/solid/src/elements/index.ts
```diff
@@ -0,0 +1,119 @@
...
+    interface IntrinsicElements {
+      ascii_font: ASCIIFontElementProps;
+      box: BoxElementProps;
+      group: GroupElementProps;
+      input: InputElementProps;
+      select: SelectElementProps;
+      tab_select: TabSelectElementProps;
+      text: TextElementProps;
+    }
+  }
```
fezproof, line 48: This allows us to still have something like `<div>` in the markup, and typescript to be fine with it. I will make a follow up when this is merged to fix this.

fezproof:
Yep all looks good to me otherwise! I wonder if there is a more a simple way to do the text chunk rendering, will have a more of a play.
It's so cool to see that things like Suspense and For just work out of the box though

PR #32, opened 2025-08-17
StyledText chunk mutation apis

Adictya:


packages/core/src/Renderable.ts
```diff
@@ -525,6 +525,19 @@ export abstract class Renderable extends EventEmitter {
...
 
+  set position(positionType: PositionTypeString) {
+    if (this._positionType === positionType) return
+
+    this._positionType = positionType
+    if (this._positionType === "relative") {
+      this.layoutNode.yogaNode.setPositionType(PositionType.Absolute)
+    } else {
+      this.layoutNode.yogaNode.setPositionType(PositionType.Absolute)
+    }
```
msmps, line 536: is this correct?

PR #34, opened 2025-08-17
feat(react): initial commit

msmps:
# TODO:

- enable text to take styled-text children over content
- impl `style` property
- impl ascii-text

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -33,6 +33,23 @@
         "typescript": "^5",
       },
     },
+    "packages/react": {
+      "name": "@opentui/react",
+      "version": "0.1.1",
+      "dependencies": {
+        "@opentui/core": "workspace:*",
+        "react-reconciler": "^0.32.0",
+      },
+      "devDependencies": {
+        "@types/bun": "latest",
+        "@types/react-reconciler": "^0.32.0",
+        "prettier": "3.6.2",
+      },
+      "peerDependencies": {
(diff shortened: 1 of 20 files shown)


bhushan6:
Just tried this locally and it works really nice! Though I wanted to understand, is there a reason to create wrapper components for primitive elements that need to be imported?
One issue I see with that approach is maintenance overhead. if we add new primitives in the core package, we need to manually add support in each renderer. This creates a tight coupling and potential for the React renderer to lag behind core updates.
Another key consideration is cross-renderer consistency, How do we ensure extended components work across all renderers? Looking at React Three Fiber, they provide an extend() function that accepts new classes and automatically adds them to the component registry. Then you can use them directly in JSX like any other element.
This pattern would be crucial for our multi-renderer strategy, with the Solid.js renderer in progress, users should ideally write a component once and use it across all environments (vanilla JS, React, Solid). Something like:

```// Define once
class ConsoleButton extends BoxRenderable { /* ... */ }

// Extend in any renderer
extend({ consoleButton: ConsoleButton })

// Use naturally in each syntax
// React: <consoleButton label="Click" onClick={handler} />
// Solid: <consoleButton label="Click" onClick={handler} />  
// Vanilla: new ConsoleButton("log-btn", 2, startY, buttonWidth, buttonHeight, logColor, "LOG", "log")```

PR #37, opened 2025-08-18
Enables loading fonts from external files

entrepeneur4lyf:
Adds functionality to load and register fonts from external JSON files.

This allows users to define and use custom fonts beyond the built-in options, improving flexibility and customization.


kommander:
Hi, thanks for contributing! That makes sense. Could we avoid the `font | FontDefinition` everywhere and make it just FontDefinition?

eas4ai:
Yes, I'll remediate here in a bit

kommander:
What happened? Would be nice to support custom ascii fonts.

eas4ai:
Oh I thought it had been fixed already lmao. No worries, I'll submit a different PR with the fix

PR #56, opened 2025-08-23
fix: convert MapIterator to array for frameRequests.forEach

hanbao-dev:
The animationRequest.values() method returns a MapIterator which doesn't have a forEach method, causing "frameRequests.forEach is not a function" runtime error when requestAnimationFrame callbacks are processed.  Wrap with Array.from() to convert the iterator to an array before calling forEach in the render loop.  Fixes animation frame callback processing in Node.js environments.


kommander:
How are you running this on node?

hanbao-dev:
> How are you running this on node?

I was getting the error when running `bun run src/examples/index.ts`

Issue #41, opened 2025-08-19
Bun v1.2.13: crash in `@opentui/core/3d` when running Sprite Animation example because `console.Console` is undefined


Montekkundan:
Running the Sprite Animation example on Bun throws `TypeError: undefined is not a constructor` where `@opentui/core/3d.js` constructs new `console.Console(...)` and overwrites `global.console`.

**Env**

* Bun v1.2.13 macOS arm64
* `@opentui/core` 0.1.6

**Repro**

```bash
bun init -y
bun add @opentui/core three
```

`index.ts`:

```ts
// run the Sprite Animation example
```

```bash
bun run index.ts
```

**Error**

```
TypeError: undefined is not a constructor (new console.Console({...}))
at node_modules/@opentui/core/3d.js:36286:20
```

kommander:
Is this in a new and empty folder with `bun init`?  What does `// run the Sprite Animation example` do exactly here?

Montekkundan:
Yes, https://github.com/Montekkundan/opentui-projects

running the same https://github.com/sst/opentui/blob/main/packages/core/src/examples/sprite-animation-demo.ts file with revised imports

kommander:
The issue is that the 3d entry point bundles all of core and executes certain sections twice, which shouldn't happen. Also, threejs >v0.177.0 is not yet supported. I fixed the entry point and pinned the threejs version in the optional dependencies. Will be in the v0.1.8 release coming shortly.

gkurt:
This also happens on any code when running with Bun hot reloading and making changes to files. Is there a plan to support Bun hot reloading?

kommander:
@KurtGokhan Not on the roadmap yet, but will be. The react/solid wrappers will support HMR though.

gkurt:
Thanks. I created #57 to prevent the crash caused by `undefined`.

I also tried to make a POC for hot reload, but for some reason `import.meta.hot` isn't available in Bun even though [the docs](https://bun.com/docs/bundler/hmr) imply it is. Will try to follow-up later.

msmps:
> Thanks. I created #57 to prevent the crash caused by `undefined`.
> 
> I also tried to make a POC for hot reload, but for some reason `import.meta.hot` isn't available in Bun even though [the docs](https://bun.com/docs/bundler/hmr) imply it is. Will try to follow-up later.

[...]

PR #57, opened 2025-08-23
fix: console constructor undefined on rerun

KurtGokhan:
Fixes `new console.Console` causing crash as can be seen in #41. This can be reproduced in various ways, but I mainly experience it when running with Bun hot reload (`bun run --hot ....`).

This PR doesn't aim to fully support hot reload though. Notably, the old TUI isn't destroyed on hot reload, so there will be glitches and memory leaks. But at least this prevents a crash. User can rerun the bun command or use `--watch` if they want to clear the old instance.

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -90,8 +90,6 @@ export interface RenderableOptions extends Partial<LayoutOptions> {
   onKeyDown?: (key: ParsedKey) => void
 }
 
-let renderableNumber = 1
-
 function validateOptions(id: string, options: RenderableOptions): void {
   if (typeof options.width === "number") {
     if (options.width < 0) {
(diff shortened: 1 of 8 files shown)


kommander:
Alright, that can have other side effects though, we need to make sure it is not initialized more than once. There are some other globals that should only be initialized once, like the "console cache instance" and the process signal listeners in renderer.ts

I am happy to support that.

gkurt:
I made some changes in that direction, but they may be opinionated. Feel free to override or discard.

To be clear, there are more issues and side effects that may happen when rerunning the code. For example, `CliRenderer` modifies `global` in its constructor. Those weren't things I was planning to tackle in this PR, so I won't touch them for now. Seems like it needs a more in-depth architectural change. It may even be unnecessary when Bun adds a proper API for `import.meta.hot`.

kommander:
Awesome! Yeah, the global stuff in the `CliRenderer` is meh. Wasn't sure how to handle that, because theoretically you can have more than one renderer to serve ssh interfaces for example, that could be done via threads though as well to isolate the renderer. In most cases you never need more than one renderer instance. The `requestAnimationFrame` only makes sense for one renderer in the same process though. Happy if you have any ideas around that.

Tested this. Going to merge it if you don't want to add anything more?

gkurt:
Nothing more to add from my side for now. I will use OpenTUI more and observe real-life usage before coming up with changes.

PR #61, opened 2025-08-24
fix: update cursor position after layout change

KurtGokhan:
I noticed cursor position doesn't update after the position of the input element changes. This is noticable especially when resizing the terminal.

Before:

https://github.com/user-attachments/assets/067bc60f-451b-4722-84f7-9afe48013507

After:

https://github.com/user-attachments/assets/d1b82852-340f-482e-9c45-6eb128c29e09

- Also, I fumbled the prettier formatting in my previous PR. Fixed them in the second commit.

### DIFF
--- a/packages/core/src/renderables/Input.ts
+++ b/packages/core/src/renderables/Input.ts
@@ -294,6 +294,16 @@ export class InputRenderable extends Renderable {
     this.needsUpdate()
   }
 
+  public updateFromLayout(): void {
+    super.updateFromLayout()
+    this.updateCursorPosition()
+  }
+
+  protected onResize(width: number, height: number): void {
+    super.onResize(width, height)
+    this.updateCursorPosition()
+  }
+
   protected destroySelf(): void {
     if (this._focused) {
       CliRenderer.setCursorPosition(0, 0, false)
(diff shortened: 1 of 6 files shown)


packages/core/src/renderables/Input.ts
```diff
@@ -294,6 +294,16 @@ export class InputRenderable extends Renderable {
...
 
+  public updateFromLayout(): void {
+    super.updateFromLayout()
+    this.updateCursorPosition()
+  }
+
+  protected onResize(width: number, height: number): void {
+    super.onResize(width, height)
+    this.updateCursorPosition()
+  }
```
gkurt, line 305: This is the only significant part of this PR. Other changes are prettier changes.

I could check if the layout has changed before calling `updateCursorPosition` to avoid unnecessary calls, but I couldn't find a reliable way to do it. Not sure about the performance impact of calling `updateCursorPosition`.

kommander: onResize is only called when the actual size of the renderable in the layout changes. We could add something similar in the base Renderable like `protected onLayoutChanged` that can be overridden by subclasses, just like onResize. And could be called by `updateFromLayout` when either size OR position change. Wdyt? This works as well though. `updateFromLayout` is not supposed to be called by subclasses, even though performance impact would be negligible. Would maybe prefix that with an underscore. I don't see a better way to handle tree update methods atm.

kommander:
I'll take this in and we can refactor the API usage later.

PR #62, opened 2025-08-24
Clean up TextNode renderer

fezproof:
Move all text node logic to its own file, and take advantage of classes.


Adictya (approved):
Niice!

PR #73, opened 2025-08-25
fix: fallback to default values when properties undefined

msmps:
## Description

- this largely covers color properties as others either respect undefined when passed or undefined is absurd 
- sets `border` to false by default on `box` renderable


kommander (approved):
If the examples still work for react and solid as well, I think this is good to go 👍

kommander (approved):
LGTM

Issue #76, opened 2025-08-25
[Proposal] Add utility function to measure height and width of children


ReverseGem7:
## Summary
Currently, there is no straightforward way to determine the height and width of a rendered child element in the terminal renderer. The only way to get it is to calculate the size of each component manually using `measureText` or `stringWidth`. This proposal suggests adding a utility function that calculates these dimensions automatically based on content and wrapping rules.

## Motivation
Knowing the exact size of a child element is necessary for:

- Implementing scroll areas correctly.
- Dynamic calculations.
- Reducing error-prone manual measurements, especially for complex layouts.

kommander:
I want to provide a ScrollContainer renderable for exactly that. Yoga nodes can have an `overflow` set to `scroll`, but only in the main axis direction of the container (column/row), so for scroll in both directions one needs a nested container setup and handle that. With overflow set, yoga provides the content size by letting the first level container grow it's layout naturally across the main axis as if there were no boundaries. So we can take the sub container dimensions and handle a scrollOffset in the scrollable root and set x/y accordingly. Combining that with a framebuffer that only has the clipping dimensions (buffered=true in scroll root), that will only draw elements that are within the clipping area. As buffer draws already check bounds, it will just not draw elements that are out of bounds and performance should come with it automatically.

msmps:
As a side note for the interim - you can grab the dimensions by using refs in solid/react and directly in core

<img width="2380" height="2294" alt="Image" src="https://github.com/user-attachments/assets/6d87eecf-d118-4ffb-8cae-46e19efddd5b" />

ReverseGem7:
> As a side note for the interim - you can grab the dimensions by using refs in solid/react and directly in core
> 
[...]

PR #78, opened 2025-08-25
feat: scroll box component

Leka74:
I wanted to have some sort of a box component which the content can be scrollable. This maybe could've also be done as part of the `box` component as well.

Works similarly to the box component, with the addition of `showScrollIndicator` and `scrollStep` properties. You'll have to add a `height` to the component to get the main benefit of this. Maybe that property has to be required if this ends up being a separate component and not just an update of `box`

fyi: part of this has been vibe coded.

https://github.com/user-attachments/assets/3cfb9e99-69ca-43a4-bacf-844390b40ed1


kommander:
Ah nice! I was just answering about that in #76. I'll check it out asap. edit: would make it a separate ScrollContainer renderable in core. Edit 2: naming it scrollbox is also fine. Need to look at this more closely first.

ReverseGem7:
It would be good to provide a more browser-like experience, adding:

* **Thumb dragging** – allow users to grab and drag the scrollbar thumb.
* **Track clicking** – allow users to click on the track (the path behind the thumb) to jump through content.

### API

```ts
type TrackClickBehavior = "jump" | "smooth";

interface ScrollAreaProps {
  axis?: ("horizontal" | "vertical")[];  // default: ["vertical"]
  trackClickBehavior?: Partial<Record<"horizontal" | "vertical", TrackClickBehavior>>;
  dragThreshold?: number;                // px to avoid accidental drags
  scrollSmoothMs?: number;               // duration if "smooth"
  onUserScroll?: (e: {
    source: "wheel" | "key" | "drag" | "track";
    axis: "x" | "y";
    offset: number;
  }) => void;
}

<ScrollArea
  axis={["vertical", "horizontal"]}
  trackClickBehavior={{
    vertical: "jump",    // default
    horizontal: "smooth"
  }}
  dragThreshold={2}
  scrollSmoothMs={120}
  onUserScroll={(e) => {
    console.log(e.axis, e.source, e.offset);
  }}
/>
```

### Demo snippet (concept using `@opentui/react`)

https://github.com/user-attachments/assets/e64b87ba-739d-4523-87e5-223b484085ce

[Code](https://gist.github.com/ReverseGem7/54f61f274c42ec691eaf9874b24a42b3)

kommander:
So, how do I bring you two together to solve this? Are you both on discord on the opencode server?

ReverseGem7:
@kommander I just joined
Username: reversegem_7

kommander:
Was solved in #120 and #125

PR #80, opened 2025-08-25
Basic CI setup

Adictya:


kommander (approved):

kommander:
I would not skip any build steps in CI though, it can go through with everything for the build, we want to know that everything works. Build in CI should build completely.

PR #86, opened 2025-08-26
Migrate to zig 0.15.1

nyxkrage:
Also includes a small change to the build script which allows setting the Zig compiler with the ZIG environment variable, as this makes it much easier than needing to change your PATH back and forth when testing different versions of zig.

Migrates:  
std.fmt.format -> std.Io.Writer.print  
std.ArrayList -> std.ArrayListUnmanaged (while std.array_list.Managed does exist and should be the same as the old ArrayList, the managed options will be removed in the future, so I thought might as well bite the bullet and use the Unmanaged versions as is recommended)  
CliRenderer.addStatSample to a method instead of a function so that it can use self.allocator for the unmanaged arraylist it now uses  
Removes the use of bufferedWriter and now just stores the buffer and constructs the stdout writer as needed.  

One consideration is the use of the std.io.GenericWriter in the CliRenderer.writer return type, this is technically deprecated, but as far as I can tell, the other options requires a much larger refactoring, so I have marked it as a TODO, however you may want to just hold out on moving to Zig 0.15 and migrate it properly at the same time as upgrading the toolchain version instead.

With these changes I also seem to get faster speeds from the src/benchmark/renderer-benchmark.ts, though unsure why that is or whether it is a fluke.  
Before:
image
After:
image

### DIFF
--- a/packages/core/scripts/build.ts
+++ b/packages/core/scripts/build.ts
@@ -78,11 +78,13 @@ if (missingRequired.length > 0) {
   process.exit(1)
 }
 
+const zigExe = process.env.ZIG ?? "zig"
+
 if (buildNative) {
   console.log(`Building native ${isDev ? "dev" : "prod"} binaries...`)
 
   const zigBuild: SpawnSyncReturns<Buffer> = spawnSync(
-    "zig",
+    zigExe,
     ["build", `-Doptimize=${isDev ? "Debug" : "ReleaseFast"}`],
     {
       cwd: join(rootDir, "src", "zig"),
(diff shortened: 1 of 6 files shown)


kommander:
This great! Thanks. The benchmark numbers fluctuate very much, would have to run it a hundred times each and median or avg over that. Need a better benchmark, this is from the very early days.

I am currently reworking a lot of the native parts to support full Unicode. I'll try to rebase on yours and maybe bring full Unicode in with 0.15. But if it gets too tedious I might have to merge my changes first and then migrate to 0.15.

kommander:
I ran into some issues with 0.15.1 and the changes are huge already. I'll merge my stuff first 😬

msmps:
closed by #439

PR #94, opened 2025-08-27
fix: ensure cursor color reset in some terminals

KurtGokhan:
The cursor color wasn't resetting when program is terminated while an input is focused. As I see `112` is the correct ANSI code for resetting cursor color, but I also kept the old code to be sure.

https://github.com/user-attachments/assets/2126aeab-e503-417c-9d67-52f1359e6ffb


gkurt:
Yay, I can add Zig to my resume now 😂

Issue #75, opened 2025-08-25
Support alt-left and alt-shift-left keys to move cursor and selection by one word at a time


remorses:
This feature makes it much easier to move around text, Claude Code already supports it 

Also `alt-delete` to delete a word

### Shortcuts to implement (right arrow variants omitted)

- [ ] **Alt + Left** — Move cursor to start of previous word (skip adjacent spaces + first punctuation).
- [ ] **Alt + Shift + Left** — Expand selection to start of previous word.
- [ ] **Alt + Delete** — Delete the word to the left of the cursor.

If we manage to get cmd key somehow:
- [ ] **Cmd + Left** — Move cursor to start of the current line.
- [ ] **Cmd + Shift + Left** — Select from cursor to start of the current line.
- [ ] **Cmd + Delete** — Delete from cursor back to the start of the current line.

kommander:
Readline movements like per word with alt+left/right is supported by default in the new Textarea. The Input renderable will be deprecated and removed. Cmd key only works with kitty and an app can manually override the bindings for the Textarea on app level when kitty is supported imo.

remorses:
awesome

PR #95, opened 2025-08-28
Add word jumping utils mimicking VSCode alt-arrow shortcuts

remorses:
In preparation of #75 

I [already implemented this feature into Zed](https://github.com/zed-industries/zed/pull/31977) so I am adding it here too, I already spent a lot of time to understand all the nuances

### DIFF
--- a/packages/core/src/lib/word-jumps.test.ts
+++ b/packages/core/src/lib/word-jumps.test.ts
@@ -71,7 +71,7 @@ describe("word-jumping", () => {
       const caret = text.indexOf("|")
       const cleanText = text.replace("|", "")
       const result = nextWordEndCrossLines(cleanText, caret)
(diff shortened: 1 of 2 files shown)


kommander:
Nice, is that meant for the text input box?

remorses:
Yes, it could also be used for moving selection with arrow keys

kommander:
Got it. Just a heads up: I'll be moving over my svim neovim clone stuff into opentui. That EditBuffer will become the base for a textarea, with optional vim like editing behaviour. It might also become the base for Input and it already support w/W motions to jump by word in visual mode.

I would focus on the selection part, would be awesome if we could control that via keys. The whole text selection I want to move to native at some point, but this shouldn't be a blocker either way.

remorses:
I think you can reuse these functions also to implement previous and next word in vim mode. Zed uses the same logic as alt-arrow shortcuts for their Vim mode.

kommander:
Yes, I think I can bring that together.

PR #97, opened 2025-08-28
feat(core): add support for gap properties

### LINKED ISSUE #89: Add support for flexbox `gap` in group and box (React)
Gap makes it easier to add spacing between elements contained in a group or box instead of using margin

I think Yoga should already support it

msmps:
## Description

- add support for `gap`, `rowGap`, and `columnGap` properties
- omit properties from "non container" renderables
- closes #89 

## Snapshot

CleanShot 2025-08-28 at 17 49 14@2x


packages/core/src/Renderable.ts
```diff
@@ -165,6 +168,16 @@ export function isSizeType(value: any): value is number | `${number}%` | undefin
   return isValidPercentage(value)
 }
 
+export function isGapType(value: any): value is number | undefined {
```
msmps: ❓ this is a direct clone of `isSizeType` but didn't know if we wanted to be explicit in it's usage

kommander:
Now that I see this... If we omit from everything but the `Box` (group will go away and Framebuffer renderable is not supposed to be a container), then we can just have the `Box` handle it only, like it does already with the borders?

Regarding the `FramebufferRenderable`: It is supposed to be a container for a custom frame buffer that was created externally and should be given to the `FramebufferRenderable` at construction time. It currently collides with the `buffered` attribute. Renderables that want their stuff buffered can use the base `buffered` attribute.

kommander (approved):
excellent 🤌

remorses:
Thank you this is great!

PR #98, opened 2025-08-28
fix(react): update types

msmps:


kommander (approved):
I might have missed that in solid as well 😔

PR #105, opened 2025-08-29
add a release workflow for the dylib; create install.sh; move the C header to go/ for now, cgo doesnt support looking in parent dirs

dnakov:
It needs proper packaging like brew and stuff, but we can at least have an install.sh script for now


kommander:
Can you run prettier?

dnakov:
i didnt touch any typescript?

kommander:
But the READMEs. Also can you make the install.sh use opentui?

kommander:
jk with opentui for the install script, but fix it and I merge.

dnakov:
why is prettier running on .md files *smh* anyway ran it :)

dnakov:
@kommander do a tag to see if release works

PR #108, opened 2025-08-30
Fix reactive component conditional rendering issue

Adictya:
- Also fixes a small bug with empty text renderables

### DIFF
--- a/packages/solid/src/reconciler.ts
+++ b/packages/solid/src/reconciler.ts
@@ -35,7 +35,15 @@ const logId = (node?: DomNode): string | undefined => {
 }
 
 function _insertNode(parent: DomNode, node: DomNode, anchor?: DomNode): void {
-  log("Inserting node:", logId(node), "into parent:", logId(parent), "with anchor:", logId(anchor), node instanceof TextNode)
+  log(
+    "Inserting node:",
+    logId(node),
+    "into parent:",
+    logId(parent),
+    "with anchor:",
+    logId(anchor),
+    node instanceof TextNode,
+  )
 
   if (node instanceof StyledText) {
     log("Inserting styled text:", node.toString())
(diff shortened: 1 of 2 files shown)


kommander (dismissed):

packages/solid/src/reconciler.ts
```diff
@@ -13,17 +13,55 @@ import {
...
+  }
+
+  if (isTextChunk(node)) {
+    // @ts-expect-error: Sending chunk to createTextNode which is not typed but supported
+    insertNode(parent, createTextNode(node), anchor)
+    return
+  }
 
   if (node instanceof TextNode) {
+    debugger
```
fezproof: Cleanup

packages/solid/src/reconciler.ts
```diff
@@ -13,17 +13,55 @@ import {
...
+    "with anchor:",
+    logId(anchor),
+    node instanceof TextNode,
+  )
+
+  if (node instanceof StyledText) {
+    log("Inserting styled text:", node.toString())
+    for (const chunk of node.chunks) {
+      // @ts-expect-error: Sending chunk to createTextNode which is not typed but supported, but we know it's a chunk
+      insertNode(parent, createTextNode(chunk), anchor)
```
fezproof: Does calling createTextNode cause circular dependency options? Can we just pull this function out and reference in 2 places like we did for _insertNode?

fezproof: Should fix the type issues too

fezproof (approved):
All good otherwise from what I can see, nice work!

packages/solid/src/reconciler.ts
Adictya: Oof good catch

packages/solid/src/reconciler.ts
Adictya: Makes sense

fezproof (approved):

PR #111, opened 2025-08-31
fix: improve nested buffer alpha blending and fix ascii background bug

KurtGokhan:
Noticed that the background of ASCII characters weren't preserved when rendered inside a framebuffer. Made some changes to skip rendering invisible foreground and background colors.

While working on that I had to also change the blending algorithm for background colors, as it was also causing subtle issues with ASCII fonts. It can be seen in the videos below, where before the overlay text bg looks more frosted, whereas in the after video it's more actual color blending. Maybe we can add color blending modes in the future to leave the preference to the user.

Before:

https://github.com/user-attachments/assets/a8fe7c1c-8e44-4e7f-9baa-9e30fac2eecd

After:

https://github.com/user-attachments/assets/ef942023-fdad-4824-abd2-6471008a9fbb

### DIFF
--- a/packages/core/src/examples/framebuffer-demo.ts
+++ b/packages/core/src/examples/framebuffer-demo.ts
@@ -97,6 +97,53 @@ export function run(renderer: CliRenderer): void {
     }
   }
 
+
+  const nestedBox = new BoxRenderable(renderer, {
+    id: "nested-box",
+    width: 20,
+    height: 10,
+    position: "absolute",
+    left: 4,
+    top: 4,
+    buffered: true,
+  })
+
+  const inset0 = {
+    position: "absolute" as const,
+    left: 0,
+    top: 0,
+    right: 0,
+    bottom: 0,
+  }
+
+  nestedBox.add(new BoxRenderable(renderer, {
+    ...inset0,
+    border: true,
+    title: 'Nested example',
+    backgroundColor: RGBA.fromInts(120, 0, 120, 120),
+  }));
+
+
+  const nestedInnerBox = new BoxRenderable(renderer, {
+    id: "nested-inner-box",
+    width: 10,
+    height: 4,
+    position: "absolute",
+    left: 5,
+    top: 3,
+    buffered: true,
+  })
+
+  nestedInnerBox.add(new BoxRenderable(renderer, {
+    ...inset0,
+    border: true,
+    title: 'Inner',
+    backgroundColor: RGBA.fromInts(0, 255, 0, 10),
+  }));
+
+  nestedBox.add(nestedInnerBox);
+  renderer.root.add(nestedBox);
+
   const boxObj = new BoxRenderable(renderer, {
     id: "moving-box",
     width: 20,
--- a/packages/core/src/examples/framebuffer-demo.ts
+++ b/packages/core/src/examples/framebuffer-demo.ts
@@ -121,30 +121,32 @@ export function run(renderer: CliRenderer): void {
     zIndex: 2,
   })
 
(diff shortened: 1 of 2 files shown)


kommander:
Could the problem be that drawing the ASCII font uses setCell instead of setCellWithAlphaBlending? 

https://github.com/sst/opentui/blob/main/packages%2Fcore%2Fsrc%2Flib%2Fascii.font.ts#L296-L298

gkurt:
Tried that. The issue persists.

bun run srcexamplesframebuffer-demo ts_2025-08-31_18-45@2x

I think the empty cells don't hit that branch. The issue happens when: 1 - an empty cell is rendered (like between characters) or, 2- inside empty pixels inside characters (like the middle of A character above)

I would expect it to solve the 2nd one but it doesn't. Anyway `setCellWithAlphaBlending` calls `blendCells` so I guess the bug is in `blendCells`

kommander:
I see how that conclusion makes sense. However blending buffers arbitrarily into each other works fine for any buffer except the ASCII renderable one. See the transparency example with the boxes. So I wonder why it works for any buffer but the ASCII one and would look into what the ASCII rendering does differently. I suspect the selection handling could be something that overrides the background color.

kommander:
Yeah, as I suspected the issue is in the `ASCIIFontRenderable` and the `renderFontToFrameBuffer` method. Check https://github.com/sst/opentui/blob/main/packages/core/src/renderables/ASCIIFont.ts#L148 - it clears the framebuffer with the `_bg` color. Removing that line correctly draws the cells opaque between ascii chars. Within the ascii chars the bg is still black, so is suspect there is a bg override there as well.

gkurt:
I noticed that. But I think ASCII font can have a transparent bg, and it should blend as well. Basically when I checked the bg was `RGBA(0,0,0,0)` and it should be transparent, but it was still black.

Actually the nested buffers don't blend well. 

I have an example. Here both the outer and inner elements have buffers. The inner element has 0 opacity. So it should look magenta. Here is how it looks in main branch:

bun run --watch srcexamplesframebuffer-demo ts_2025-08-31_21-42@2x


And I thought I solved this case, but appearently it doesn't look good in my branch as well. Do you have an idea of what's wrong? There are also positioning bugs with buffers. I noticed these kinds of issues when trying to use them in scroll-box.

gkurt:
Here is how it looks in my branch by the way. 
bun run --watch srcexamplesframebuffer-demo ts_2025-08-31_21-48@2x

It looks correct for 0 opacity, but there is some other bug. When I have an opacity like 0.1 the inner element looks fully opaque because it rerenders a 100 times and the opacity quickly adds up. 


https://github.com/user-attachments/assets/416c7050-ff32-4d32-868f-bd0f951c6d5b

So nested buffers don't work well generally. Maybe there wasn't a use case for them before, but I guess they are worth fixing.

I think this PR is in the right direction to fix those bugs. I have gone over the `buffer.zig` changes many times and they are just corrections. I can revert the `perceptualAlpha` change if you want though, that is kinda optional.

kommander:
Can this still be reproduced on current main?

gkurt:
Not this exact issue, but there are issues if I enable `buffered` on outer or inner box.

On outer:

bun run srcexamplesframebuffer-demo ts_2025-09-02_23-50@2x


On inner:

bun run srcexamplesframebuffer-demo ts_2025-09-02_23-52@2x

gkurt:
I am closing this PR, as the issues this PR solved are already solved. We can track other issues separately.

PR #117, opened 2025-08-31
fix(text): fallback to default props when undefined

### LINKED ISSUE #116: React renderer should unset a node style attribute when it is `undefined`
Currently the React renderer ignores style attributes with value null or undefined here (I think), instead it should somehow unset these properties. Not sure if opentui already has a method for that
https://github.com/sst/opentui/blob/618e3b372ee3444544dfb68002d98f4e298b9c50/packages/react/src/utils/index.ts#L169

Repro, notice how the style does not reset on hover end:

```tsx
import { render } from '@opentui/react'
import { useState } from 'react'

function HoverExample() {
    const [hovered, setHovered] = useState<boolean>(false)

    return (
        <box style={{ padding: 2 }}>
            <box
                onMouseOut={() => setHovered(false)}
                onMouseMove={() => setHovered(true)}
                style={{ padding: 1 }}
            >
                <text bg={hovered ? '#FF0000' : undefined}>Hover over me</text>
            </box>
        </box>
    )
}

render(<HoverExample />)
```

https://github.com/user-attachments/assets/c75777e8-7bd4-4aee-80a0-9a8e66ffa3a7

msmps:
## Description

- closes #116 

## Screencast

CleanShot 2025-08-31 at 20 43 52

### DIFF
--- a/packages/core/src/renderables/Text.ts
+++ b/packages/core/src/renderables/Text.ts
@@ -100,6 +100,7 @@ export class TextRenderable extends Renderable {
     const newColor = value ? parseColor(value) : this._defaultOptions.selectionBg
     if (this._selectionBg !== newColor) {
       this._selectionBg = newColor
+      this.syncSelectionToTextBuffer()
       this.requestRender()
     }
   }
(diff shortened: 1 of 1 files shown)


remorses:
wow you are fast!

kommander:
I think we might need to `syncSelectionToTextBuffer` when selection bg/fg changes?

msmps:
> I think we might need to `syncSelectionToTextBuffer` when selection bg/fg changes?

you might be right! i just added this and tested it (_super edge case_) 👍

kommander (approved):

PR #118, opened 2025-08-31
More efficient `useKeyboard`

remorses:
I added useEvent which let you create a stable function that always uses the latest version

This way there is no need to pass handler to the useEffect dependencies (which would usually be different on every render)

### DIFF
--- a/packages/react/src/hooks/use-event.tsx
+++ b/packages/react/src/hooks/use-event.tsx
@@ -1,0 +1,22 @@
+import { useCallback, useLayoutEffect, useRef } from "react"
+
+/**
+ * Returns a stable callback that always calls the latest version of the provided handler.
+ * This prevents unnecessary re-renders and effect re-runs while ensuring the callback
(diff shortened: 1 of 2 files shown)


msmps:
I struggle with this one as this is something that I would expect to be handled in user land with memoization? Is there a specific example you have where this has caused friction? [Ink](https://github.com/vadimdemedes/ink/blob/a3ba1f14a4a3787c3a333b72117e86da1a4a2d57/src/hooks/use-input.ts#L117) doesn't handle this for you.


https://github.com/user-attachments/assets/686fcae1-d00e-481a-a69a-514a00ed6317

remorses:
Adding memoization works too but it becomes easier to forget to add a dependency on the array and create bugs

useKeyboard should always use the function defined in the last render anyway, there is no use case where you want to use a memorized function from a previous render. Using useEvent ensures this behavior and makes the API simpler to use

msmps (approved):

PR #119, opened 2025-08-31
feat: add some utilities and missing yoga properties

KurtGokhan:
Added these yoga props:
- alignSelf
- flexWrap
- overflow
- Fixed types of min/max width/height

Added `translateX` and `translateY` which will be helpful when implementing scroll area to avoid recalculating Yoga layout. It can be used for other elements as well.

Some Typescript goodies:
- Excluded `dist` to allow getting dist folder in autocomplete
- Some type-safety stuff with generic types

Added `"` to open console keys. That's what I have in my keyboard :(

Emit `resize` event on resizes. This can be used in user-land to detect element resizes.


kommander (approved):
👍

PR #132, opened 2025-09-03
Add ScrollBox to SolidJs

Adictya:


packages/solid/src/elements/hooks.ts
```diff
@@ -51,7 +51,7 @@ export const useTerminalDimensions = () => {
   return terminalDimensions
 }
 
-export const useKeyHandler = (callback: (key: ParsedKey) => void) => {
+export const useKeyboard = (callback: (key: ParsedKey) => void) => {
```
msmps, line 54: ⚠️  this will be a breaking change for consumers (_even though alpha_) so might be a good idea to document this and/or export both for now (_one aliased and mark `useKeyHandler` as deprecated_)

msmps (dismissed):

packages/solid/src/elements/hooks.ts
Adictya: Makes sense, added 👍🏻 

Although I don't think we have settled on the api surface yet. For eg. I think it might be better for scrollbox to just be box and set a property inside it to make it scrollable. So I don't know if having a stable api can be within expectation. 

We can try to put up notices / release notes rather than adding compat stuff

PR #133, opened 2025-09-03
fix: add reactive scroll properties

msmps:


packages/core/src/renderables/ScrollBar.ts
```diff
@@ -209,6 +208,22 @@ export class ScrollBarRenderable extends Renderable {
...
+  public set thumbOptions(options: ScrollBarOptions["thumbOptions"]) {
+    Object.assign(this.slider, { thumbColor: options?.backgroundColor })
+    this.requestRender()
+  }
+
+  public set trackOptions(options: ScrollBarOptions["trackOptions"]) {
+    Object.assign(this.slider, { trackColor: options?.backgroundColor })
+    this.requestRender()
+  }
+
```
msmps: these are just mapping single values `trackOptions.backgroundColor` to `trackColor: value` we could cut the middle out and just replace `trackOptions.backgroundColor` with `trackColor` ? same for the other

packages/core/src/renderables/ScrollBar.ts
```diff
@@ -359,7 +373,7 @@ export class ArrowRenderable extends Renderable {
 
   set fg(value: RGBA) {
     if (this._fg !== value) {
-      this._fg = value
+      this._fg = parseColor(value)
```
kommander: Using parseColor the setter should take a string as well as RGBA

packages/core/src/renderables/ScrollBar.ts
```diff
@@ -1,17 +1,16 @@
...
 import { type BoxOptions } from "./Box"
-import type { OptimizedBuffer } from "../buffer"
 import { SliderRenderable } from "./Slider"
 
 export interface ScrollBarOptions extends RenderableOptions<ScrollBarRenderable> {
   orientation: "vertical" | "horizontal"
   showArrows?: boolean
-  trackOptions?: BoxOptions
-  thumbOptions?: BoxOptions
+  trackOptions?: Pick<BoxOptions, "backgroundColor">
```
kommander: The track is a slider basically and can take slider options?

packages/core/src/renderables/ScrollBar.ts
```diff
@@ -209,6 +208,22 @@ export class ScrollBarRenderable extends Renderable {
     }
   }
 
+  public set arrowOptions(options: ScrollBarOptions["arrowOptions"]) {
+    Object.assign(this.startArrow, options)
+    Object.assign(this.endArrow, options)
+    this.requestRender()
+  }
+
+  public set thumbOptions(options: ScrollBarOptions["thumbOptions"]) {
```
kommander: Yeah I wouldn't separate track and thumb, that's both in Slider options?

packages/core/src/renderables/ScrollBar.ts
```diff
@@ -293,7 +308,6 @@ export interface ArrowOptions extends RenderableOptions<ArrowRenderable> {
   fg?: string | RGBA
   bg?: string | RGBA
   attributes?: number
-  backgroundColor?: string | RGBA
```
kommander, line 296: Was that unused and should be used?

packages/core/src/renderables/ScrollBox.ts
```diff
@@ -45,7 +45,6 @@ export interface ScrollBoxOptions extends BoxOptions<ScrollBoxRenderable> {
   scrollbarOptions?: Omit<ScrollBarOptions, "orientation">
   verticalScrollbarOptions?: Omit<ScrollBarOptions, "orientation">
   horizontalScrollbarOptions?: Omit<ScrollBarOptions, "orientation">
-  arrowOptions?: Omit<ArrowOptions, "direction">
```
kommander, line 48: :+1:

packages/core/src/renderables/ScrollBar.ts
msmps: it was **unused** but i've flipped it now to align on properties so `fg/bg` is now `foregroundColor/backgroundColor`

kommander (approved):
I think that looks good. We can figure out further interface details with some more usage and testing.

PR #140, opened 2025-09-05
Fix solid not anchoring to text chunk correctly

Adictya:


Adictya:
@fezproof take a look when you find some time. Merging pre-emptively

PR #147, opened 2025-09-08
Fix solid renderable lifecycle issues

fezproof:
Fixes issue with solid reusing old nodes in list arrangement. This should also make things more performant.
Fixes issue where parents where destroying themselves instead of their children first, leaving dangling children.


kommander:
Good find! We are using the examples for repros and testing basically, I am guilty of that as well. The examples are getting crowded and should be more of a feature showcase. Can we just move the reproductions into a dev/reproductions folder or something until we have proper testing set up?

Issue #148, opened 2025-09-09
requestAnimationFrame is broken


arv:
Here is a minimal repro:

```tsx
import { render } from "@opentui/react";
import { useEffect, useState } from "react";

// eslint-disable-next-line react-refresh/only-export-components
function App() {
  const [called, setCalled] = useState(false);

  useEffect(() => {
    requestAnimationFrame(() => {
      setCalled(true);
    });
  }, []);

  return <text>{called ? "Called" : "Not called"}</text>;
}

render(<App />);

```

When the app is first rendered the UI says `Not called`. If you force a render (by clicking or resizing the terminal) the callback gets called.

msmps:
## Notes
- https://github.com/sst/opentui/blob/b1c33a0b2db8a7f458b3cc3d5f2dab17f5924188/packages/core/src/renderer.ts#L354-L367
- only runs callbacks on `loop` when `live` is true

msmps:
fixed by #149

PR #149, opened 2025-09-09
fix: trigger render on raf to run animation request

### LINKED ISSUE #131: Changing variables not re-rendering
I am trying to use [Zero Sync Engine](https://zero.rocicorp.dev) like this:

```tsx
import { useQuery, useZero, ZeroProvider } from "@rocicorp/zero/react";
import { schema, Schema } from "./schema";

function App() {
  const z = useZero<Schema>();
  const [messages] = useQuery(z.query.message);

  return (
    <box>
	{messages.map(m => <box>
	  <text>{m.body}</text>
	</box>)}
    </box>
  )
}
```

This code works using [ink](https://github.com/vadimdemedes/ink), but not with OpenTUI.

However, this does work with OpenTUI when I add some other state, such as this

```tsx
import { useQuery, useZero, ZeroProvider } from "@rocicorp/zero/react";
import { schema, Schema } from "./schema";

function App() {
  const z = useZero<Schema>();
  const [messages] = useQuery(z.query.message);

  // adding this will fix it
  useEffect(() => {
    const interval = setInterval(() => {
      setCounter((prevCount) => prevCount + 1)
    }, 60)
    return () => clearInterval(interval)
  }, []);

  return (
    <box>
	{messages.map(m => <box>
	  <text>{m.body}</text>
	</box>)}
    </box>
  )
}
```

Any idea on why messages wouldn't rerender by itself?

Thanks for the help,
Max

msmps:
## Description

- handles triggering a render on `requestAnimationFrame` to run the callbacks
- closes #131 and #148

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -354,7 +354,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     global.requestAnimationFrame = (callback: FrameRequestCallback) => {
       const id = CliRenderer.animationFrameId++
       this.animationRequest.set(id, callback)
-      this.intermediateRender()
+      this.requestLive()
       return id
     }
     global.cancelAnimationFrame = (handle: number) => {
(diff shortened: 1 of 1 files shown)


msmps:
closes #148

PR #154, opened 2025-09-11
Update scroll examples to use solid-isms

fezproof:
Updated solid scroll examples to use solid-isms like For and Index
Added an Index demo


packages/solid/examples/index.tsx
```diff
@@ -2,6 +2,8 @@ import { render } from "@opentui/solid"
 import { ConsolePosition } from "@opentui/core/src/console"
 import ExampleSelector from "./components/ExampleSelector"
 
+process.env.DEBUG = "true"
```
Adictya: intentional?

fezproof: Ill remove

PR #155, opened 2025-09-11
Fix text node rendering in Solid

fezproof:
https://github.com/user-attachments/assets/161493a4-1e47-4c25-98de-b70e67fdcb07

Adds `<span>` element
Makes it so only `<text>` can have text

You can no longer used styled text directly in the renderer, you must use span. This can be fixed later

### DIFF
--- a/packages/solid/examples/components/text-selection-demo.tsx
+++ b/packages/solid/examples/components/text-selection-demo.tsx
@@ -128,7 +128,7 @@ export default function TextSelectionDemo() {
             <text style={{ width: 27, height: 1, zIndex: 26, selectionBg: "#4a5568", selectionFg: "#ffffff" }}>
               <span style={{ fg: "yellow" }}>Important:</span>{" "}
               <span style={{ bold: true, fg: "cyan" }}>Nested content</span>{" "}
-              <span style={{ italic: true, fg: "greem" }}></span>
+              <span style={{ italic: true, fg: "green" }}></span>
             </text>
           </box>
         </box>


packages/core/src/renderables/TextNode.ts
```diff
@@ -229,11 +238,15 @@ export class TextNodeRenderable extends BaseRenderable {
 }
 
 export class RootTextNodeRenderable extends TextNodeRenderable {
+  textParent: TextRenderable
```
kommander, line 241: What does it use the textParent for?

packages/solid/examples/components/text-style-demo.tsx
```diff
@@ -18,16 +17,21 @@ export default function TextStyleScene() {
 
   return (
     <box>
-      <text>Simple text works! {counter()} times</text>
-      <text>{underline(bold(`Chunk also works! ${counter()} times`))}</text>
-      <text>{t`${italic(fg("#adff2f")("Styled"))} ${bold(fg("#ff8c00")("Text"))} also works! ${counter()} times`}</text>
+      <text style={{ bg: "red", fg: "black" }}>Simple text works! {counter()} times</text>
+      <text style={{ bg: "red", fg: "black" }}>
+        Hello {counter()} <span style={{ bg: "yellow", fg: "black" }}>World</span>{" "}
```
kommander, line 22: Does StyledText as direct child not work anymore?

packages/solid/src/reconciler.ts
```diff
@@ -28,12 +40,19 @@ export type DomNode = Renderable | TextNode | TextChunk
...
-  if (isTextChunk(node)) {
-    return node.text
-  }
   return node.id
 }
 
+const getNodeChildren = (node: DomNode) => {
+  let children
+  if (node instanceof TextRenderable) {
+    children = node.getTextChildren()
```
kommander, line 49: As we delegate add/remove to the rootTextNode, we could delegate getChildren there as well and not have another method? I may just make the TextRenderable the actual root node.

packages/solid/src/reconciler.ts
```diff
@@ -45,85 +64,68 @@ function _insertNode(parent: DomNode, node: DomNode, anchor?: DomNode): void {
...
+
+  return TextNode.fromString(value, { id })
+}
+
+function _getParentNode(childNode: DomNode): DomNode | undefined {
+  log("Getting parent of node:", logId(childNode))
+
+  let parent = childNode.parent ?? undefined
+  if (parent instanceof RootTextNodeRenderable) {
+    parent = parent.textParent ?? undefined
```
kommander, line 126: Ah, I see. Then it might help to just make the TextRenderable the RootTextNode?

fezproof: Yeah I think that would be the way forward

packages/solid/examples/components/text-style-demo.tsx
fezproof: I does not, but this was causing so many issues and complexity. It might be better to just implement things like b and i tags for simple things, and use span for complex styles. Seems to work quite well

packages/solid/src/reconciler.ts
fezproof: Can we merge this for now and I clean that up in a follow up PR? I want to know if this event fixes @thdxr issues

PR #169, opened 2025-09-17
Feat(solid): Handle anchor node and add Dynamic and Portal

Adictya:
- Simple `bun test` should also work now
- Added tests for Dynamic and Portal
- Brought in universal renderer for custom addAnchorNode logic
- Fixes a bunch of bugs around Show and For 
- We throw on not adding unwrapped text/span.


Adictya:
~WIP, fixing and linting~
Done

packages/solid/tests/control-flow.test.tsx
```diff
@@ -201,7 +201,7 @@ describe("SolidJS Renderer - Control Flow Components", () => {
       await testSetup.renderOnce()
 
       children = testSetup.renderer.root.getChildren()[0]!.getChildren()
-      expect(children.length).toBe(1)
+      expect(children.length).toBe(2)
```
Adictya, line 204: Solid subs existing element with the empty node so number of children remain same

packages/core/src/renderables/Slot.ts
```diff
@@ -0,0 +1,130 @@
+import { BaseRenderable, isTextNodeRenderable, TextNodeRenderable, TextRenderable } from ".."
```
kommander: Are the slots something reconciler specific that we could have in the solid package? For core itself they don't seem to have a use-case.

packages/solid/tests/dynamic-portal.test.tsx
```diff
@@ -0,0 +1,311 @@
...
+            </Dynamic>
+          </box>
+        ),
+        { width: 20, height: 5 },
+      )
+
+      await testSetup.renderOnce()
+
+      setComponentType("box")
+      await testSetup.renderOnce()
```
kommander, line 104: Is this missing an expectation or just testing that it doesn't crash?

packages/solid/tests/layout.test.tsx
```diff
@@ -51,6 +51,31 @@ describe("SolidJS Renderer Integration Tests", () => {
       expect(frame).toMatchSnapshot()
     })
 
+    it("should throw on rendering text without parent <text> element", async () => {
+      expect(
+        testRender(() => <box>This text is not wrapped in a text element</box>, {
```
kommander, line 56: I hope we can get this to work in the future, would be so neat.

packages/solid/tests/dynamic-portal.test.tsx
Adictya: Yup I'll document

packages/solid/tests/layout.test.tsx
Adictya: Already on it dw

packages/core/src/renderables/Slot.ts
Adictya: Yeah will have to explore around the api to make it work well for core. But I was thinking if it had a replaceSelf method, which allows us to create a renderable with slots and expose those such that on usage we could just slot in custom renderables. For eg Dialog component with header, content and footer slots.

This is sort of a low cost node that can be replaced when needed.

I will move it for now tho.

PR #171, opened 2025-09-19
Feat(core+solid): make id mutable and fix yoga import issue

Adictya:


packages/core/scripts/build.ts
```diff
@@ -301,8 +301,8 @@ if (buildLib) {
         repository: packageJson.repository,
         bugs: packageJson.bugs,
         exports,
-        dependencies: packageJson.dependencies,
```
Adictya, line 304: Since we bundle, the npm package wouldn't have dependencies

packages/core/src/Renderable.ts
```diff
@@ -148,6 +149,15 @@ export abstract class BaseRenderable extends EventEmitter {
   public abstract getRenderable(id: string): BaseRenderable | undefined
   public abstract requestRender(): void
 
+  public get id(): string {
+    console.log("=>> get id", this._id)
```
kommander: leftover

packages/core/src/tests/renderable.test.ts
```diff
@@ -202,6 +202,22 @@ describe("Renderable - Child Management", () => {
     expect(result).toBe(-1)
     expect(parent.getChildrenCount()).toBe(0)
   })
+
+  test("can change renderable id and updates parent mapping", () => {
```
kommander, line 206: Yay tests!

kommander (dismissed):

PR #172, opened 2025-09-20
Replace npx with bunx

ytsruh:
Whilst working in the repo to try and get the Go bindings working I ran into this issue where Node/NPM was required even though this is a bun project. 

This is a very small change to switch the npx usage for bunx & removes the node/npm dependency


kommander:
Wait, why are the workflows not running again? They should run on every push. Need to check.

PR #173, opened 2025-09-22
Fix(solid): Solid insert sibling order issue

Adictya:


packages/solid/src/reconciler.ts
```diff
@@ -344,7 +348,7 @@ export const {
       log("No parent found for node:", logId(node))
       return undefined
     }
-    const siblings = getNodeChildren(node)
```
Adictya, line 347: smh

kommander (approved):

PR #198, opened 2025-10-03
Scroll acceleration for macOS

remorses:
This PR adds scroll accelearation, enabled by default for macOS


remorses:
The curve is a bit dumb right now. it is just 3 lines so it does not feel great

kommander:
This is cool. Can we make the acceleration a generic interface that the MacOS acceleration implements then? I would love to set up some test coverage for the scrollbox, the slider tests use the mock mouse input for example that we can test that with. The scrollbox is getting really complex. I don't think this breaks anything, but some basic coverage would be great to make sure it keeps working as expected when I doctor around to debug the viewport intersection bug.

remorses:
I added some test and added the accelearation interface

remorses:
There seems to be a problem right now with Ghostty where it sends many scroll events for one mouse wheel tick: https://github.com/ghostty-org/ghostty/discussions/7577

I can open a PR to ignore ticks that happen closer than 6ms if Ghostty does not plan to fix that

https://github.com/ghostty-org/ghostty/discussions/3955

PR #202, opened 2025-10-06
Show error stack trace when an error is thrown in React render

remorses:
Added an ErrorBoundary component that displays the error stack trace instead of a black screen

Screenshot 2025-10-06 at 13 33 17


kommander:
Time to set up react testing.

PR #209, opened 2025-10-09
Add support for testing React components

remorses:
Added support for passing a custom renderer in React render and make it wait until React commits

I added a ready field to the TestRenderer to be able to await the React commit. Because React renders asynchronously

### DIFF
--- a/packages/react/src/reconciler/renderer.test.tsx
+++ b/packages/react/src/reconciler/renderer.test.tsx
@@ -66,6 +66,4 @@ test("renders input and box with state", async () => {
     │hello world                           │
     └──────────────────────────────────────┘"
   `)
-
-
 })


msmps:
i'm not sure on this one... i had started looking into changing the api for this previously but didn't finish it! i was leaning towards exposing a `createRoot` method that takes a renderer to align with standards.

Something like:

```tsx
import { createRoot } from "@opentui/react";

const renderer = await createCliRenderer(<config>);
createRoot(renderer).render(<App />);
```

remorses:
That makes sense. I also have some issues creating the promise for ready in a way that does not create issues with Solid

kommander:
I think this was solved by #309

PR #215, opened 2025-10-12
fix: standardize TextNode remove method to use ID-based lookup to resolve "Child not found in children" error

eli0shin:
There is an architectural mismatch in the class hierarchy:

1. **BaseRenderable** declares `abstract remove(id: string): void` - expecting ID-based lookup
2. **Renderable** (which most components inherit from) implements this correctly:
   - Uses `renderableMapById.get(id)` to look up child by ID
   - Properly removes it from layout and internal data structures
3. **TextNodeRenderable** (which `SpanRenderable` inherits from) implements it incorrectly:
   - Signature: `remove(child: string | TextNodeRenderable): this`
   - Uses `indexOf(child)` to find the child - NO ID LOOKUP
   - Expects a direct child reference or string literal

For example, this code works:
```javascript
<box>
  {
    showTextA ?
      <text><span>text a</span></text> :
      <text><span>text b</span></text>
  }
</box>
```

But this code throws the error because the child of `<text>` is conditional
```javascript
<box>
   <text>
    {
      showTextA ?
        <span>text a</span> :
        <span>text b</span>
    }
  </text>
</box>
```

The solid reconciler [passes node reference directly for TextNodeRenderables](https://github.com/sst/opentui/blob/7e53ea49e6f65040c0673b2fb8682b9602ddd1e9/packages/solid/src/reconciler.ts#L121) and passes it to `remove` [but the react one does not](https://github.com/sst/opentui/blob/7e53ea49e6f65040c0673b2fb8682b9602ddd1e9/packages/react/src/reconciler/host-config.ts#L59-L61), which raises the question: Why are `TextNodeRenderable`s different than other renderables? Why do they use remove by reference instead of by id?

This PR aligns TextNodeRenderable with other renderables and updates the solid reconciler to pass children to remove by id.

Resolves https://github.com/sst/opentui/issues/214

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -16,6 +16,7 @@
       },
       "devDependencies": {
         "@types/bun": "latest",
+        "@types/node": "latest",
         "@types/three": "0.177.0",
         "commander": "^13.1.0",
         "typescript": "^5",
@@ -46,6 +47,8 @@
       },
       "devDependencies": {
         "@types/bun": "latest",
+        "@types/node": "latest",
+        "@types/react": "^19.0.0",
         "@types/react-reconciler": "^0.32.0",
         "typescript": "^5",
       },
@@ -67,6 +70,7 @@
       "devDependencies": {
         "@types/babel__core": "7.20.5",
         "@types/bun": "latest",
+        "@types/node": "latest",
         "typescript": "^5",
       },
       "peerDependencies": {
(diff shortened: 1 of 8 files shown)


kommander:
The pipeline takes latest bun, which is now 1.3 and the build fails with some type error. Doesn't happen locally, need to check that.

Issue #219, opened 2025-10-20
Feedback + contributing guideline


hanpham32:
Since learning about OpenTUI from Opencode, I'm beginning to use OpenTUI to build new TUI apps. It's been a fun experience using TS to put together a UI in the terminal. I can see OpenTUI be a powerful framework for future TUI app developments. I would love to contribute to this open source project, however, there isn't a standard guideline for that. Would also love to see an official OpenTUI API documentation. I understand this is an on-going development framework so there will be many breaking changes.

kommander:
Cool, want to write a guideline? :) Bug fixes are always welcome, feature suggestions, also in the form of a PR as well. Features are subject to discussion though and I cannot guarantee anything. Support/binding packages like react/solid/vue are handled by their respective maintainers, where I am also trying to look after solid, as that is what is used in opencode. Should add tests and must pass in the pipeline. Otherwise I am open to suggestions on what the guidelines could be.

hanpham32:
Yes, happy to work on this :)

PR #220, opened 2025-10-21
feat: contributing guideline

hanpham32:
Addresses issue #219


hanpham32:
Who are the other active maintainers for the other packages?

kommander:
There is a codeowners file in the .github dir, I think that should suffice, no need to mention maintainers here again.

hanpham32:
I've removed the maintainers section. Is there anything else you think will be good?

PR #225, opened 2025-10-22
feat: add support for transparent terminal background rendering

KeremP:
This addresses https://github.com/sst/opentui/issues/184 by defaulting to terminal background if renderer alpha channel is 0.

The transparent background can be set by using `renderer.setBackgroundColor("transparent")` for example in `packages/react/examples/basic.tsx`.
```diff
import { bold, fg, italic, t, TextAttributes } from "@opentui/core"
import { render, useKeyboard, useRenderer } from "@opentui/react"
import { useCallback, useState } from "react"

export const App = () => {
  const renderer = useRenderer()
+  renderer.setBackgroundColor("transparent")
  const [username, setUsername] = useState("")
  const [password, setPassword] = useState("")
```

Example in ghostty
image


kommander:
Hey, that would be a great addition. I think something went wrong with the merge commits though, can you try rebasing on main?

KeremP:
ohp - my bad. should be good now 😀

kommander:
Ah cool! Simple, I like it. Does this allow for mixed transparent and non transparent areas? I need to take it for a spin.

packages/core/src/renderer.ts
```diff
@@ -67,6 +67,7 @@ export interface CliRendererConfig {
   useConsole?: boolean
   experimental_splitHeight?: number
   useKittyKeyboard?: boolean
+  backgroundColor?: ColorInput
```
kommander, line 70: I think this is unused? Can be used, makes sense.

kommander:
Works great as far as I have tested. I'd say it could even be alpha 0 by default for the background color then? So if someone has set opacity for their terminal it will "just work" as expected. Run `bun prettier:write" and fix backgroundColor renderer option and I'll bring this in.

PR #238, opened 2025-10-27
Implement Javascript Output Mode to Support SSH

### LINKED ISSUE #187: Uses process IO Instead of SSH2 socket IO
Hey,

As of Bun v1.2.5, I noticed that the ssh2 server package should be working correctly. So I decided to try integrating @opentui/react with it. Unfortunately, things didn’t go quite as expected.

When I start the SSH server, the TUI output is written to the server’s process IO instead of the client’s stream when connecting (provided by ssh2). I attempted to debug this myself, but due to my limited understanding of the codebase, I wasn’t able to figure out the issue.

Interestingly, when using nearly identical code with Ink instead of @opentui/react, only replacing the render and elements, everything works, the UI renders to the SSH client stream, with minor issues e.g. the known LF/CRLF issue which can be patched.

### Create ssh key
Generate key in same folder as the code
`ssh-keygen -t rsa -b 4096 -f host.key`

### Code snippet

```typescript
// server.ts
import { Server as SSHServer } from "ssh2";
import { render } from "@opentui/react";
import { readFileSync } from "fs";
import { PassThrough } from "stream";

const ssh = new SSHServer({
  hostKeys: [readFileSync('host.key')]
});

ssh.on("connection", (client, info) => {
  client.on("authentication", (ctx) => {
    ctx.accept();
  });

  client.on("ready", () => {
    client.on("session", (accept, reject) => {
      const session = accept();
      session.on("pty", (acceptPty) => {
        acceptPty();
      });

      session.on("shell", async (acceptShell) => {
        const stream = acceptShell();

        const sshInput = new PassThrough();
        const sshOutput = new PassThrough();

        // Pipe streams between SSH and OpenTUI
        stream.pipe(sshInput);
        sshOutput.pipe(stream);

        const App = () => (
          <box style={{ flexDirection: "column", padding: 1 }}>
            <text>Welcome to my TUI over SSH!</text>
          </box>
        );

        // render with custom stdin/stdout
        await render(<App />, {
          stdin: sshInput as unknown as NodeJS.ReadStream,
          stdout: sshOutput as unknown as NodeJS.WriteStream,
          exitOnCtrlC: true,
        });
      });
    });
  });
});

ssh.listen(2222, "0.0.0.0", () => {
  console.log("SSH server listening on port 2222");
});
```

kitlangton:
This resolves #187 

I'll add more details later. This is to enable something like terminal.shop to be rewritten in opentui and be run via ssh (so teej can rewrite 😜).

### DIFF
--- a/packages/core/src/output-strategy.test.ts
+++ b/packages/core/src/output-strategy.test.ts
@@ -1,0 +1,48 @@
+import { EventEmitter } from "events"
+import { expect, test } from "bun:test"
+import type { Pointer } from "bun:ffi"
+
+import { createOutputStrategy } from "./output-strategy"
+import type { RenderLib } from "./zig"
+import { createCapturingStdout } from "./testing/stdout-mocks"
+
+test("javascript output strategy flushes via provided write function", () => {
+  const frame = Buffer.from("frame-bytes")
+  const writes: Buffer[] = []
+
+  const stdout = createCapturingStdout()
+  stdout.write = () => {
+    throw new Error("stdout.write should not be hit when writeToTerminal is provided")
+  }
+
+  const stdin = new EventEmitter() as unknown as NodeJS.ReadStream
+
+  const libMock = {
+    setWriteTarget: () => {},
+    getWriteBufferLength: () => frame.length,
+    copyWriteBuffer: (_renderer: Pointer, target: Uint8Array) => {
+      target.set(frame)
+      return frame.length
+    },
+  } as unknown as RenderLib
+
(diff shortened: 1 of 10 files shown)


kommander:
The build failures are just some typescript type errors. 

I would just try to avoid copying the output buffer for every frame to avoid reallocations and the copy overhead by exposing pointers to the output buffers raw memory via lib.zig, connect them once with toArrayBuffer when setting up and then write out from that directly into buns buffered stdout. 

Other than that I love this and want to have that in!

Issue #245, opened 2025-10-31
Components that render @opentui/react components do not support key prop


macklinu:
```tsx
import { test, expect } from "bun:test"

test("components that wrap opentui react components support React key prop", () => {
  const Double = ({ children }: { children: number }) => <text>{children * 2}</text>

  const numbers = [1, 2, 3]

  expect(() => (
    <box>
      {numbers.map((number) => (
        <Double key={number}>{number}</Double>
      ))}
    </box>
  )).not.toThrow()
})

```

results in:

```
src/components/index.test.tsx:11:17 - error TS2322: Type '{ children: number; key: number; }' is not assignable to type '{ children: number; }'.
  Property 'key' does not exist on type '{ children: number; }'.

11         <Double key={number}>{number}</Double>
```

I would expect the Double component to accept the `key` prop. I'm having trouble figuring out where to make this change locally after poking around the `@opentui/react` types otherwise I would have submitted a PR.

macklinu:
Actually figured it out - https://github.com/sst/opentui/pull/246.

msmps:
fixed by #246

PR #246, opened 2025-10-31
Fix React key prop

macklinu:
This PR resolves the issue I opened in #245. The test I added doesn't need to be the exact unit test but does at least show the TypeScript error for `key` no longer exists.


msmps (approved):

msmps:
thanks for this fix @macklinu

PR #249, opened 2025-11-01
fix(solid): init border when supporting props are present

### LINKED ISSUE #186: difference in opentui/solid and opentui/react border implementations
in react this renders with a border
```tsx src/index.tsx
import { TextAttributes } from "@opentui/core";
import { render } from "@opentui/react";

function App() {
  return (
    <box alignItems="center" justifyContent="center" flexGrow={1}>
      <box justifyContent="center" alignItems="flex-end" borderStyle="single">
        <ascii-font font="tiny" text="OpenTUI" />
        <text attributes={TextAttributes.DIM}>What will you build?</text>
      </box>
    </box>
  );
}

render(<App />);
```

in solid this doesn't render with a border:
```tsx src/index.tsx
import { TextAttributes } from "@opentui/core";
import { render } from "@opentui/solid";

render(() => (
  <box alignItems="center" justifyContent="center" flexGrow={1}>
    <box justifyContent="center" alignItems="flex-end" borderStyle="single">
      <ascii_font font="tiny" text="OpenTUI" />
      <text attributes={TextAttributes.DIM}>What will you build?</text>
    </box>
  </box>
));
```
you have to specify the `border` prop also for some reason
```diff src/index.tsx
import { TextAttributes } from "@opentui/core";
import { render } from "@opentui/solid";

render(() => (
  <box alignItems="center" justifyContent="center" flexGrow={1}>
-     <box justifyContent="center" alignItems="flex-end" borderStyle="single">
+     <box justifyContent="center" alignItems="flex-end" borderStyle="single" border>
      <ascii_font font="tiny" text="OpenTUI" />
      <text attributes={TextAttributes.DIM}>What will you build?</text>
    </box>
  </box>
));
```

I think the react version of this makes more sense (and the react version also aligns with the logic in the Box renderable). would you accept a pr making the solid implementation match closer to the react version?

msmps:
## Fix: Auto-enable border when border props are set in Solid

Fixes #186

### Problem
In Solid, setting `borderStyle`, `borderColor`, or `focusedBorderColor` on `<box>` elements doesn't display borders unless `border` is explicitly set to `true`. This behavior differs from React, where borders appear automatically.

```tsx
// Works in React, doesn't work in Solid
<box borderStyle="single">
  <text>Content</text>
</box>
```

### Root Cause
The Box constructor has logic to auto-enable borders when border-related props are provided. However:
- **React reconciler**: Passes all props during construction ✅
- **Solid reconciler**: Only passes `{ id }` during construction, sets props via setters afterwards ❌
    - `createElement` doesn't have access to props

### Solution
Extracted border initialization logic into a private `initializeBorder()` method and call it from the property setters for:
- `borderStyle`
- `borderColor`
- `focusedBorderColor`

This ensures borders are auto-enabled regardless of when/how props are set.

### Testing
Added test cases in `packages/solid/tests/layout.test.tsx` verifying borders auto-enable for all border-related props.

### DIFF
--- a/packages/core/src/renderables/Box.ts
+++ b/packages/core/src/renderables/Box.ts
@@ -1,6 +1,5 @@
 import { Edge, Gutter } from "yoga-layout"
 import { type RenderableOptions, Renderable } from "../Renderable"
-import { isValidPercentage } from "../lib/renderable.validations"
 import type { OptimizedBuffer } from "../buffer"
 import {
   type BorderCharacters,
@@ -11,6 +10,7 @@ import {
   getBorderSides,
 } from "../lib"
 import { type ColorInput, RGBA, parseColor } from "../lib/RGBA"
+import { isValidPercentage } from "../lib/renderable.validations"
 import type { RenderContext } from "../types"
 
 export interface BoxOptions<TRenderable extends Renderable = BoxRenderable> extends RenderableOptions<TRenderable> {
@@ -91,6 +91,18 @@ export class BoxRenderable extends Renderable {
     }
   }
 
+  private initializeBorder(): void {
+    // https://github.com/sst/opentui/issues/186
+    // Solid-js reconciler does not pass props to constructor on init,
+    // so we need to initialize the border when supporting properties are set.
+    // borderStyle, borderColor, focusedBorderColor
+    if (this._border === false) {
+      this._border = true
+      this.borderSides = getBorderSides(this._border)
+      this.applyYogaBorders()
+    }
+  }
+
   public get customBorderChars(): BorderCharacters | undefined {
     return this._customBorderCharsObj
   }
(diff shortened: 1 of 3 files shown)


packages/core/src/renderables/Box.ts
```diff
@@ -132,9 +144,10 @@ export class BoxRenderable extends Renderable {
 
   public set borderStyle(value: BorderStyle) {
     let _value = value ?? this._defaultOptions.borderStyle
-    if (this._borderStyle !== _value) {
+    if (this._borderStyle !== _value || !this._border) {
```
msmps, line 147: `!this._border` condition required for scenarios where value === defaultValue so this condition never fires

msmps:
i can also solve this at the reconciler level (_have that fix locally too_) given it might be quite a framework specific issue? lmk your thoughts @kommander

PR #256, opened 2025-11-03
Fix CJK character corruption in text rendering

### LINKED ISSUE #255: bug: CJK char corruption in text rendering
## Problem

Both Solid and React reconcilers corrupt certain CJK characters during rendering. ~Core handles them correctly.~ 

```
Expected → Actual (Solid/React)
你好世界 → 你好世ç界
中文   → 中文æ
한글   → í한글
```

Emoji & ASCII unaffected.

## Root Cause

Encoding/displayWidth mismatch in JSX→TextNode conversion path. Likely culprit: reconciler text handling vs core displayWidth calc.

## Investigation Notes

- Core library handles all CJK/emoji correctly
- Issue affects both Solid and React reconciler text rendering paths
- Reproduction: specific CJK chars render as different glyphs

## Next Steps

1. Debug TextNode creation from Solid/React JSX
2. Verify displayWidth consistency with core
3. Check for encoding issues in reconciler text handling

---

*Created with assistance from OpenCode & Claude Haiku 4.5*

zenyr:
## Summary

This PR fixes the root cause of CJK character corruption (#255) by modifying the renderer to skip continuation cells instead of writing space placeholders.

## Root Cause Analysis

CJK characters (width=2) occupy 2 terminal columns but are stored as 1 cell in the buffer. The original implementation had a fundamental issue:

1. CJK character rendered at cell position x
2. Continuation cell at (x+1) marked as occupied
3. Space character written at continuation cell **overwrites the next actual character**
4. Result: CJK corruption

## Changes in Latest Commit

**renderer.zig** (6449abb):
- Skip rendering continuation cells entirely - they're placeholder cells that don't need output
- Removed space character writing for continuation cells

**TextNode.ts** (6449abb):
- Removed forced default color application in `fromString()`
- Colors now properly inherit from parent (supports light/dark mode)
- Reverts the symptom-masking workaround

**TextNode.test.ts** (6449abb):
- Updated tests to reflect new behavior (no forced default colors)
- All 1957 tests pass ✓

## Previous Changes (448dbfe, d536896)

- Added comprehensive CJK rendering tests
- Added React testing utilities

## Testing

✅ UTF-8 width calculation: CJK width correctly calculated as 2
✅ CJK rendering: All existing CJK tests pass
✅ Full test suite: 1957 pass, 0 fail

## Impact

**This is a radical fix that addresses the actual renderer bug** rather than masking symptoms with forced colors. The approach:
- Removes the need for default color workarounds
- Properly supports light/dark mode via color inheritance
- Makes the renderer logic cleaner and more correct

Fixes #255

### DIFF
--- a/packages/core/src/renderables/TextNode.test.ts
+++ b/packages/core/src/renderables/TextNode.test.ts
@@ -589,6 +589,64 @@ describe("TextNodeRenderable", () => {
       expect(node.attributes).toBe(1)
     })
 
+    it("should apply default colors in fromString when not specified", () => {
+      const node = TextNodeRenderable.fromString("Test Text")
+
+      expect(node.children).toEqual(["Test Text"])
+      // Default fg: white (255, 255, 255, 255)
+      expect(node.fg?.r).toBe(1)
+      expect(node.fg?.g).toBe(1)
+      expect(node.fg?.b).toBe(1)
+      expect(node.fg?.a).toBe(1)
+      // Default bg: transparent black (0, 0, 0, 0)
+      expect(node.bg?.r).toBe(0)
+      expect(node.bg?.g).toBe(0)
+      expect(node.bg?.b).toBe(0)
+      expect(node.bg?.a).toBe(0)
+    })
+
+    it("should override default colors when explicitly provided in fromString", () => {
+      const node = TextNodeRenderable.fromString("Colored Text", {
+        fg: "#00ff00",
+        bg: "#0000ff",
+      })
+
+      expect(node.children).toEqual(["Colored Text"])
+      // Custom fg: green
+      expect(node.fg?.r).toBe(0)
+      expect(node.fg?.g).toBe(1)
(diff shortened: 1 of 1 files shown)


zenyr:
## Investigation Update

Attempted to fix CJK corruption by modifying `renderer.zig` to skip continuation cells instead of masking with forced colors.

**Finding**: Corruption persists without the default color fix.
- Japanese renders correctly ✅
- Korean/Chinese still corrupt ❌ (e.g., "한" → "í한")

**Conclusion**: The corruption root cause is deeper than continuation cell handling. Default colors in `TextNode.fromString()` work around it, but don't solve the underlying issue.

**Status**: Current PR (#256) with default colors is pragmatic and necessary for now. Root cause (likely ANSI positioning or buffer sync issues) requires deeper investigation.

zenyr:
CJK workaround: it "works" but this is not a right way to solve the issue.
Separated unit tests to #262 for future improvements.

Issue #229, opened 2025-10-23
Allow for <select> to have no currently selected option, or allow to hide that the option is selected when the <select> is not focused


pizzaisdavid:
I have four `<select>` elements side-by-side:

<img width="1483" height="749" alt="Image" src="https://github.com/user-attachments/assets/107d4932-656a-4dda-8fa5-e91a227cf4c9" />

In the screenshot, the second `<select>` is focused.

I would like the possibility to have no selected option when an select isn't focused.

Meaning: no triangle, no yellow text, no blue background.

An alternative could be that the selected option looks different when the `<select>` is not focused compared to when it is focused.

Below is my sample code. You can change the focused `<select>` using the left and right arrow keys.

```
import { render, useKeyboard } from "@opentui/react"
import type { SelectOption } from "@opentui/core"
import { useState } from "react"

function App() {
  const [focused, setFocused] = useState<0 | 1 | 2 | 3>(0)

  useKeyboard((key) => {
    if (key.name === "left") {
      if (focused === 0) {
        // none
      } else if (focused === 1) {
        setFocused(0)
      } else if (focused === 2) {
        setFocused(1)
      } else if (focused === 3) {
        setFocused(2)
      }
    } else if (key.name === "right") {
      if (focused === 0) {
        setFocused(1)
      } else if (focused === 1) {
        setFocused(2)
      } else if (focused === 2) {
        setFocused(3)
      } else if (focused === 3) {
        // none
      }
    }
  })

  const options: SelectOption[] = [
    { name: "Option 1", description: "Option 1 description", value: "opt1" },
    { name: "Option 2", description: "Option 2 description", value: "opt2" },
    { name: "Option 3", description: "Option 3 description", value: "opt3" },
  ]

  return (
    <box style={{ flexDirection: 'row', width: '100%' }}>
      <box style={{ border: true, height: '100%', width: '25%' }}>
        <select
          style={{ height: '100%', width: '100%' }}
          options={options}
          focused={focused === 0}
        />
      </box>

      <box style={{ border: true, height: '100%', width: '25%' }}>
        <select
          style={{ height: '100%', width: '100%' }}
          options={options}
          focused={focused === 1}
        />
      </box>

      <box style={{ border: true, height: '100%', width: '25%' }}>
        <select
          style={{ height: '100%', width: '100%' }}
          options={options}
          focused={focused === 2}
        />
      </box>

      <box style={{ border: true, height: '100%', width: '25%' }}>
        <select
[...]

PR #269, opened 2025-11-04
feat: add index control in the select component

aprogramq:
Issue #229

### DIFF
--- a/packages/core/src/renderables/Select.test.ts
+++ b/packages/core/src/renderables/Select.test.ts
@@ -617,7 +617,7 @@ describe("SelectRenderable", () => {
       expect(lastOption).toEqual(sampleOptions[2])
     })
 
-    test("should not emit events when movement is blocked", async () => {
+    test("should emit events even when movement is blocked", async () => {
       const { select } = await createSelectRenderable(currentRenderer, {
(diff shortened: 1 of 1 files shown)


kommander:
Cool

kommander:
/oc can you generate a test suite for the select component, covering the basics like we do for the Slider or Input for example?

kommander (approved):

kommander:
/oc there is a test failing in the pipeline for this PR, check it. can you make sure we clamp the index from the given config to length-1 of the options?

packages/react/src/types/components.ts
```diff
@@ -115,6 +115,7 @@ export type InputProps = ComponentProps<InputRenderableOptions, InputRenderable>
 
 export type SelectProps = ComponentProps<SelectRenderableOptions, SelectRenderable> & {
   focused?: boolean
+  selectedIndex?: number
```
msmps:
```suggestion
```

kommander:
/oc one more test failing. Check it. moveUp is emitting the event even though the actual selection did not change. moveDown does the same basically. Just leave it like that and adjust the test expectation to current behaviour for now.

PR #274, opened 2025-11-06
docs: add badges + showcasing

### LINKED ISSUE #252: Add showcase of projects built with OpenTUI
It would be great to add a showcase (perhaps in the readme) with projects built on top of OpenTUI. This could help demonstrate it's potential.

Full disclosure: I built [Red](https://github.com/evertdespiegeleer/red-cli) with OpenTUI, and would love to have it included in such a showcase 😉

msmps:
- closes #252


kommander:
The default-parsers is auto generated. Need to adjust the generator, but formatting is unnecessary as it would be overwritten again.

msmps:
> The default-parsers is auto generated. Need to adjust the generator, but formatting is unnecessary as it would be overwritten again.

Resolved! 🫡

PR #284, opened 2025-11-08
fix: Ensure correct screen before render loop

### LINKED ISSUE #283: bug: Does not switch to correct screen after suspend
Reproducible with opencode:

https://github.com/user-attachments/assets/58ed1ff4-9efa-47c7-8e41-02b7c165a015

**Cause:**
Editors like neovim, emacs, etc switch to the alternate screen. If this screen doesn't match the configured screen of the CliRenderer (determined by `useAlternateScreen`), the terminal clearing commands in performShutdownSequence are applied to the wrong screen.

**Fix:**
Ensure the correct screen (determined by `useAlternateScreen`) before starting the render loop, by emitting the appropriate ANSI sequences.

---
Related to https://github.com/sst/opencode/issues/3648.

veracioux:
Fixes #283.

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -533,6 +533,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     ) {
       this.updateScheduled = true
       process.nextTick(() => {
+        this.switchToCorrectScreen()
         this.loop()
         this.updateScheduled = false
       })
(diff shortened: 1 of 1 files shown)


kommander:
Hey, thanks for looking into this. Your proposed change would try to switch the screen for every frame basically. I think the correct solution to let the tui editor take over and come back would be making suspend/resume work correctly. I added my thoughts on that here: https://github.com/sst/opencode/pull/3983#issuecomment-3496101965

veracioux:
Gotcha, though isn't

PR #285, opened 2025-11-09
fix: Exception when selecting text on destroyed component

veracioux:
Fixes https://github.com/sst/opencode/issues/4099.

### DIFF
--- a/packages/core/src/tests/renderer.selection.test.ts
+++ b/packages/core/src/tests/renderer.selection.test.ts
@@ -1,0 +1,49 @@
+import { test, expect, beforeEach, afterEach } from "bun:test"
+import { createTestRenderer, type TestRenderer } from "../testing/test-renderer"
+import { TextRenderable } from "../renderables/Text"
+
+let renderer: TestRenderer
+let renderOnce: () => void
+
+beforeEach(async () => {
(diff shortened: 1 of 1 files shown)


kommander:
The `TextBufferView` is a low level API that wraps the FFI bindings and I made the guard throw so it can be caught, instead of causing a segfault when accessing destroyed native structs.

Trying to return something in some methods but not others would be inconsistent and can lead to hard to debug issues, because it hides the actual error.

Renderables are built on top of these lower level apis, which have to deal with destroyed objects. In this case the `Text` renderable is destroyed, which also destroys the underlying `TextBufferView`.

IMO the selection logic in `renderer.ts` needs to make sure it deals with destroyed objects and does not trigger selection methods on destroyed Renderables, as there are and might be more Renderables that handle selection, but do not use a `TextBufferView`, like the `ASCIIFontRenderable` for example.

I would try to reproduce this in a test for the renderer selection logic in `src/tests` and then make sure the logic does not try to update the selection on destroyed Renderables.

veracioux:
I see, I'll try to implement it that way.

veracioux:
@kommander Implemented according to your suggestion.
- Without the fixes, the test fails reproducibly
- Tested in opencode too, so the downstream issue should be fixed by this as well

PR #292, opened 2025-11-10
fix: Use mock Readable as stdin in test renderer

### LINKED ISSUE #276: testRender: user input leaks through
When I run tests in a terminal and then press some keys, they leak into the test.
The tests were set up using plain `await testRender(() => <App/>)`

For example, I ran a few tests and kept holding the `a` key. This is what happened:

Image

veracioux:
Instead of falling back to process.stdin in test renderer, create a mock `ReadableStream` instead.

All the APIs currently expect a `ReadStream`, but functionally a `ReadableStream` might be enough, I propose to try to change all `NodeJS.ReadStream` types to `NodeJS.ReadableStream` in the production code if possible.

Alternatively, we can keep the change localized to just the test renderer setup and use an `as ReadStream` type assertion to satisfy the typechecker.

Fixes #276.


veracioux:
@kommander What do you think?

kommander:
I am fine with this when tests pass and prettier is fixed

veracioux:
Done.

PR #297, opened 2025-11-11
linear scroll acceleration by default

kommander:


remorses:
What was the main issue with scroll acceleration on Mac? Did users complain? I wonder because I really like the new opencode scroll compared to the linear one.

PR #302, opened 2025-11-11
feat: improve lifecycle management

### LINKED ISSUE #248: Importing OpenTUI stops Effect from properly running teardown hooks
### Description

As the title says, simply importing `@opentui/core` disables all teardown hooks on Effect (or any other library, for that matter).

#### Example

```ts
import { Effect } from "effect";
import { BunRuntime } from "@effect/platform-bun";

// Import opentui with side-effects
import "@opentui/core";

const program = Effect.async<void, Error>((resume) => {
    console.log("Running effect...");

    // Complete the effect after 10s
    const id = setTimeout(() => {
        console.log("Effect completed normally!");
        resume(Effect.void);
    }, 10000);

    // Return an effect to run if the fiber is interrupted
    return Effect.sync(() => {
        console.log("Effect was interrupted!");
        clearTimeout(id);
    });
});

BunRuntime.runMain(program);
```

#### Expected behavior

If I press CTRL+C while this program is running, I would expect the logs to be as follows:

```
Running effect...
^CEffect was interrupted!
```

#### Observed behavior

If I press CTRL+C while this program is running, the log is as follows:

```
Running effect...
^C
```

### Proposed fixes

I've already determined that the cause is the following code snippet:

https://github.com/sst/opentui/blob/53017b68f41d544b6b0de6b3f72e7533941e3f99/packages/core/src/renderer.ts#L132-L138

I'm only speaking for myself here, but I was not expecting that importing `@opentui/core` would have these kinds of side effects, and I initially thought it was a problem with my `effect` code.

Here are some proposals for dealing with this issue:
 
1. Use `process.exitCode = 0;` instead of `process.exit()`.
  - This does not immediately cause bun to exit, and `effect` runs all teardown hooks, as expected. This is a simple fix that works for `effect`, and users can still use `process.on("beforeExit", ...)` and `process.on("exit", ...)`.
  - However, as far as I am aware, after setting `exitCode`, it is impossible to stop the runtime from terminating. As such, this might still be a problem for some users who might want to ignore SIGINT and some other signals.
  - Furthermore, this fix still doesn't eliminate side effects when importing `@opentui/core`.

2. OR: do not add the event handlers for the process exit signals.
  - If this is needed for some sort of cleanup, or to allow the process to be shut down when these signals are received, I would propose that a new option can be added to the `render` function, which would allow users to specify what to do when these signals are received.
  - This option could have as the default implementation `(signalType: ...) => process.exitCode = 0;`.

3. OR: put this singleton under an environment variable.

(I'd like to note that I'm not really familiar with the `opentui` codebase, and I don't know how much effort these ideas would take to implement. These are just what I thought of when writing this issue!)

Thanks!

limwa:
Fixes #248.

### Core
- Adds new `exitSignals` option to renderer, which specifies a list of signals that the renderer should destroy itself upon receiving.
  - The handlers are mounted when the renderer is created or resumed, and are unmounted when the renderer is suspended.
- The exit handler is now run on the `beforeExit` event. The `exit` event does not allow for more tasks to be added to the event loop, and thus the `Bun.sleep(200)` call would be effectively ignored, preventing the captured logs from being dumped.

### React
- Deduplicates repeated logic between the `createRoot` and `render` methods.
- Exposes the renderer on the `render` methods to make it easier to implement external lifecycle control.

### Solid

- Exposes the renderer on the `render` methods to make it easier to implement external lifecycle control.
- Applies the changes from #270 to the `@opentui/solid` package as well. This is helpful for lifecycle control because it inherently exposes the `renderer` (the user has to create one beforehand).
  - A `createRoot` method already exists in the `solid-js` package, which could create some confusion. Should I rename it? Open to suggestions!
  - Also updated all `@opentui/solid` examples and README.md to reflect the new API.

### DIFF
--- a/packages/solid/examples/index.tsx
+++ b/packages/solid/examples/index.tsx
@@ -1,4 +1,4 @@
-import { createRoot, render } from "@opentui/solid"
+import { createRoot } from "@opentui/solid"
 import { ConsolePosition, createCliRenderer } from "@opentui/core"
 import ExampleSelector from "./components/ExampleSelector"
 
--- a/packages/solid/examples/repro-filter-list.tsx
+++ b/packages/solid/examples/repro-filter-list.tsx
@@ -1,5 +1,5 @@
 import { createSignal, For } from "solid-js"
-import { createRoot, render, useRenderer } from "@opentui/solid"
+import { createRoot, useRenderer } from "@opentui/solid"
 import { createCliRenderer } from "@opentui/core"
 
 process.env.DEBUG = "true"
--- a/packages/solid/examples/repro-onSubmit.tsx
+++ b/packages/solid/examples/repro-onSubmit.tsx
@@ -1,5 +1,5 @@
 import { createSignal, Match, Show, Switch } from "solid-js"
-import { createRoot, render, useKeyboard, useRenderer } from "@opentui/solid"
+import { createRoot, useKeyboard, useRenderer } from "@opentui/solid"
 import { createCliRenderer } from "@opentui/core"
 
 process.env.DEBUG = "true"


packages/react/src/reconciler/renderer.ts
```diff
@@ -7,17 +7,13 @@ import { _render } from "./reconciler"
 /**
  * @deprecated Use `createRoot(renderer).render(node)` instead
  */
-export async function render(node: ReactNode, rendererConfig: CliRendererConfig = {}): Promise<void> {
+export async function render(
+  node: ReactNode,
+  rendererConfig: CliRendererConfig = {},
+): Promise<{ renderer: CliRenderer }> {
   const renderer = await createCliRenderer(rendererConfig)
-  engine.attach(renderer)
```
kommander, line 12: We somehow still need to get this attached for now, as animations with useTimeline won't work otherwise. 

This is another outstanding refactor, because the engine is currently global, but needs to be per renderer to support multiple renderers in the same process.

kommander:
Thanks for looking into this. This makes sense. Only thing I need to work with this is the `engine` for the Timeline. We could just let the renderer have an engine getter (engine was just for the lack of better naming, can be renamed), which lazily creates an engine instance and attaches it to the renderer and the `useTimeline` hook can use `renderer.engine` instead of the global one. Then it should work. I don't think that is test covered currently I am afraid.

packages/react/src/reconciler/renderer.ts
msmps: This is still being attached in the new `createRoot` method that I exposed [here](https://github.com/sst/opentui/blob/3dbe57087b5231278a6c6a2e849ad1f03fc0c310/packages/react/src/reconciler/renderer.ts#L32)... It looks like @limwa just updated this deprecated method to wrap that instead so this should be fine 👍 I'll be removing this deprecated render method in future versions

limwa: Yes, the code was duplicated between the `render` and the `createRoot` methods, so I just decided to rewrite the `render` method to use the `createRoot` method instead.

Probably out of the scope of the PR, so I can revert the change if you would prefer it.

limwa:
Also changed the `exitHandler` to run on `beforeExit` instead of `exit`. On `exit`, the capture dumping would not run because the runtime does not allow for more tasks to be scheduled on the event loop at that point. Using the `beforeExit` event fixes that problem!

packages/react/src/reconciler/renderer.ts
kommander: Ah I see it now, missed that, cool.

kommander:
Nice, I'll take it for a quick spin in the morning and check what I need to do for opencode, but this should be fine.

kommander:
Sorry I've broken this

limwa:
> Sorry I've broken this

I'll take a look at it tomorrow, no worries!

kommander:
@limwa is what @msmps did in #383 working for you?

limwa:
Oops, I forgot about this. I can't test it at the moment, but looking at the diff I imagine it will work just fine. One thing to note is that, unlike this PR, if `OTUI_DUMP_CAPTURE` is true and `CTRL + C` is pressed, it looks like the captures will be dumped twice (once for the signal handler and another time for the `beforeExit` handler).

kommander:
I need to fix `OTUI_DUMP_CAPTURE`, as it uses a timeout at shutdown, which may note get executed as anything async is just dropped at shutdown. Closing this.

PR #304, opened 2025-11-11
fix: refactor asciifont and fix seg fault on color setters

msmps:
## Summary

Refactors `ASCIIFont` to use `color` and `backgroundColor` properties instead of `fg`/`bg`, and updates examples and documentation.

## Changes

- **API Refactor**: `ASCIIFont` now accepts `ColorInput | ColorInput[]` for `color` and `ColorInput` for `backgroundColor`
- **Removed properties**: `fg` and `bg` options removed from `ASCIIFontOptions`
- **Updated examples**: All ASCII font examples migrated to new API
- **Documentation**: Added missing `<textarea>` and `<code>` components to README

## Breaking Changes

- `fg` and `bg` properties removed - use `color` and `backgroundColor` instead
- Color parsing handled internally - pass `ColorInput` types (string, RGBA, etc.) directly

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -9,7 +9,7 @@
     },
     "packages/core": {
       "name": "@opentui/core",
-      "version": "0.1.40",
+      "version": "0.1.41",
       "dependencies": {
         "bun-ffi-structs": "0.1.2",
         "jimp": "1.6.0",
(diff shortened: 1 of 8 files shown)


packages/core/src/lib/ascii.font.ts
```diff
@@ -294,7 +294,13 @@ export function renderFontToFrameBuffer(
             if (renderX >= 0 && renderX < width) {
               const fontChar = segment.text[charIdx]
               if (fontChar !== " ") {
-                buffer.setCellWithAlphaBlending(renderX, renderY, fontChar, segmentColor, bg)
+                buffer.setCellWithAlphaBlending(
+                  renderX,
+                  renderY,
+                  fontChar,
+                  parseColor(segmentColor),
+                  parseColor(backgroundColor),
```
kommander, line 302: ASCII font rendering is awfully slow, need to move that to native at some point. But for most cases it doesn't matter much as it is rendered once and then drawn from the buffer.

PR #320, opened 2025-11-17
fix: enable bracketed paste (and more) on win

Hona:
Sibling of https://github.com/sst/opencode/pull/4401
image
Unsure if there are more non windows things that should be platform agnostic?


kommander:
I think actually all the things currently in the `switch else` should be ran for windows as well. I think the os switch is a leftover from when there actually _was_ a real distinction, which was removed at some point and the switch just stayed because I didn't test anything on windows.

Hona:
@kommander would any of the kitty things etc break even though windows terminal doesn't support it? I assume they just would fail their ifs?

Hona:
Testing after the refactor I noticed no change:

image

kommander:
:+1: Unknown sequences are usually just ignored and don't yield a response so the features are not enabled.

PR #333, opened 2025-11-21
fix: Use layout index as fallback criterion for zIndex sorting

### LINKED ISSUE #332: bug: Stacking order does not respect layout order when zIndex is absent
Title. Originally discovered this in opencode when cycling between parent session and subagent session:

https://github.com/user-attachments/assets/39f729aa-3770-446f-8d76-0c2c04192461

Will have a PR ready shortly.

veracioux:
Fixes #332.

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -582,7 +582,13 @@ export abstract class Renderable extends BaseRenderable {
 
   private ensureZIndexSorted(): void {
     if (this.needsZIndexSort) {
-      this._childrenInZIndexOrder.sort((a, b) => (a.zIndex > b.zIndex ? 1 : a.zIndex < b.zIndex ? -1 : 0))
+      const layoutIndexes = new Map<Renderable, number>()
+      for (let i = 0; i < this._childrenInLayoutOrder.length; i++) {
+        layoutIndexes.set(this._childrenInLayoutOrder[i], i)
+      }
+      this._childrenInZIndexOrder.sort((a, b) =>
+        a.zIndex > b.zIndex ? 1 : a.zIndex < b.zIndex ? -1 : (layoutIndexes.get(a) ?? 0) - (layoutIndexes.get(b) ?? 0),
+      )
       this.needsZIndexSort = false
     }
   }


packages/core/src/Renderable.ts
```diff
@@ -582,7 +582,13 @@ export abstract class Renderable extends BaseRenderable {
 
   private ensureZIndexSorted(): void {
     if (this.needsZIndexSort) {
-      this._childrenInZIndexOrder.sort((a, b) => (a.zIndex > b.zIndex ? 1 : a.zIndex < b.zIndex ? -1 : 0))
+      const layoutIndexes = new Map<Renderable, number>()
+      for (let i = 0; i < this._childrenInLayoutOrder.length; i++) {
```
kommander, line 586: I know this only runs when something changes, but in a long list for example that can be frequent. It seems like the opencode issue is that the "copied to clipboard" notification should just have a higher zIndex?

veracioux: This is a fix for a general issue, not just for the opencode issue. Opencode is just the first place I discovered it. For example if you look at this code:
https://github.com/veracioux/opentui-332-repro/blob/main/index.tsx#L7

If you remove the `Show` element around the inner `box`, the issue isn't reproducible. I think this is confusing behavior and should be fixed.

If you are concerned about performance, would you be ok with updating this map inside Renderable.add function and only reading it in ensureZIndexSorted?

veracioux: @kommander Do you think it would be ok to implement it that way?

kommander: > would you be ok with updating this map inside Renderable.add function and only reading it in ensureZIndexSorted?

I would prefer something like that, yeah. 

I think I am failing to understand what the issue is actually. Isn't the "copied to clipboard" toast positioned absolutely? I wonder why the layout order should influence that then. 

Could we maybe have that issue reproduced in a test? Can be a snapshot test so ensure it is rendered in expected order.

veracioux: In browsers the expectation is that the stacking order is determined by zIndex, but if two elements have the same zindex (or undefined), then the element that comes later in the layout should be displayed over the element that comes before. Absoluteness has nothing to do with it.

Open this minimalistic [index.html](https://gist.github.com/veracioux/78d56c9530842810b0f04bed8fdd9a5d) in a browser to see what I mean. Btw, the behavior is the same with position: static instead of relative.

I think opentui should do the same.

I'll try to create a test reproducing the issue like https://github.com/veracioux/opentui-332-repro does.

PR #337, opened 2025-11-22
fix: Input inserting KeyEvent with modifiers as text

veracioux:
This is both an issue report and a fix for the same.

https://github.com/user-attachments/assets/fb472a65-00e8-4bc9-8b68-0f139f2c8d33

### DIFF
--- a/packages/core/src/renderables/Textarea.ts
+++ b/packages/core/src/renderables/Textarea.ts
@@ -228,6 +228,7 @@ export class TextareaRenderable extends EditBufferRenderable {
     const keyShift = typeof key === "string" ? false : key.shift
     const keyMeta = typeof key === "string" ? false : key.meta
     const keySuper = typeof key === "string" ? false : key.super
+    const keyHyper = typeof key === "string" ? false : key.hyper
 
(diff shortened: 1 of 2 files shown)


kommander:
/oc add a test for the Input renderable and implement this for Textarea and test cover it as well

kommander:
OMG

kommander:
@veracioux sorry I just wanted it to implement the same fix for the text area, it got that wrong.

veracioux:
@kommander No worries haha. I think for that it suffices to add `!keyHyper && !keySuper` to https://github.com/sst/opentui/blob/main/packages/core/src/renderables/Textarea.ts#L239

kommander:
Yes, that's what I wanted it to do and just test cover that with two small tests expecting the chars not to land in the text when modifiers are pressed.

kommander:
/oc revert what you did there. All I wanted was the fix applied to Textarea as well, so that keys with modifiers do not end up as input. And if Textarea does not have that issue, make sure it is test covered. Keep it simple. Don't make shit up.

kommander:
cleaned up in #375

PR #354, opened 2025-11-27
feat(terminal-palette): add tmux OSC sequence passthrough support

melMass:
- wrap OSC sequences for tmux using DCS format (ESC P tmux; ...)
- detect tmux environment and automatically wrap sequences when needed
- consolidate OSC writes through new writeOsc() method for cleaner passthrough handling

### DIFF
--- a/packages/core/src/lib/terminal-palette.ts
+++ b/packages/core/src/lib/terminal-palette.ts
@@ -57,13 +57,6 @@ function wrapForTmux(osc: string): string {
   return `\x1bPtmux;${escaped}\x1b\\`
 }
 
-/**
- * Check if running inside tmux
- */
-function isTmux(): boolean {
-  return !!process.env.TMUX
-}
-
 export class TerminalPalette implements TerminalPaletteDetector {
   private stdin: NodeJS.ReadStream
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/terminal-palette.ts
```diff
@@ -46,17 +46,45 @@ function toHex(r?: string, g?: string, b?: string, hex6?: string): string {
...
+function wrapForTmux(osc: string): string {
+  // Replace ESC with ESC ESC for tmux (escape the escape)
+  const escaped = osc.replace(/\x1b/g, "\x1b\x1b")
+  return `\x1bPtmux;${escaped}\x1b\\`
+}
+
+/**
+ * Check if running inside tmux
+ */
+function isTmux(): boolean {
```
kommander: Could it use `renderer.capabilities` here?

melMass: I didn't know about it, I moved it to the renderer and tested it still works

https://github.com/sst/opentui/blob/c2e2b7bfd13ed3596a1fc1ef35d67fa527c4538e/packages/core/src/renderer.ts#L1753-L1756

PR #356, opened 2025-11-28
Replace yoga-layout with bun-yoga

### LINKED ISSUE #355: Feature request: Remove top-level await to enable bytecode compilation
Bun's bytecode compilation reduces startup from ~520ms to ~44ms but requires no top-level await. This is blocking https://github.com/sst/opencode/issues/4843

There are 3 top-level await locations:

**yoga-layout WASM**

https://github.com/sst/opentui/blob/5dbe0b4be9b647f2d10ac90b63acdecf1a6998d8/packages/core/src/index.ts#L18

yoga-layout uses async `WebAssembly.instantiate()` internally. Consider using FFI to native Yoga instead. 

**Platform import**

https://github.com/sst/opentui/blob/5dbe0b4be9b647f2d10ac90b63acdecf1a6998d8/packages/core/src/zig.ts#L22

Can use sync `require()` instead.

**Native lib path resolution**

https://github.com/sst/opentui/blob/5dbe0b4be9b647f2d10ac90b63acdecf1a6998d8/packages/core/scripts/build.ts#L139-L142

Can use sync `require.resolve()` instead.

The platform import and native lib path are straightforward. Yoga is the main blocker.

remorses:
Replace yoga-layout (WASM) with bun-yoga (native FFI via Zig)

- Faster layout calculations using native code instead of WASM
- Supports Windows, Linux, and macOS

Closes #355

I created the bun-yoga package using Bun FFI: https://github.com/remorses/bun-yoga

### DIFF
--- a/packages/core/scripts/build.ts
+++ b/packages/core/scripts/build.ts
@@ -183,6 +183,7 @@ if (buildLib) {
   mkdirSync(distDir, { recursive: true })
 
   const externalDeps: string[] = [
+    ...Object.keys(packageJson.dependencies || {}),
     ...Object.keys(packageJson.optionalDependencies || {}),
     ...Object.keys(packageJson.peerDependencies || {}),
   ]


kommander:
How about I'll release a snapshot of this and we can test it on an opencode branch? Opencode branches can be used via `bunx opencode-ai@branch-name`.

remorses:
Sure. Let me know when it is released and I will test it

remorses:
Edit: reverted back to use import instead of commonjs to resolve the issues with bun compile in opencode

I wanted to use require instead of import to enable `--bytecode` and make faster startup but even after finding a way to include the libraries in a way that got included using import attributes `with { type: 'file' }` I found another Bun bug: `Expected CommonJS module to have a function wrapper`

This happens because some dependency in opencode tree cannot be converted to commonjs (which is what bytecode option does in compile)

~~I am trying to link my local opentui branch with opencode and there are Bun compile issues where it no longer finds the libs. Fixing these first~~

remorses:
I am using this branch together with https://github.com/sst/opencode/pull/4791 as my opencode daily driver. for now everything works alright

kommander:
I'll release a snapshot for opencode tomorrow then others can test as well.

kommander:
I created a branch `bun-yoga-pr-356`, from where I can release snapshots. Can you PR against that? I released a snapshot [0.0.0-20251201-fe4cc80e](https://www.npmjs.com/package/@opentui/core/v/0.0.0-20251201-fe4cc80e) and tried to use that in opencode, but getting this: 
Screenshot 2025-12-01 at 21 41 26
The dependency is missing in the bundle package.json, needs some adjustment in the `core/scripts/build.ts` I think.

kommander:
What is that crash in the tests? It doesn't fail on main.

remorses:
The issue is that yoga-layout let's you run some methods on destroyed nodes while instead I was throwing an error. Latest version is aligned with yoga-layout, it returns default values if freed

kommander:
Should bun-yoga be an external dependency? Otherwise it probably needs to have the binary packages as optional dependencies in the npm dist package it seems.
Screenshot 2025-12-02 at 12 47 40

remorses:
Yes external dependency should work. I am trying to make it work with Bun compile. require.resolve seems to not work in Bun. Will open a new PR later

kommander:
Screenshot 2025-12-02 at 20 54 48

Getting this for the latest snapshot [0.0.0-20251202-52296102](https://www.npmjs.com/package/@opentui/core/v/0.0.0-20251202-52296102) when using it in opencode.

remorses:
So the issue is that bun-yoga was getting bundled 

I updated the build script to externalize dependencies.

remorses:
bun-yoga still needs more testing. the build script stuff is too hard. will reopen a new PR when ready or when testing locally in opencode is easier

Issue #361, opened 2025-11-29
changes to package.json not being published


ericc-ch:
So #348 should be included in the latest release

<img width="983" height="376" alt="Image" src="https://github.com/user-attachments/assets/9553baec-466d-4423-b8b1-d245c39b3812" />

<img width="1523" height="348" alt="Image" src="https://github.com/user-attachments/assets/0e981e61-1a75-440e-bc76-31fc977dc666" />

But somehow the change to package.json isnt published to the latest version in npm

<img width="1205" height="834" alt="Image" src="https://github.com/user-attachments/assets/28ef9001-6a25-4995-a4ab-563d5f4d4768" />

Changes to other files are included tho, for example changes to readme from the same PR

<img width="853" height="264" alt="Image" src="https://github.com/user-attachments/assets/db2fb949-9f8c-4b5c-900d-e9ecbefb7dfa" />

Or #357 

<img width="1565" height="613" alt="Image" src="https://github.com/user-attachments/assets/2af5feae-0913-4990-ba73-9a64339a5fc4" />

<img width="584" height="433" alt="Image" src="https://github.com/user-attachments/assets/64f55dd9-1b49-45ba-9965-462d710d7815" />

ericc-ch:
Okay while writing this I did some digging and the build script doesnt pickup the exports field in package.json so my last commit basically did nothing
Well now it should actually probably definitely be exported

msmps:
closed by #362

PR #362, opened 2025-11-29
Actually properly export solid plugin in build script

ericc-ch:
My stupid ass didn't check the build script for #348 
Sorry lol


ericc-ch:
Closes #361

PR #363, opened 2025-11-29
fix: key repeat breaking change from PR #357

melMass:
## Problem

PR #357 introduced a breaking change: key repeats were emitted as separate `keyrepeat`  events instead of  `keypress`

This broke components like `Textarea` that only subscribe to  `keypress`.   
They stopped receiving key repeats, so holding a key no longer triggered continuous input.

## Solution

Following the @kommander 's suggestion: emit repeat events as  `keypress`  with a `repeated: true`  flag


kommander (approved):
Thanks!

PR #366, opened 2025-11-30
fix: tilde-terminated CSI sequences

melMass:
Adds parsing for tilde-terminated CSI sequences with event types (`CSI number;modifiers:event ~`).

This enables press/repeat/release detection for PageUp, PageDown, Insert, Delete, Home, End, and F5-F12 keys which
use a different format than arrow keys and F1-F4.

## Changes

- Added `tildeKeyMap` for tilde key number → name mapping
- Added `parseKittyTildeKey()` function
- Added tests for tilde keys with modifiers and event types

### DIFF
--- a/packages/core/src/renderables/TabSelect.ts
+++ b/packages/core/src/renderables/TabSelect.ts
@@ -103,11 +103,11 @@ export class TabSelectRenderable extends Renderable {
   }
 
   private refreshFrameBuffer(): void {
-    if (!this.frameBuffer || this._options.length === 0) return
-
+    if (!this.frameBuffer) return
     // Use focused colors if focused
     const bgColor = this._focused ? this._focusedBackgroundColor : this._backgroundColor
     this.frameBuffer.clear(bgColor)
+    if (this._options.length === 0) return
 
     const contentX = 0
     const contentY = 0


packages/core/src/lib/parse.keypress-kitty.ts
```diff
@@ -222,11 +246,83 @@ function parseKittyFunctionalKey(sequence: string): ParsedKey | null {
...
 
+/**
+ * Parse Kitty keyboard protocol tilde keys with event type
+ * Format: CSI number;modifiers:event_type ~
+ * Examples:
+ *   \x1b[5;1:1~ = pageup press
+ *   \x1b[5;1:3~ = pageup release
+ *   \x1b[3;5:1~ = ctrl+delete press
+ */
+function parseKittyTildeKey(sequence: string): ParsedKey | null {
```
kommander: I think that method can be combined with the functional parsing as it is the same format just with tilde, so the regex can detect both, otherwise it seems like duplicate logic?

melMass: Definitely! It was easier to reason about separately while iterating but it can be combined

PR #367, opened 2025-11-30
feat(core): emit capabilities event on terminal capability response

melMass:
As far as I can tell the only way to get the end of the detections sequence:

1. `\x1b[?1016;2$y` - sgr_pixels
2. `\x1b[?2027;0$y` - unicode query
3. `\x1b[?2031;2$y` - color_scheme_updates
4. `\x1b[?1004;2$y` - focus_tracking
5. `\x1b[?2004;2$y` - bracketed_paste
6. `\x1b[?2026;2$y` - sync
7. `\x1b[1;2R` - explicit_width (CPR)
8. `\x1b[1;3R` - scaled_text (CPR)
9. `\x1bP>|kitty(0.42.2)\x1b\\` - xtversion
10. `\x1b[?0u` - kitty keyboard

Is by using a timeout or similar, this just adds an event you can subscribe too. At my app layer I count it and settle at 10 but I'm not sure of all the implications in adding that to core but it could be a nice addition too like `capabilities:ready`

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -197,7 +197,7 @@ yogaConfig.setPointScaleFactor(1)
 export abstract class Renderable extends BaseRenderable {
   static renderablesByNumber: Map<number, Renderable> = new Map()
 
-  private _isDestroyed: boolean = false
+  protected _isDestroyed: boolean = false
   protected _ctx: RenderContext
   protected _translateX: number = 0
   protected _translateY: number = 0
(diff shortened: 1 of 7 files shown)


kommander:
That makes sense, I wanted something like that, then apps and upcoming Renderables like Image can react to capabilities changes and don't have to wait at and delay the startup.

packages/core/src/lib/terminal-capability-detection.test.ts
```diff
@@ -132,3 +132,68 @@ describe("real-world terminal capability sequences", () => {
...
+      "\x1b[?2026;2$y", // 6. sync
+      "\x1b[1;2R", // 7. explicit_width (CPR)
+      "\x1b[1;3R", // 8. scaled_text (CPR)
+      "\x1bP>|kitty(0.42.2)\x1b\\", // 9. xtversion (triggers kitty detection)
+      "\x1b[?0u", // 10. kitty keyboard query
+    ]
+
+    for (const response of kittyResponses) {
+      // @ts-expect-error - accessing private for test
+      renderer._stdinBuffer.process(Buffer.from(response))
```
kommander: Thanks for adding a test! I need to be nitpicky here, I'd input the sequences on renderer.stdin like the test input mock utils do, because the _stdinBuffer is an implementation detail that might change.

melMass: I was looking for this.
Done!

PR #368, opened 2025-11-30
feat(diff): add word-level highlighting for modified lines

remorses:
Adds word-level highlighting to show which words changed within modified lines (like GitHub's diff view).

**Algorithm:**
1. Parse diff and group consecutive remove/add lines into change blocks
2. Pair lines positionally within each block (1st remove with 1st add, etc.)
3. For each pair, compute similarity score using character-level diff
4. If similarity >= 40%, compute word boundaries with `diffWordsWithSpace`
5. Pass highlight column ranges to LineNumberRenderable for rendering

Complexity is O(L + B×P×C²) where L=lines, B=change blocks, P=pairs per block, C=chars per line. In practice C is small (~100) and only change blocks are processed.

**Performance guards:** skips word diff for dissimilar lines, skips entirely for large blocks (>50 lines).

Screenshot 2025-11-30 at 18 01 50

**New options:**
- `disableWordHighlights` - Turn off word-level highlighting (default: `false`)
- `addedWordBg` - Background color for added words (default: brighter `addedBg`)
- `removedWordBg` - Background color for removed words (default: brighter `removedBg`)
- `lineSimilarityThreshold` - Min similarity (0-1) to pair lines for word diff (default: `0.4`)

**Also includes:**
- Fix for word highlights scrolling with content when `wrapMode: "none"` and horizontal scrolling is used


kommander:
This would be cool. I'd make it an option though so both view styles work. 

I am concerned about multi width graphemes though, it cannot use String.length or utf16 char offsets from js as the highlights operate on display width currently.

That's a bit tricky. What could work to get content start and end for the highlight offsets would be to measure the string before the content starts with Bun.stringWidth, then the content that should be highlighted itself with Bun.stringWidth and add that to the former for the end offset.

remorses:
Also works with text wrapping:

Screenshot 2025-12-14 at 18 22 01

remorses:
Here is how it looks in opencode

Screenshot 2025-12-16 at 17 50 23

Hona (changes requested):
on pure aesthetics i think it shouldnt be green it should be a more opaque version of the + bg color
thinking about themes and how github does it:

image

you may have already covered that in config right as a new option?
maybe the config you've already done is the way to go

remorses:
I now increase opacity, previously it was only increasing brightness

remorses:
Conflicts resolved

kommander:
I just didn't move forward yet with this because of performance concerns. There are a lot of reports of laggy performance in opencode and it us unclear what parts are opentui and what opencode tui usage or the server. I want't the performance to be sorted out and stable before introducing potential bottlenecks.

ibrahimsag:
Tested this with opencode, word highlighting works.

remorses:
I changed the word highlights algo to be the exact same as GitHub Desktop app, now it uses prefix and postfix checks instead of doing a word diff on the lines, it is faster and also looks better

remorses:
I have been using this branch in [critique](https://critique.work/) for a week and it works really well for me. I had to switch back to npm opentui version because of a [Bun bug](https://github.com/oven-sh/bun/issues/17946) with pkg.pr.new installs. It would be great to merge, let me know if you need any help

remorses:
Added a benchmark comparing diff performance with/without word highlights: `bun run bench:diff`

Results on a 2352-line diff (50 hunks, 20 changes each):
- With word highlights: ~2.7ms
- Without: ~2.2ms
- Overhead: ~20%

Absolute overhead is <1ms even for large diffs. Also added ASCII fast-path optimization to skip `Bun.stringWidth()` for non-Unicode content.

PR #369, opened 2025-11-30
feat: add automatic markdown table alignment

remorses:
Automatically formats markdown tables by padding columns to align content when rendering. Using tree sitter AST

### DIFF
--- a/packages/core/src/lib/tree-sitter/table-formatting.test.ts
+++ b/packages/core/src/lib/tree-sitter/table-formatting.test.ts
@@ -49,7 +49,9 @@ async function renderTable(markdown: string, conceal: boolean = true): Promise<s
 
   renderer.root.add(code)
   await renderOnce()
-  // Wait for async highlighting to complete
+  // Wait for async highlighting to complete (longer timeout for CI)
+  await new Promise((resolve) => setTimeout(resolve, 500))
+  await renderOnce()
   await new Promise((resolve) => setTimeout(resolve, 100))
   await renderOnce()
 


remorses:
Here is how table rendering looks like now. Formatting is handled both in conceal and non conceal mode using treesitter AST

I added a lot of inline snapshots so you can see how the result looks like. There is a small delay because treesitter parsing runs in the background asynchronously

Screenshot 2025-12-19 at 01 35 58

remorses (commented):
Regarding the Copilot suggestions:

1. **Double-delete issue**: Not applicable - when we delete `tree` and create `newTree`, we return early after deleting `newTree`. The finally block only runs for non-markdown paths.

2. **Negative width/padding**: Not applicable - column widths are initialized to minimum 3 with `fill(3)`, so `width - 2` is always >= 1.

3. **Caching suggestion**: References old code that was already refactored to use tree-sitter AST for cell parsing.

No changes needed.

kommander:
You'll probably hate me for this 😬, but I don't like that the formatting happens in the parser worker, that it is type aware and that it re-parses after formatting. I think the formatting should just happen before the markdown source is handed to the parser. I could see it as CodeRenderable `transform` prop, were the transform method gets the type and content and can format it before it is set as content and before it is handed to the parser.

kommander:
Markdown parsing perspectively: I don't want to use this kind of markdown parsing in opencode long term. I want to have a `StreamingMarkdownRenderable` that used default components for blocks that can be overridden with custom components, so it would also use a proper `TableRenderable`.

The flow would be kind of:
- hold a raw markdown buffer
- take markdown chunks
- detect a block
  - if full block render that
  - use `CodeRenderable` for fenced code blocks and `TableRenderable` for table etc.
- shift and write rest that is not a detected block yet to raw markdown buffer

That will allow showing fenced code blocks for example with custom background/borders/whatever and it does not have to re-parse the full markdown for every update it gets, only the last block worst case.

franlol:
I previously implemented this same solution at Opentui level, using Tree-sitter, and it produced really good results:

Screenshot_2025-12-02_at_14 56 07



However, I ultimately had to delete the PR because it felt more like a hack than a true solution. Also, this is parsing the text on each render.

Instead, I moved to a more robust approach and developed a parser at the plugin level that parses the text only once once the `text.complete` is triggered: [OpenCode MD Table Formatter](https://github.com/franlol/opencode-md-table-formatter)

remorses:
Would you implement the markdown renderable in Zig or TypeScript?

I thought about using marked npm package which exposes a fast lexer to find top level AST nodes, then I would use marked parser to render the nodes via opentui renderables

Tables could even use CSS grid when it is in Yoga so that they render well even if there is no space horizontally by wrapping the text

I don't like much the plugin @franlol because it uses regex to parse the markdown, which probably have unknown bugs and edge cases

kommander:
@remorses For now it is only possible to do that in typescript. Using the marked package seems fine, as long as it does not re-parse the whole text for every incoming chunk in the stream, but incremental.

I think something like @franlol did works well, as markdown is a really simple, small and very well defined set. Simple compared to html or a programming language. Not that much that can go wrong. We should have that formatting by default in opencode. Can we add it as a default plugin maybe?

remorses:
Markdown parsing is more complex than it looks trust me, using marked is totally worth it

PR #372, opened 2025-12-01
fix(core): wrap line-home/end at line boundaries, cmd-left and cmd-right do not get stuck at line start and line end on macOS

remorses:
- gotoLineHome at col 0 now wraps to end of previous line
- gotoLineEnd at EOL now wraps to start of next line
- Stays at buffer boundaries (first/last line)

Matches macOS Cmd+Left/Right behavior - you can hold the key to quickly navigate through lines.


kommander:
/oc run `bun prettier:write` in the repo and commit the prettier fixes

kommander:
/oc run prettier write with the exact version in the root package.json, otherwise the formatting will be off.

PR #373, opened 2025-12-01
fix(OSC4): tmux 3.6 has native support for OSC4 and should not be wrapped

noirbizarre:
#354 fix was only working for tmux<3.6 because it doesn't have native OSC4 support, so OSC4 sequence needed escaping.
But, tmux 3.6 has native OSC4 support and wrapping the sequence in a tmux escape prevent OSC4 to work with tmux 3.6.

This fixes sst/opencode#3688 for tmux 3.6 (regression was introduced by upgrading to OpenTUI 0.1.51 in https://github.com/sst/opencode/commit/204a31b6bbad951e1e03d985c2edc62165062a4e.


melMass:
Ohhh tmux was updated [last week](https://raw.githubusercontent.com/tmux/tmux/3.6/CHANGES), this explains my confusion! I introduced that bug my bad!

noirbizarre:
It's OK: I did my first OpenCode/OpenTUI contrib and I've learned about OSC sequences 😉

PR #374, opened 2025-12-02
Replace yoga-layout with bun-yoga, snapshot PR

remorses:
Replace yoga-layout (WASM) with bun-yoga (native FFI via Zig)

This new bun-yoga version should be statically analyzable by bun now


remorses:
Currently I make yoga operations fail if they are applied to a destroyed yoga node

Should I silently ignore them instead to make the test pass?

kommander:
Doesn't yoga wasm fail for operations on destroyed nodes? It probably should be the same behaviour as wasm to be a drop-in replacement.

PR #376, opened 2025-12-02
bun-yoga updates

remorses:
- Update bun-yoga from 0.2.11 to 0.2.12

This should fix the Bun compile issue. 

Using NAPI would make this so much simpler. But there is a segmentation fault in windows because of a Bun bug.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,9 +4,6 @@
   "workspaces": {
     "": {
       "name": "@opentui",
-      "dependencies": {
-        "bun-yoga": "^0.2.12",
-      },
       "devDependencies": {
         "prettier": "3.6.2",
       },
--- a/package.json
+++ b/package.json
@@ -24,8 +24,5 @@
   "prettier": {
     "semi": false,
     "printWidth": 120
-  },
-  "dependencies": {
-    "bun-yoga": "^0.2.12"
   }
 }


bun.lock
```diff
@@ -4,6 +4,9 @@
   "workspaces": {
     "": {
       "name": "@opentui",
```
kommander, line 6: Why does it need to be in the root package.json?

remorses: sorry fixed

kommander:
NAPI would void much of the performance gain I think. At some point I want to move the layout to native opentui and handle the renderable tree there, to thin out the typescript layer, so that is the correct path imo.

remorses:
I tried rewriting bun-yoga using NAPI and it was actually a bit faster

FFI (ReleaseFast) - 
| Test | Avg Time | vs WASM |
|------|----------|---------|
| Simple Layout | 0.0022ms | 2.41x faster |
| Column 100 | 0.1177ms | 3.13x faster |
| Column 500 | 0.5365ms | 3.29x faster |
| Nested 4x3 | 0.2679ms | 2.42x faster |
| Nested 4x4 | 0.6974ms | 2.51x faster |
| Overall | | 2.75x faster |


NAPI (ReleaseFast) - final results:
| Test | Avg Time | vs WASM |
|------|----------|---------|
| Simple Layout | 0.0014ms | 4.63x faster |
| Column 100 | 0.1457ms | 2.59x faster |
| Column 500 | 0.5893ms | 3.04x faster |
| Nested 4x3 | 0.3026ms | 2.12x faster |
| Nested 4x4 | 0.7611ms | 2.23x faster |
| Overall | | 2.92x faster |

remorses:
Opencode still can't find the file. I will give another try tomorrow. Or just switch to NAPI if bun makes a new release soon fixing the NAPI windows bug

PR #379, opened 2025-12-03
Yoga ffi snapshot

remorses:


remorses:
I managed to test this locally using the submodule solution. Using local linking script was causing strange segfaults and issues like duplicate core packages.

remorses:
It finally works with opencode! 🎉 Everything seems to work fine.

PR #386, opened 2025-12-05
Preliminary image support

acoliver:
Add experimental terminal image rendering support with pixel sizing and React exposure.

Highlights:
- New Image renderable that accepts pixelWidth/pixelHeight and fit modes (contain/cover/fill) with sharp-based scaling.
- Renderer loads/caches images, emits kitty/iTerm2 graphics escapes, deletes kitty ids on move/resize, and surfaces pixelResolution via capabilities and a pixelResolution event.
- React type exports and components include Image; tests cover graphics protocol parsing and renderer image path.

Notes:
- Native build/tests not run here; JS tests only.
- Pixel resolution query/response handled in JS; header sizing in downstream apps should use the exposed resolution and avoid hardcoded px-per-cell fallbacks.

Addresses #387


acoliver:
converting to draft while I figure out this bun crash issue.

packages/core/src/graphics/protocol.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+registerEnvVar({
+  name: "OTUI_PREFER_KITTY_GRAPHICS",
+  description: "Force-enable kitty graphics protocol when available.",
+  type: "boolean",
+  default: false,
+})
+
+export function detectGraphicsSupport(): GraphicsSupport {
```
remorses, line 17: I think there is a better way to detect kitty graphics support, there should be already something in the codebase

simonklee:
Image support added in main.

PR #389, opened 2025-12-05
Fix ScrollBox scissor clipping for nested scrollboxes

### LINKED ISSUE #388: Nested scrollbox content not clipped (inner scrollbox text overdraws outside container)
### Summary
Nested scrollbox content isn’t clipped: an inner scrollbox’s text draws outside its border and into the outer scrollbox (and even into the header area in our app). Layout sizes/borders render correctly, but overflow clipping doesn’t apply.

### Steps to Reproduce
1. Render an outer `scrollbox` with a fixed height (e.g., 15), `border: true`, `overflow: "hidden"`, `scrollY: true`.
2. Inside it, render a single child: an inner `scrollbox` with a fixed height (e.g., 8), `border: true`, `overflow: "hidden"`, `scrollY: true`, containing 20+ `<text>` lines.
3. Scroll the inner (and/or outer).

### Expected
Inner scrollbox content is clipped to its viewport; no text is drawn outside its border or the outer scrollbox.

### Actual
Inner scrollbox text overdraws past its border and bleeds into the outer scrollbox (and other UI). Clipping appears to be ignored for nested scrollboxes.

### Notes
- Observed via the React renderer on macOS/iTerm2/kitty; outer scrollbox also had `overflow: "hidden"`.
- A screenshot showing the bleed is available; happy to attach if helpful.
- A minimal repro should be possible with the steps above; this looks like a renderer clipping issue rather than a layout sizing issue.

acoliver:
Fixes #388.

## Summary
- intersect scissor rectangles when nesting ScrollBoxes to prevent children from drawing outside parents
- remove extra padding on ScrollBox viewport to avoid bleed
- add core and React regression tests covering nested scrollboxes and stacked tool blocks

## Testing
- bun test packages/core/src/tests/scrollbox.test.ts
- bun test packages/react/tests/layout.test.tsx

### DIFF
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -33,8 +33,8 @@ class ContentRenderable extends BoxRenderable {
 
   protected _getVisibleChildren(): number[] {
     if (this._viewportCulling) {
-      return getObjectsInViewport(this.viewport, this.getChildrenSortedByPrimaryAxis(), this.primaryAxis, 0).map((child) =>
-        child.num,
+      return getObjectsInViewport(this.viewport, this.getChildrenSortedByPrimaryAxis(), this.primaryAxis, 0).map(
+        (child) => child.num,
       )
     }
     return this.getChildrenSortedByPrimaryAxis().map((child) => child.num)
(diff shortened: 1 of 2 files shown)


acoliver:
before:

image


after:

image

you'll also find a test.

PR #390, opened 2025-12-06
fix(diff): add fg prop to support light mode themes

code-yeongyu:
Fix unreadable diff text on light mode themes by adding `fg` prop.

## Problem
When using DiffRenderable with light mode themes, syntax-unhighlighted tokens render as white text (hardcoded default in CodeRenderable), making them invisible on light backgrounds.
This is problematic on OpenCode light theme diff editor where no syntax highlighting is supported- the letters are all rendered as white.

## Solution
Add `fg` prop to DiffRenderable, allowing consumers to pass a foreground color that propagates to underlying CodeRenderables.

## Changes
- Add `fg` option to DiffRenderableOptions
- Pass `fg` to CodeRenderables on construction and updates
- Add getter/setter for dynamic updates
- Add 7 tests

### DIFF
--- a/packages/core/src/renderables/Diff.test.ts
+++ b/packages/core/src/renderables/Diff.test.ts
@@ -2479,3 +2479,191 @@ test("DiffRenderable - line numbers update correctly after resize causes wrappin
   leftCodeRenderable.off("line-info-change", lineInfoChangeListener)
   renderer.destroy()
 })
+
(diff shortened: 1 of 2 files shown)


kommander:
I wonder if the diff renderable and code renderable should just take the default fg/bg from the `default` syntax style if given, so there wouldn't be two ways to set a default. But this fixes the opencode issue when opencode then also sets fg, so I'd go with it and refactor eventually.

PR #396, opened 2025-12-09
allow shift+space insert as space char

### LINKED ISSUE #380: bug: shift+space doesn't insert a space on wezterm
## Summary
Input components do not accept shift+space key combinations when using the "wezterm" terminal. The key combination works correctly in other terminals like Ghostty. Often times when in the flow of typing, the shift key can sometimes remain pressed for longer than expected, resulting in a missed space.

## Environment
- **Terminal**: wezterm
- **OS**: Linux (and likely others)
- **Package**: @opentui/core

## Steps to Reproduce
1. Run any example with Input components (e.g., Input & Select layout demo)
2. Focus on an input field
3. Press shift+space
4. Observe that no space character is inserted

## Expected Behavior
shift+space should insert a space character in the input field, just like regular space.

## Actual Behavior
shift+space is ignored and no character is inserted.

## Root Cause
wezterm sends the ModifyOtherKeys escape sequence `\u001b[27;2;32~` for shift+space, which the Input component's text insertion logic does not handle. The component only checks for single-character sequences, but ModifyOtherKeys sequences are multi-character escape sequences.

## Technical Details
- **wezterm sequence**: `\u001b[27;2;32~` (ModifyOtherKeys protocol)
- **Parsed result**: `{ name: "space", sequence: "\u001b[27;2;32~", shift: true }`
- **Issue**: Input component checks `keySequence.length === 1`, which fails for escape sequences
- **Working terminals**: Ghostty sends `\u001b[32;2u` (Kitty protocol) which works correctly

## Possible Fix
Update the Input component's `handleKeyPress` method to handle ModifyOtherKeys sequences for printable characters, specifically the space character case:
```typescript
// Handle ModifyOtherKeys sequences (e.g., wezterm shift+space)
else if (typeof key !== "string" && key.name === "space" && !key.ctrl && !key.meta && !key.super && !key.hyper) {
  this.insertText(" ")
  return true
}
```

but this feels like an ugly hack so idk! I suppose the other workaround would be to just change the wezterm config and handle shift+space differently.

kommander:
Fixes #380

### DIFF
--- a/packages/core/src/renderables/Input.ts
+++ b/packages/core/src/renderables/Input.ts
@@ -327,25 +327,21 @@ export class InputRenderable extends Renderable {
       }
     }
 
-    // Handle space key specifically (including shift+space)
-    if (keyName === "space" && !keyCtrl && !keyMeta && !keySuper && !keyHyper) {
-      this.insertText(" ")
-      return true
-    }
+    if (!keyCtrl && !keyMeta && !keySuper && !keyHyper) {
+      if (keyName === "space") {
+        this.insertText(" ")
+        return true
+      }
 
-    // Only insert text if no modifier keys are pressed (except shift for uppercase)
-    if (
-      keySequence &&
-      keySequence.length === 1 &&
-      keySequence.charCodeAt(0) >= 32 &&
-      keySequence.charCodeAt(0) <= 126 &&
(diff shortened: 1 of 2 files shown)


phall1:
You guys rock

PR #401, opened 2025-12-10
feat(core): add opacity property to Renderable

### LINKED ISSUE #393: Add opacity to box
This is just a wishlist thing, but it would be really nice for Box to have an opacity that effects everything inside of it. I tracked things down to drawBox in buffer.zig but it gets dicey after that. Any advice on how this could, or even IF it could be implemented?

muhammedaksam:
Fixes #393

## Summary
- Add `opacity` property (0.0–1.0) to the base `Renderable` class
- Implement opacity stack in the native Zig buffer layer for efficient alpha blending
- Nested renderables inherit compounded opacity (parent × child)
- Add opacity render commands (`pushOpacity` / `popOpacity`) to the render pipeline

## Changes
- `buffer.zig`: Add `opacity_stack` with push/pop/get/clear functions
- `lib.zig`: Add FFI exports for opacity stack operations
- `zig.ts`: Add FFI bindings and `RenderLib` interface methods
- `buffer.ts`: Add TypeScript wrapper methods
- `Renderable.ts`: Add `opacity` property, render commands, stack management
- `examples/opacity-example.ts`: Core example with animation
- `examples/opacity.tsx`: React example with animation
- `tests/opacity.test.ts`: Added 8 unit tests

## Testing
```bash
bun test packages/core/src/tests/opacity.test.ts
bun run packages/core/src/examples/opacity-example.ts
bun run packages/react/examples/opacity.tsx
```

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -448,7 +448,7 @@ export abstract class Renderable extends BaseRenderable {
     this.parent?.propagateLiveCount(delta)
   }
 
-  public handleKeyPress?(key: KeyEvent | string): boolean
+  public handleKeyPress?(key: KeyEvent): boolean
   public handlePaste?(event: PasteEvent): void
 
   public findDescendantById(id: string): Renderable | undefined {
--- a/packages/core/src/editor-view.test.ts
+++ b/packages/core/src/editor-view.test.ts
@@ -593,4 +593,344 @@ describe("EditorView", () => {
       expect(view.getVisualCursor().visualCol).toBe(8)
     })
   })
+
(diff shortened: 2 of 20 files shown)


kommander:
This looks really good, thank you. The box uses `.fillRect` for the background which probably also needs to respect the opacity stack then? The text seems to work though. 
Screenshot 2025-12-11 at 14 17 40

muhammedaksam:
Thank you. I've updated it to check the current opacity level, so it should properly respect inherited opacity now.

image

> This looks really good, thank you. The box uses `.fillRect` for the background which probably also needs to respect the opacity stack then? The text seems to work though. Screenshot 2025-12-11 at 14 17 40

PR #404, opened 2025-12-11
fix(core): Allow word-forward and word-backword for ctrl+left and ctrl+right (linux)

MilesNorton:
A fix for https://github.com/sst/opentui/pull/404

### DIFF
--- a/packages/core/src/renderables/Textarea.ts
+++ b/packages/core/src/renderables/Textarea.ts
@@ -89,6 +89,8 @@ const defaultTextareaKeybindings: KeyBinding[] = [
   { name: "b", meta: true, action: "word-backward" },
   { name: "right", meta: true, action: "word-forward" },
   { name: "left", meta: true, action: "word-backward" },
+  { name: "right", ctrl: true, action: "word-forward" },
+  { name: "left", ctrl: true, action: "word-backward" },
   { name: "f", meta: true, shift: true, action: "select-word-forward" },
   { name: "b", meta: true, shift: true, action: "select-word-backward" },
   { name: "right", meta: true, shift: true, action: "select-word-forward" },


MilesNorton:
I haven't tested this yet but I'll try to get to this asap

MilesNorton:
tested and works. 

Delving into it deeper and I can see that it allows keybind overrides. I do still like the idea of having it here still but It does mean that Mac keyboards have 2 ways of navigating this now.

remorses:
I think opencode uses ctrl arrows for switching tasks right now, so this would conflict with that. not sure if opencode is able to override it or not

MilesNorton:
Thanks @remorses I forgot I removed that keybind. I guess we will have to get opencode to allow these fields in the keybinds config

kommander:
I am currently trying to create a more opinionated and overloaded default mapping, covering as many cases as possible and will allow to custom map _all_ the text area actions in the opencode config.

MilesNorton:
Thanks @kommander, I've opened a pr up in the opencode repo so that we could get this in but happy for it to be delayed. Let me know if I can help anywhere

kommander:
`ctrl+arrows letf/right` should not be mapped by default in opencode for anything, because on macos that is preserved by the OS to switch desktops. I brought the word forward/backward with `ctrl+left/right` in as default for the textarea in #407, on macos it will then just not do anything because the OS captures it. @MilesNorton I still need to figure out some defaults from different expectations, I went through the issues as far as I could, but probably missed something. If you can think of any other defaults that are missing that would help. I'll push the config bindings for textarea actions in opencode shortly.

MilesNorton:
very nice @kommander - appreciate it. I'll keep an eye out for the other defaults

PR #410, opened 2025-12-12
Move 3D module to separate @opentui/3d package

### LINKED ISSUE #408: Move 3D module to separate @opentui/3d package
Installing @opentui/core pulls in 84 packages including three, planck, bun-webgpu, and rapier due to optionalDependencies.

Most users only need the core TUI functionality. Moving src/3d/ to a separate @opentui/3d package would keep the core package lightweight and let users opt-in to 3D features explicitly.

remorses:
Extracts the 3D rendering module from @opentui/core into a separate @opentui/3d package. This keeps the core package lightweight - users who don't need 3D features won't have three, planck, rapier, and bun-webgpu installed as optional dependencies.

Note: build and publish scripts were copied from other packages. These could be simplified in a future PR for all packages.

Fix #408

### DIFF
--- a/packages/react/src/reconciler/reconciler.ts
+++ b/packages/react/src/reconciler/reconciler.ts
@@ -18,8 +18,6 @@ export function _render(element: React.ReactNode, root: RootRenderable) {
     "",
     console.error,
     console.error,
-    // @ts-expect-error the types for `react-reconciler` are not up to date with the library.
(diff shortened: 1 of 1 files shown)


kommander:
Could it be that the prebuilt examples now don't contain the 3d examples anymore?

remorses:
Yeah. Then it's better to keep examples in core. What is the use case for building the examples?

kommander:
Letting users see what it can do without having to build it themselves and to provide debug tools like the now built in key handling debug demo.

PR #413, opened 2025-12-13
fix: absolute positioned elements now position relative to parent

### LINKED ISSUE #406: Relative and absolute positions do not work properly
Seems like it's a little broken. 

```
export const Application = () => {
	return (
		<box width="100%" height="100%" backgroundColor="#000000">
			<box style={{ flexGrow: 1, backgroundColor: "red" }} />
			<box
				flexShrink={0}
				style={{ position: "relative", backgroundColor: "blue", height: 4 }}
			>
				<box
					style={{
						flexDirection: "column",
						border: ["left"],
						borderStyle: "heavy",
						borderColor: "#fb8500",
						backgroundColor: "green",
						position: "absolute",
						top: 0,
						left: 0,
						right: 0,
					}}
				>
					<text>I'm absolute green in blue top:0</text>
				</box>
			</box>
		</box>
	);
};
```

Image

---

```
export const Application = () => {
	return (
		<box width="100%" height="100%" backgroundColor="#000000">
			<box style={{ flexGrow: 1, backgroundColor: "red" }} />
			<box
				flexShrink={0}
				style={{ position: "relative", backgroundColor: "blue", height: 4 }}
			>
				<box
					style={{
						flexDirection: "column",
						border: ["left"],
						borderStyle: "heavy",
						borderColor: "#fb8500",
						backgroundColor: "green",
						position: "absolute",
						bottom: 0,
						left: 0,
						right: 0,
					}}
				>
					<text>I'm absolute green in blue bottom:0</text>
				</box>
			</box>
		</box>
	);
};
```

Image

edlsh:
Fixes #406

Yoga's `getComputedLayout()` returns positions relative to parent ([ref](https://github.com/facebook/yoga/issues/787)), but the x/y getters only added parent offset for `relative` elements. Removed the position type check.

**Verification:** Tested nested absolute child with `bottom: 0` - now renders at bottom of parent container instead of global offset.

### DIFF
--- a/packages/core/src/tests/absolute-positioning.snapshot.test.ts
+++ b/packages/core/src/tests/absolute-positioning.snapshot.test.ts
@@ -9,7 +9,12 @@ let captureFrame: () => string
 let resize: (width: number, height: number) => void
 
 beforeEach(async () => {
-  ;({ renderer: testRenderer, renderOnce, captureCharFrame: captureFrame, resize } = await createTestRenderer({
+  ;({
+    renderer: testRenderer,
+    renderOnce,
+    captureCharFrame: captureFrame,
+    resize,
+  } = await createTestRenderer({
     width: 40,
     height: 20,
   }))


kommander:
I am surprised the tests passed on this, so either the cases I am thinking about are not test covered yet or I am just wrong. I was sure there was a reason that was needed, but that was also months ago. Can we add some absolute positioning snapshot tests and see if it renders correctly? I'll create a snapshot from this and we can test with opencode.

edlsh:
@kommander

kommander:
I am slower on weekends, I need some time to work through this.

PR #414, opened 2025-12-13
fix: reset flex properties on style removal

### LINKED ISSUE #391: bug: flex system bug when state changes
Firstly, I love the work you guys are going on OpenTUI!! Keep up the good work!!

Repo: https://github.com/k2on/opentui-flex-bug

This code:

```tsx
import { createRoot, useKeyboard } from "@opentui/react";
import { useState } from "react";

function App() {
  const [bool, setBool] = useState(false);

  useKeyboard(key => {
    if (key.name == "t") return setBool(v => !v);
  })

  if (!bool) return <box alignItems="center"><text fg="blue">My in the center!</text></box>;

  return <box><text fg="red">Why am I aligned to the right?</text></box>;
}

createRoot(renderer).render(<App />);
```

Does not align the text correctly when the state changes.

Expected:

Image

Actual after switching state (pressing "t"):

Image

Thank you for your help.
Have a blessed day,
Max

edlsh:
## Fix flex state change bug

Fixes #391

### Problem

When switching from `<box alignItems="center">` to `<box>` via React state, alignment stuck at previous value instead of resetting to default.

### Root Causes

1. `setStyle` didn't reset removed style properties to `null`
2. Flex properties only set conditionally—removing them left stale Yoga node values
3. `parseAlign` defaulted to `Auto`, not `Stretch` (CSS spec for `align-items`)

### Changes

- **`yoga.options.ts`**: Added `parseAlignItems()` defaulting to `Stretch`
- **`Renderable.ts`**: Set flex properties unconditionally; use `parseAlignItems` for `alignItems`
- **`utils/index.ts`**: Reset removed style keys to `null` in `setStyle`

### References

Per [MDN align-items](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items):
> Initial value: `normal`. For flex items, the keyword behaves as `stretch`.

Per [W3C CSS Flexbox Level 1](https://www.w3.org/TR/css-flexbox-1/):
> The official specification for flexbox alignment behavior.

Per [Yoga Layout Styling](https://www.yogalayout.dev/docs/styling/):
> Documents Yoga's defaults and differences from web CSS (notably `align-content` defaults to `flex-start`—but `align-items` is not listed, meaning it follows CSS default of `stretch`).

### Verification

```sh
cd packages/react && bun test layout.test.tsx
# 21 pass, 0 fail
```

Tests added:
- `should use default alignItems when conditionally switching components with React state`
- `should correctly align text when box has no explicit alignItems`
- `should reset alignItems when style prop removes the property`

### DIFF
--- a/packages/react/src/utils/index.ts
+++ b/packages/react/src/utils/index.ts
@@ -20,11 +20,9 @@ function initEventListeners(instance: Instance, eventName: string, listener: any
 }
 
 function setStyle(instance: Instance, styles: any, oldStyles: any) {
-  // Clear removed style keys first
   if (oldStyles != null && typeof oldStyles === "object") {
     for (const styleName in oldStyles) {
       if (oldStyles.hasOwnProperty(styleName)) {
-        // If the key existed in old styles but not in new styles, reset it
         if (styles == null || !styles.hasOwnProperty(styleName)) {
           // @ts-expect-error props are not strongly typed in the reconciler
           instance[styleName] = null
@@ -33,7 +31,6 @@ function setStyle(instance: Instance, styles: any, oldStyles: any) {
     }
   }
 
-  // Set new/changed style keys
   if (styles != null && typeof styles === "object") {
     for (const styleName in styles) {
       if (styles.hasOwnProperty(styleName)) {
--- a/packages/react/tests/layout.test.tsx
+++ b/packages/react/tests/layout.test.tsx
@@ -485,12 +485,7 @@ describe("React Renderer | Layout Tests", () => {
   })
 
   describe("Layout Property Reset on Component Change (Issue #391)", () => {
-    it("should use default alignItems when conditionally switching components with React state", async () => {
-      // This test reproduces the exact bug from issue #391:
-      // When React state changes cause a component tree to switch between
-      // <box alignItems="center"> to <box> (without alignItems),
-      // the text position should reset to default alignment
-
+    it("should reset alignItems when conditionally switching components", async () => {
       let setToggle: (value: boolean) => void
 
       function TestComponent() {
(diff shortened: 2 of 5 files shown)


packages/react/tests/layout.test.tsx
```diff
@@ -481,4 +485,142 @@ describe("React Renderer | Layout Tests", () => {
...
+          )
+        }
+        return (
+          <box width={40} height={3}>
+            <text>Default</text>
+          </box>
+        )
+      }
+
+      const testRenderer = await createTestRenderer({ width: 40, height: 5 })
```
msmps: you should not need to `createTestRenderer` here or `createRoot` as this is all handled for you in the `testRender` method [here](https://github.com/sst/opentui/blob/4403a69a7b801c594b2dbf06255e5e2cec4fc4c7/packages/react/src/test-utils.ts#L10) ([examples](https://github.com/sst/opentui/blob/1588e49b81dfc575e6b5f7e94aeb821c7fe5b82c/packages/react/tests/layout.test.tsx#L35-L44))

packages/react/src/utils/index.ts
```diff
@@ -20,20 +20,27 @@ function initEventListeners(instance: Instance, eventName: string, listener: any
...
+  }
+
+  // Set new/changed style keys
+  if (styles != null && typeof styles === "object") {
+    for (const styleName in styles) {
+      if (styles.hasOwnProperty(styleName)) {
+        const value = styles[styleName]
+        const oldValue = oldStyles?.[styleName]
+        if (value !== oldValue) {
+          // @ts-expect-error props are not strongly typed in the reconciler
```
msmps: i think it would be better for this to be reverted and the condition checked [here](https://github.com/sst/opentui/blob/1588e49b81dfc575e6b5f7e94aeb821c7fe5b82c/packages/react/src/utils/index.ts#L127-L129) instead! although if this is specific to the `style` prop which is an object this could work! i will pull this down this evening and give it a quick test

packages/react/tests/layout.test.tsx
edlsh: You are correct this is redundant will remove

packages/react/src/utils/index.ts
edlsh: If you think this is a better place to catch this generically I'm open to it. Let me know what you find when you test.

kommander: If you two can agree on where to put this, lets bring this in.

msmps: @kommander this is fine! happy for this to go in 👍

PR #415, opened 2025-12-13
fix(core): wrap kitty graphics query in tmux DCS passthrough

### LINKED ISSUE #334: Kitty Graphics Protocol query leaks Gi=31337 response into tmux pane title
# Describe the bug
When running an application using @opentui/core (opencode) inside tmux, a cryptic string resembling `Gi=31337,s=1,v=1,a=q,t=d,f=24;AAAA` appears in the tmux pane title or leaks into the shell output immediately after startup.

# Root Cause
The issue stems from the queryTerminalSend function in packages/core/src/zig/terminal.zig, which sends the kittyGraphicsQuery escape sequence unconditionally during initialization.
https://github.com/sst/opentui/blob/main/packages/core/src/zig/terminal.zig#L176
// packages/core/src/zig/terminal.zig
ansi.ANSI.kittyGraphicsQuery ++
This sends \x1b_Gi=31337... to probe for Kitty graphics support. When running inside tmux, the terminal's response to this query is often not trapped correctly by the application.

# To Reproduce
1. Start tmux.
2. Run an opentui-based application (e.g., opencode).
3. Observe that the pane title changes to the graphics query response
 
# Proposed Solution
1. Conditionality: The graphics probe should not be sent unconditionally. It should check if the environment supports it or if a specific environment variable is set to disable it.
2. Tmux Support: If the intention is to support graphics inside tmux, the query should be wrapped in the correct passthrough escape sequence when TERM=screen or TERM=tmux.
3. Escape Hatch: Add an environment variable (e.g., OPENTUI_NO_GRAPHICS) that allows users to disable this specific probe entirely.

edlsh:
Fixes #334

## Summary
- Re-enable kitty graphics query (was disabled due to tmux pane title corruption)
- Wrap query in tmux DCS passthrough when `TMUX` env var is present
- Skip query entirely for `TERM=screen*/tmux*` without `TMUX` (could be GNU Screen)
- Add `OPENTUI_NO_GRAPHICS` env var to disable query manually

## Technical Details
The kitty graphics protocol uses APC sequences (`\x1b_G...`) which tmux intercepts and leaks into pane titles. The fix wraps the query in tmux's DCS passthrough format (`\x1bPtmux;...\x1b\\`) with ESC chars doubled per the protocol spec.

**References:**
- [Kitty Graphics Protocol - Querying Support](https://sw.kovidgoyal.net/kitty/graphics-protocol/#querying-support-and-available-transmission-mediums)
- [tmux DCS Passthrough](https://man.archlinux.org/man/tmux.1.en)

## Testing
```bash
# Native tests
cd packages/core/src/zig && zig build test --summary all
# 1241/1242 tests passed
```

Manual verification in tmux - pane title no longer shows `Gi=31337...`

### DIFF
--- a/packages/core/src/zig/ansi.zig
+++ b/packages/core/src/zig/ansi.zig
@@ -90,7 +90,6 @@ pub const ANSI = struct {
     pub const csiUQuery = "\x1b[?u";
     pub const kittyGraphicsQuery = "\x1b_Gi=31337,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\\\x1b[c";
 
-    // Capability queries that need DCS wrapping for tmux
     pub const capabilityQueries = decrqmSgrPixels ++
         decrqmUnicode ++
         decrqmColorScheme ++
@@ -103,7 +102,6 @@ pub const ANSI = struct {
     pub const tmuxDcsStart = "\x1bPtmux;";
     pub const tmuxDcsEnd = "\x1b\\";
 
-    // Helper to wrap a sequence for tmux passthrough
     pub fn wrapForTmux(comptime seq: []const u8) []const u8 {
         comptime {
             var result: []const u8 = tmuxDcsStart;
@@ -118,7 +116,6 @@ pub const ANSI = struct {
         }
     }
 
-    // tmux DCS passthrough variants (ESC chars doubled)
     pub const kittyGraphicsQueryTmux = wrapForTmux(kittyGraphicsQuery);
     pub const capabilityQueriesTmux = wrapForTmux(capabilityQueries);
     pub const sixelGeometryQuery = "\x1b[?2;1;0S";
--- a/packages/core/src/zig/terminal.zig
+++ b/packages/core/src/zig/terminal.zig
@@ -169,21 +169,16 @@ pub fn queryTerminalSend(self: *Terminal, tty: anytype) !void {
         ansi.ANSI.hideCursor ++
         ansi.ANSI.saveCursorState);
 
-    // Send capability queries - DCS wrapped if tmux detected via env vars
     if (self.in_tmux) {
         try tty.writeAll(ansi.ANSI.capabilityQueriesTmux);
     } else {
-        // Send unwrapped, but mark pending in case xtversion detects tmux
         try tty.writeAll(ansi.ANSI.capabilityQueries);
         self.capability_queries_pending = true;
     }
 
-    // Explicit width detection
     try tty.writeAll(ansi.ANSI.home ++
         ansi.ANSI.explicitWidthQuery ++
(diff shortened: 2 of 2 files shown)


edlsh:
I can also update so:

1. TMUX env → use passthrough. 
2. TERM=tmux* (without TMUX) → also use passthrough (likely still tmux)
3. TERM=screen* (without TMUX) → skip (could be GNU Screen)

This way we try passthrough for anything tmux-like, only skip for ambiguous screen* cases.

kommander:
I tried playing with passthrough as well, the issue is opentui applications running remote, where the env will say "no tmux" and only delayed when maybe receiving an xtversion response will it know about tmux, but at that point will already have sent without DCS and mess up the window title.

edlsh:
Could we restructure the query sequence so queryTerminalSend emits the graphics query only after parsing xtversion reveals the terminal type?

kommander:
Yeah, that would work. Sorry for the late answer. The xtversion query should be sent as very first query. Some caveat: the xtversion query may never get a response, as for example xterm.js (VSCode) does not send any. Maybe there are better ways to detect it remotely. So the tricky part would be to figure out a timeout that makes sense.

edlsh:
hey @kommander — tweaked the deferred flow so tmux still gets the DCS-wrapped kitty query even if xtversion never responds (TMUX env). also added send/flush error logs. let me know if you want a timeout-based fallback for non-tmux terminals.

kommander:
Could work better, yeah. I think it's fine if it only works for local tmux to start with, as long as it doesn't send the escaped query for non-tmux. 

A little nitpick: We have `capabilities.terminal.name`, could we just use that and set it when detected via `env` for tmux for example, instead of having an additional `in_tmux` flag? Then it can just always check `capabilities.terminal.name` compared to checking both. Then it also wouldn't matter if the name is from `xtversion` or not. If it is coming from `env` we can be pretty sure and from `xtversion` it then just came later. 

One other thing, `/scripts` is meant for things executed regularly, manually or automated. I would put the [test-tmux-graphics-334.sh](https://github.com/sst/opentui/pull/415/files#diff-464c38509cdf1e4311760f22760021e437e750bd31e56b310d52c87547331ca7) in `core/dev` if you want to keep that.

edlsh:
Hey, quick note on the tmux detection nit: if we drop  and rely only on , xtversion from the inner terminal (e.g. kitty) can overwrite it and we’ll stop wrapping the kitty graphics query, which can reintroduce the pane-title corruption in tmux. The current code keeps  from TMUX/TERM and uses that in sendPendingGraphicsQuery, so it stays safe even if xtversion later reports kitty. If we want to avoid the extra flag, we’d need to persist the tmux env in the name and avoid later overwrite, or keep a separate tmux-env flag.

PR #416, opened 2025-12-14
use packages=external in build scripts

remorses:
Build scripts were manually listing dependencies to externalize, but core was missing regular `dependencies` (only had `optionalDependencies` and `peerDependencies`). This caused deps like `yoga-layout`, `diff`, `jimp`, and `bun-ffi-structs` to be bundled into dist.

Switched all packages to use Bun's `packages: "external"` option which automatically externalizes all npm imports.

Tested compiling examples to standalone binaries with `bun build --compile` - yoga-layout WASM and other dependencies are still bundled correctly by Bun at compile time.


kommander:
Need to run some tests first with opencode builds.

remorses:
This PR will let me override yoga-layout with bun-yoga in @Latitudes-Dev fork so we can test if there are any issues with yoga ffi in windows and linux

https://github.com/Latitudes-Dev/shuvcode

simonklee:
Outdated, closing

PR #426, opened 2025-12-19
Fix wrong Kitty keyboard protocol flags

devxoul:
The Kitty flag values were incorrectly entered in #398

Document: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#id5

Screenshot 2025-12-19 at 7 04 21 PM

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -87,13 +87,7 @@ export interface CliRendererConfig {
   useAlternateScreen?: boolean
   useConsole?: boolean
   experimental_splitHeight?: number
-  useKittyKeyboard?: {
-    disambiguate?: boolean // Disambiguate escape codes (fixes ESC timing, alt+key ambiguity, ctrl+c as event). Default: true
-    alternateKeys?: boolean // Report alternate keys (numpad, shifted, base layout) for cross-keyboard shortcuts. Default: true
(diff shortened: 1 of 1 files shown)


kommander:
Does that change current behavior?

devxoul:
@kommander possibly, but I'm not 100% sure what you intendend originally.

I found this typo while I was debugging that key bindings like <kbd>ctrl</kbd>+<kbd>C</kbd> didn't work in Korean keyboard. In the Korean keyboard the Kitty event was: <kbd>ctrl</kbd>+<kbd>ㅊ</kbd> (<kbd>ㅊ</kbd> is at the same location of <kbd>C</kbd>)

Alternate key flag, as far as I know, converts language specific key events into base keys: <kbd>ctrl</kbd>+<kbd>ㅊ</kbd> into <kbd>ctrl</kbd>+<kbd>C</kbd>. As the original code at #398 specifies the alternate keys as the default flag, I thought this was the original intention. But this doesn't seem to be working at the moment.

So yeah, the behavior might change, but I believe the new behavior is what it was originally intended.

```ts
let flags = KITTY_FLAG_ALTERNATE_KEYS

if (config.events) {
  flags |= KITTY_FLAG_EVENT_TYPES
}
```

devxoul:
(FYI) follow-up PR in opencode: https://github.com/sst/opencode/pull/5956

PR #428, opened 2025-12-19
feat: implement osc8 hyperlinks

thehamsti:
### Changes:
* Parses hyperlinks with a regex in tree-sitter-styled-text
* Underlines the hyperlinks (Not sure if this is too opinionated or what style you'd want)
* Registers the link in a link registry to map screen positions to link IDs
* On option(alt)+left click in TextBufferRenderable lookup if theres a link at that position with callback with url to app

### Possible issues:
* Styles with underline
* Pretty limited url parsing regex
* Just alt? Just ctrl? Alt and ctrl?


kommander:
I integrated links with cell attributes to draw buffers as expected, with overlays not rendering the links, like it would work in a browser. In #447.

PR #430, opened 2025-12-19
fix(core): fix memory leaks in Zig unicode encoding and grapheme pool

edlsh:
## Summary

Fixes critical memory leaks in the Zig native code for unicode encoding and grapheme pool operations.

### 1. Fix `errdefer` that never runs in `encodeUnicode` (lib.zig)

**Problem**: The function returns `bool`, but `errdefer` only triggers on error union returns. All failure paths used `catch return false`, which is a normal return—not an error return—so cleanup never executed.

**Fix**: Replace `errdefer` with `defer` + success flag pattern:
- Added `var success = false;` at start
- Changed `errdefer` to `defer { if (!success) { ... } }`
- Set `success = true;` before final `return true;`

### 2. Fix grapheme leak on realloc failure (lib.zig)

**Problem**: After `pool.alloc()` + `pool.incref()`, if `realloc` fails, the newly allocated grapheme wasn't tracked yet in the result array, so it would leak.

**Fix**: Track pending grapheme with `var pending_gid: ?u32 = null;`:
- Set `pending_gid = gid` immediately after allocation
- Clear `pending_gid = null` after successfully stored in result array
- Cleanup decrefs `pending_gid` if set

### 3. Add bounds checks in GraphemePool (grapheme.zig)

**Problem**: `incref`, `decref`, `get`, and `getRefcount` didn't validate `class_id` before array access.

**Fix**: Added `if (class_id >= MAX_CLASSES) return GraphemePoolError.InvalidId;` to all four methods.

## Testing

- [x] `bun run build` passes (all 6 platform binaries)
- [x] `bun run test:native` passes (1241/1242 tests, 1 skipped)

### DIFF
--- a/packages/core/src/zig/grapheme.zig
+++ b/packages/core/src/zig/grapheme.zig
@@ -113,6 +113,7 @@ pub const GraphemePool = struct {
 
     pub fn incref(self: *GraphemePool, id: IdPayload) GraphemePoolError!void {
         const class_id: u32 = (id >> (GENERATION_BITS + SLOT_BITS)) & CLASS_MASK;
+        if (class_id >= MAX_CLASSES) return GraphemePoolError.InvalidId;
         const slot_index: u32 = id & SLOT_MASK;
         const generation: u32 = (id >> SLOT_BITS) & GENERATION_MASK;
         try self.classes[class_id].incref(slot_index, generation);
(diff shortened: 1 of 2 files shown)


edlsh:
Closing - PR included unrelated commits. Reopening with clean branch.

PR #433, opened 2025-12-19
feat(core): add MarkdownRenderable with automatic table alignment

remorses:
- Adds `MarkdownRenderable` that extends `CodeRenderable` with markdown-specific features
- Uses `marked` to parse markdown and automatically align table columns based on content width
- Accounts for concealed characters (formatting markers like backticks, bold, italic) when calculating column widths
- Related to #369 (table alignment) but this implements it as a dedicated Markdown renderable instead of modifying CodeRenderable
Screenshot 2025-12-20 at 00 44 44

Tables are rendered via flexbox boxes:

```
Here's the current structure:
tableBox (flexDirection: "row")
├── columnBox[0] (flexDirection: "column", border: top/bottom/left)
│   ├── headerBox (border: bottom)
│   │   └── TextRenderable "Emoji" (paddingLeft: 1, paddingRight: 1)
│   ├── cellBox (border: bottom)
│   │   └── TextRenderable "🎉"
│   ├── cellBox (border: bottom)
│   │   └── TextRenderable "🚀"
│   └── TextRenderable "日本語"
│
├── columnBox[1] (flexDirection: "column", border: all sides)
│   ├── headerBox
│   │   └── TextRenderable "Name"
│   ├── cellBox
│   │   └── TextRenderable "Party"
│   ...

```

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Publish preview packages
         run: |
-          npx pkg-pr-new publish --no-template \
+          npx pkg-pr-new publish --no-template --commentWithSha --bun \
             './packages/core/dist' \
             ${{ steps.native-packages.outputs.dirs }} \
             './packages/react/dist' \
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
     },
     "packages/core": {
       "name": "@opentui/core",
-      "version": "0.1.72",
+      "version": "0.1.74",
       "dependencies": {
         "bun-ffi-structs": "0.1.2",
         "diff": "8.0.2",
@@ -28,12 +28,12 @@
       },
       "optionalDependencies": {
         "@dimforge/rapier2d-simd-compat": "^0.17.3",
-        "@opentui/core-darwin-arm64": "0.1.72",
-        "@opentui/core-darwin-x64": "0.1.72",
-        "@opentui/core-linux-arm64": "0.1.72",
-        "@opentui/core-linux-x64": "0.1.72",
-        "@opentui/core-win32-arm64": "0.1.72",
-        "@opentui/core-win32-x64": "0.1.72",
+        "@opentui/core-darwin-arm64": "0.1.74",
+        "@opentui/core-darwin-x64": "0.1.74",
+        "@opentui/core-linux-arm64": "0.1.74",
+        "@opentui/core-linux-x64": "0.1.74",
+        "@opentui/core-win32-arm64": "0.1.74",
+        "@opentui/core-win32-x64": "0.1.74",
         "bun-webgpu": "0.1.4",
         "planck": "^1.4.2",
         "three": "0.177.0",
@@ -44,7 +44,7 @@
     },
     "packages/react": {
       "name": "@opentui/react",
-      "version": "0.1.72",
+      "version": "0.1.74",
       "dependencies": {
         "@opentui/core": "workspace:*",
         "react-reconciler": "^0.32.0",
@@ -72,7 +72,7 @@
     },
     "packages/solid": {
       "name": "@opentui/solid",
-      "version": "0.1.72",
+      "version": "0.1.74",
       "dependencies": {
         "@babel/core": "7.28.0",
         "@babel/preset-typescript": "7.27.1",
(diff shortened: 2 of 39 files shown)


kommander:
Uhh looks good! I need to work through that.

kommander:
It currently does not support streaming markdown properly. I can look into that. It needs an `append` API to append markdown chunks when they come in from an LLM response, only reparsing with marked what is not yet a finished block and not yet added as renderables, without removing and recreating all renderables for every update.

remorses:
Added incremental parsing and switched to using renderables for blocks so they do not rerender every time

For the streaming parser I used the same logic I used in [holocron](https://holocron.so), which I already tested quite a bit: https://github.com/remorses/holocron/blob/046ff49e60ab315b59929ef8b521546fe756cc98/contesto/src/lib/incremental-markdown-parser.ts#L52

Here is a demo of it inside opencode


https://github.com/user-attachments/assets/ae4a4deb-56a2-4806-96df-0d7a128b3d06

remorses:
I updated table rendering to show a row at a time during streaming, this should improve performance (because right now tables rerender fully each time) and looks less janky

kommander:
Cool, that works better. It does not support updating itself when `conceal` or `syntaxStyle` changes though it seems, which would cause issues for many opencode users I think. That would be the last detail to make it work for opencode, besides the table rendering not rendering full border joins, but it looks good as-is now and we can refine the table rendering in another iteration.

remorses:
with full table border joins you mean not overlapping border right? should be an easy fix with custom borders

kommander:
That works. Switching the theme in the demo blocks the main thread for ~1s. Is that `marked` parsing? That's one reason I made tree-sitter threaded, to not block when it takes longer for any reason.

remorses:
Most time is spend on rendering, because I just clear all the renderables when switching conceal and theme

I can probably only update fields in every renderable right? This should be much faster

Screenshot 2025-12-23 at 15 34 32

Screenshot 2025-12-23 at 15 34 25

remorses:
Tried rerendering instead of recreating the renderable and got a bit worse. But I cannot reproduce the 1 second freeze. For me it freezes only 300ms or less

remorses:
Added a performance fix for theme/conceal switching.

**Problem**: When `syntaxStyle` or `conceal` changed, we were clearing all block states and rebuilding everything from scratch (reparsing markdown, recreating all renderables).

**Fix**: Now we defer the work to `renderSelf()` and only update existing renderables instead of destroying/recreating them. This follows the same pattern as `CodeRenderable` which uses `_highlightsDirty`.

**CPU profiling results** (theme switch in the demo):

| Function | Before | After | Improvement |
|----------|--------|-------|-------------|
| textBufferSetStyledText | 113ms | 67ms | ↓41% |
| updateLayout | 99ms | 17ms | ↓83% |
| render | 160ms | 63ms | ↓61% |
| createTextBuffer | 35ms | 19ms | ↓47% |

**Estimated theme switch latency**: ~100-150ms (down from ~300ms)

kommander:
I can only check that tomorrow, not allowed to touch the keyboard today. Happy holidays!

remorses:
Merry Christmas!

kommander:
I made it update the tables in-place instead of recreating all elements, which is somewhat better. I found that when streaming a lot of markdown (endless mode in the demo), it becomes slower and slower. Seems like something is resetting the `parserState` so it re-parses the markdown and/or the tables are re-created for incremental updates which is expensive and in long content takes >10ms. There are some perf updates for text rendering coming in core, but I wouldn't rely on that. 

I wonder if there is a better way to render tables, without relying on yoga layout, as for large tables it would be quite a lot of nested nodes.

remorses:
Thank you for fixing the reset. I think using Yoga is completely fine. I noticed no lags rendering a lot of tables in opencode with the markdown renderable.

kommander:
https://github.com/user-attachments/assets/bd367a9a-9606-4f15-975c-5d05d62a430f

Check the "Overall" render time in the stats, it should be low and stable, but it increases the more content is being streamed. In the markdown demo when I set stream speed to fastest and shift+E for endless stream (which can be stopped with shift+X again).

remorses:
I found what was the issue with the lower FPS when the is content growing, the issue is layout recalculations

Every time a yoga node changes we also mark the parents dirty, so yoga calls measure functions again which can be slow when there are a lot of nodes

Using bun-yoga makes this problem less noticeable but does not remove it 

To make layout calculation faster during streaming with a lot of content there would need to be a way to cache layout for nodes if content did not change like the browser does

Screenshot 2025-12-29 at 19 40 09

remorses:
Adding a cache for `measureFunc` in `packages/core/src/renderables/TextBufferRenderable.ts` lowers the time 3x, Yoga seems to call measure 3 times but we can cache it

kommander:
I'll do the performance improvements this week and then this should be good.

kynnyhsap:
Hope this can be merged soon

kommander:
I still have some concerns about performance with marked, the current incremental parsing seems to not properly work. 

Will merge this anyway now and we can iterate on it. I would put it into opencode with a flag to roll it out to a broader audience with the ability to switch back and forth should it not behave as expected.

PR #441, opened 2025-12-24
fix: hide terminal cursor when showCursor is false

remorses:
When `showCursor` prop changes from `true` to `false`, the cursor was still visible because `renderCursor` just returned early without hiding the cursor.

This fix calls `setCursorPosition(0, 0, false)` to hide the terminal cursor when `showCursor` is false or the component is not focused.

### DIFF
--- a/packages/core/src/renderables/EditBufferRenderable.ts
+++ b/packages/core/src/renderables/EditBufferRenderable.ts
@@ -279,7 +279,6 @@ export abstract class EditBufferRenderable extends Renderable implements LineInf
   set showCursor(value: boolean) {
     if (this._showCursor !== value) {
       this._showCursor = value
-      // Hide cursor when showCursor changes to false while focused
       if (!value && this._focused) {
         this._ctx.setCursorPosition(0, 0, false)
       }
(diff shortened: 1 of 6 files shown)


packages/core/src/renderables/EditBufferRenderable.ts
```diff
@@ -279,6 +279,10 @@ export abstract class EditBufferRenderable extends Renderable implements LineInf
   set showCursor(value: boolean) {
     if (this._showCursor !== value) {
       this._showCursor = value
+      // Hide cursor when showCursor changes to false while focused
+      if (!value && this._focused) {
```
kommander, line 282: Why only call `setCursorPosition` when it is `false`? Doesn't it set the cursor in the next render loop through `requestRender`?

remorses: When `showCursor` becomes `false`, `renderCursor()` exits early at line 581:
```ts
if (!this._showCursor || !this._focused) return
```
It never reaches any `setCursorPosition` call, so `requestRender()` alone won't hide the cursor - it would remain visible at its last position.

When `showCursor` becomes `true`, the render loop handles it because `renderCursor()` runs fully and calls `setCursorPosition(..., true)`.

remorses: Basically React props use this setter when the prop changes

PR #454, opened 2025-12-29
fix: handle flushSync rename in react-reconciler 0.32.0

remorses:
`flushSync` was renamed to `flushSyncFromReconciler` in react-reconciler 0.32.0.

This fix uses nullish coalescing to check for the new name first, falling back to the old name for backwards compatibility.

Ref: https://github.com/facebook/react/issues/35424

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
     },
     "packages/core": {
       "name": "@opentui/core",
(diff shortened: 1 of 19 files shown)


remorses:
I also added an example to showcase flushSync works correctly

msmps (approved):

Issue #467, opened 2026-01-03
Scrollbox receives clicks outside its visible bounds (hit-testing ignores clipping)


nyxkrage:
When a `scrollbox` contains content taller than its visible area, clicks outside the scrollbox's visible bounds are incorrectly captured by the scrollbox. The content is visually clipped correctly, but hit-testing uses the full unclipped content height.

- When scrolled to top: clicking below the dialog hits the scrollbox
- When scrolled to bottom: clicking above the dialog hits the scrollbox

## Steps to Reproduce

1. Create a dialog with an overlay (for click-to-close)
2. Add a scrollbox inside the dialog with content taller than the visible area (e.g., 50 lines)
3. Open the dialog
4. Click directly below the scrollable content (in the overlay area)

Expected: The overlay receives the click and the dialog closes  
Actual: The scrollbox receives the click (even though it's visually contained within the dialog)

## Minimal Example

```tsx
import { render, useKeyboard, useTerminalDimensions, useRenderer } from "@opentui/solid"
import { createSignal, Show, For } from "solid-js"

function App() {
  const [showDialog, setShowDialog] = createSignal(false)
  const [lastClick, setLastClick] = createSignal("none")
  const dimensions = useTerminalDimensions()
  const renderer = useRenderer()

  useKeyboard((key) => {
    if (key.name === "q") {
      renderer.destroy()
      process.exit(0)
    }
    if (key.name === "d") setShowDialog(!showDialog())
    if (key.name === "escape") setShowDialog(false)
  })

  const lines = Array.from({ length: 50 }, (_, i) => `Line ${i + 1}: This is some content`)

  return (
    <box flexDirection="column" height={dimensions().height} width={dimensions().width} backgroundColor="#1a1a2e">
      <text>Press 'd' to toggle dialog, 'q' to quit, 'esc' to close</text>
      <text>Last click: {lastClick()}</text>

      <Show when={showDialog()}>
        {/* Overlay - clicking here should close the dialog */}
        <box
          position="absolute"
          top={0}
          left={0}
          width={dimensions().width}
          height={dimensions().height}
          backgroundColor="#ff000033"
          onMouseDown={() => {
            setLastClick("OVERLAY (red)")
            setShowDialog(false)
          }}
        >
          <box
            position="absolute"
            top={Math.floor(dimensions().height / 4)}
            left={Math.floor(dimensions().width / 4)}
            width={Math.floor(dimensions().width / 2)}
[...]

PR #470, opened 2026-01-04
Fix scrollbox hit-testing to respect overflow clipping

nyxkrage:
Exports OptimizedBuffer.clipRectToScissor to the TypeScript interface and uses it to clip hit rects to the active scissor before adding them to the hit grid.  
This fixes scrollbox content receiving clicks outside its visible bounds.    

See #467 for repro example.

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -1325,10 +1325,7 @@ export abstract class Renderable extends BaseRenderable {
     }
 
     this.markClean()
-    const clippedRect = buffer.clipRectToScissor(this.x, this.y, this.width, this.height)
-    if (clippedRect) {
-      this._ctx.addToHitGrid(clippedRect.x, clippedRect.y, clippedRect.width, clippedRect.height, this.num)
-    }
+    this._ctx.addToHitGrid(this.x, this.y, this.width, this.height, this.num)
 
     if (this.buffered && this.frameBuffer) {
       buffer.drawFrameBuffer(this.x, this.y, this.frameBuffer)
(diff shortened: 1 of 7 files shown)


kommander:
Cool! I'd try to do that in native only, but this works.

nyxkrage:
Yeah, my original idea was also to handle it all on the Zig side, but I didn't really see a good place to bridge the renderer<->buffer gap with the hitgrid being tied to the renderer context and the scissor stack being tied to the buffer, so ended doing it in TS, since it ended up being pretty straight forward.

But now taking a look again, I think I got poisoned by looking at the TS code first, because the zig renderer obviously has direct access to the buffer. Let me push up a "fix" to handle it purely in native code.

Though would there be any reason to want to add something to the hitgrid unclipped? I am not seeing any reason to not just let addToHitGrid also handle the clipping itself all the time.

kommander:
> Though would there be any reason to want to add something to the hitgrid unclipped?

I've been thinking about this and I can't come up with a good reason, as only things that are actually drawn would be clickable. Like you did it for the nextRenderBuffer will now only apply the scissor from the nextRenderBuffer? 

I am struggling with some concepts of global vs. renderer vs. buffers, as anything should be renderer scoped to support multiple renderer instances in the same process (not fully there yet), but I also want to keep things like the buffers renderer agnostic to be able to draw a whole renderable tree with any renderer basically, if that makes sense. 

Thinking about it... `addToHitGrid` should be a render command like `pushScissor` and be executed in the _render pass_, not the layout pass. The command can carry a reference to the buffer and use the buffer pointer to pass it to native and `addToHitGrid` can apply the scissor if any.

kommander:
Closed in favor of #480

PR #472, opened 2026-01-04
bench(text-buffer-view): add measureForDimensions benchmark

simonklee:
Add a benchmark for measureForDimensions().

This function creates a fresh arena and performs a full rope walk on every call.

The benchmark has two scenarios:

- static: repeated calls on unchanged content (measures cache-miss penalty)
- streaming: calls interleaved with token appends (simulates LLM output)

This adds a baseline for measuring future fixes like result caching, arena reuse.

Refs: anomalyco/opencode#6172


kommander:
Thanks, yeah that is awful.

PR #475, opened 2026-01-05
perf(text-buffer-view): cache measureForDimensions results

simonklee:
Yoga calls measureForDimensions repeatedly during layout. This adds caching
keyed by (buffer, epoch, width, wrap_mode) to avoid repeated rope walks.

- Use epoch instead of dirty flag since other code paths clear dirty
- Reuse measure arena with retain_capacity to reduce mmap churn during streaming
- Add fast path for width=0 or wrap_mode=.none that skips virtual line calculation

I can split up the arena-reuse if you prefer that in another PR.

### DIFF
--- a/packages/core/src/text-buffer-view.test.ts
+++ b/packages/core/src/text-buffer-view.test.ts
@@ -637,5 +637,69 @@ describe("TextBufferView", () => {
       expect(result!.lineCount).toBe(4)
       expect(result!.maxWidth).toBe(10)
     })
+
+    it("should cache measure results for same width", () => {
+      const styledText = stringToStyledText("ABCDEFGHIJKLMNOPQRST")
+      buffer.setStyledText(styledText)
+
+      view.setWrapMode("char")
+
+      // First call - cache miss
+      const result1 = view.measureForDimensions(10, 10)
+      expect(result1).not.toBeNull()
+      expect(result1!.lineCount).toBe(2)
+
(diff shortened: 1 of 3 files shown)


remorses:
See this [PR](https://github.com/anomalyco/opentui/pull/456) for something similar. Notice I had to measure with the max height otherwise there were some visual bugs, like in opencode the absolutely positioned autocomplete was appearing out of place, did you already test if that works in opencode with this branch?

kommander:
@remorses are these issues you had reproducible with a test maybe?
@simonklee I migrated over some tests as I was working on this on another branch as well, but your implementation is better.

kommander:
@remorses I can't reproduce an issue with the autocomplete position on this branch.

kommander:
Taking this over #456 for having it in native, plus the arena fix.

simonklee:
Thanks @kommander, for adding the typescript integration tests and merging. I haven't really gotten around looking into that part of the code much yet so it wasn't on my mind that it was a possibility. I originally started looking into this based on the https://github.com/anomalyco/opencode/issues/6172 issue. I felt this PR was about as minimal as you can go with mitigation the issue, without rewriting things to consider only the viewport or something like that.

remorses:
Cool! What is the performance increase after this change?

kommander:
~50x performance improvement for static content, according to the benchmark from ~90ms to ~1.5ms.

PR #480, opened 2026-01-06
fix(hitgrid): stale hit grid after scroll/translate

simonklee:
Add immediate hit grid sync for scroll and translate changes

Before this change, when a scrollbox scrolled, the hit grid stayed stale until the next render completed. Hover states wouldn't update even though elements visually moved. Users would hover over items that had already scrolled away.

The fix adds on-demand hit grid sync. When translateX/Y changes, the renderer
marks the grid dirty and rechecks hover state. The renderer now:

1. Marks the hit grid dirty
2. Rechecks hover state using the latest pointer position
3. Rebuilds the hit grid immediately (before the next render) if a hit test occurs while dirty

The hit grid rebuild mirrors the render traversal but uses screen coordinates for scissor rects. Buffered renderables need this because they render at (0,0) in their local buffer, but hit testing happens in screen space.

Captured renderable (during drag) is excluded from the hit grid so drop targets receive events correctly.

I added the demo from #467 as an example as well as another debug example i used when implementing it. Both should probably be deleted before merging. You can copy these examples to 'main' to see the difference between how things react in main vs the branch.

The PR is a slightly different approach than what you discussed in https://github.com/anomalyco/opentui/pull/470#issuecomment-3711893881 comment, and I don't mind if you prefer your approach, but maybe you'll find some ideas here.

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -1285,11 +1285,12 @@ export abstract class Renderable extends BaseRenderable {
       const scissorRect = this.getScissorRect()
       renderList.push({
         action: "pushScissorRect",
-        renderable: this,
         x: scissorRect.x,
         y: scissorRect.y,
         width: scissorRect.width,
         height: scissorRect.height,
+        screenX: this.x,
+        screenY: this.y,
       })
     }
     const visibleChildren = this._getVisibleChildren()
@@ -1521,11 +1522,12 @@ interface RenderCommandBase {
 
 interface RenderCommandPushScissorRect extends RenderCommandBase {
   action: "pushScissorRect"
-  renderable: Renderable
   x: number
   y: number
   width: number
   height: number
+  screenX: number
+  screenY: number
 }
 
 interface RenderCommandPopScissorRect extends RenderCommandBase {
@@ -1608,13 +1610,7 @@ export class RootRenderable extends Renderable {
           break
         case "pushScissorRect":
           buffer.pushScissorRect(command.x, command.y, command.width, command.height)
-          // Hitgrid always uses absolute screen coordinates
-          this._ctx.pushHitGridScissorRect(
-            command.renderable.x,
-            command.renderable.y,
-            command.renderable.width,
-            command.renderable.height,
-          )
+          this._ctx.pushHitGridScissorRect(command.screenX, command.screenY, command.width, command.height)
           break
         case "popScissorRect":
           buffer.popScissorRect()


simonklee:
Nice that you upgraded to 15, guess I have to refactor for ci to pass :)

kommander:
Oh yeah, that just happened earlier today. Finally.

packages/core/src/Renderable.ts
```diff
@@ -1003,6 +1005,25 @@ export abstract class Renderable extends BaseRenderable {
     }
   }
 
+  public updateFromLayoutRecursive(): void {
```
kommander: This is very expensive, especially when it causes updateFromLayout to be called multiple times.

kommander:
Cool, this is very close to what I had in mind. The `updateFromLayoutRecursive` and `hasDirtyLayoutRecursive` methods can accumulate to be very expensive though, because calling yoga methods is ridiculously expensive. I wonder if the eager recursive updates and dirty lookups are really necessary when the hitgrid has its own scissor stack.

packages/core/src/Renderable.ts
simonklee: Yeh, this wasn't really needed.

simonklee:
> Cool, this is very close to what I had in mind. The `updateFromLayoutRecursive` and `hasDirtyLayoutRecursive` methods can accumulate to be very expensive though, because calling yoga methods is ridiculously expensive. I wonder if the eager recursive updates and dirty lookups are really necessary when the hitgrid has its own scissor stack.

Good call. I removed the recursive layout dirty check + update from syncHitGridIfNeeded. Hitgrid rebuilds now just mirror the render traversal with screen‑space scissors and use the last computed layout + current translate. Layout sync stays in the normal render pass, so hit tests remain consistent with what's on screen while avoiding Yoga calls on hover/scroll.

Left is branch, right is `main`:

https://github.com/user-attachments/assets/22e09ebb-e3fd-4ee4-b466-c2aea93cc9f1

kommander:
@simonklee I simplified the implementation to eliminate the need for an additional tree walk to gather the hitgrid, dirty tracking and force recheck for hover state. As far as I can see the behaviour for the hover state is still correct this way. WDYT?

simonklee:
> @simonklee I simplified the implementation to eliminate the need for an additional tree walk to gather the hitgrid, dirty tracking and force recheck for hover state. As far as I can see the behaviour for the hover state is still correct this way. WDYT?

Thanks for simplifying, but it also changes semantics. With the new code, hit‑grid updates only after a render and hover only updates on pointer move (translateX/Y no longer rechecks). So if the pointer stays still after a scroll, hover remains stale until the user moves the mouse. I'll record a quick video.

simonklee:
https://github.com/user-attachments/assets/bc1fd0f5-aba6-4b69-a0f1-bc03663cb5f8

See how the hovered doesnt update in the first example while scrolling and not moving mouse.

kommander:
I see. Let me try reproducing that in a test. It should be possible to conditionally recheck after a frame was rendered to solve that.

simonklee:
A fun test: if you try this in "Chrome" you'll notice that if you scroll the "hover" doesn't detect things under the cursor until after you stop scrolling and ~500ms have passed. I just randomly thought I'd try. Chrome scrolls on the compositor thread and throttles pointer/hover work, so :hover doesn't update until scrolling settles.

I think:

  - It's reasonable to coalesce hover updates during scroll and only recheck once scrolling stops (debounce or "scroll‑idle" timer).
  - But we should still make sure hover eventually updates. 
  
If we do something similar to Chrome then probably best to implement that as an improvement in a new branch later. With your latest changes the `examples/scrollbox-overlay-hit-test.ts` still works so that's good. I think you know the library 10x better than me and you're better to judge what to do.

kommander:
Yeah, let me merge this and do the hover state update in another PR. It should be possible to reproduce this in a test.

I would not want the renderer to know anything about scroll state, as that could be nested and whatnot, making it complex again. It should be enough to debounce the hover event when a change was detected after a frame. 

Do you want to take that on?

EDIT: simple debounce for hover event would also work for anything that is animated and moving etc. so not scroll specific.

simonklee:
> Yeah, let me merge this and do the hover state update in another PR. It should be possible to reproduce this in a test.

Agree.

> I would not want the renderer to know anything about scroll state, as that could be nested and whatnot, making it complex again. It should be enough to debounce the hover event when a change was detected after a frame.
Sounds fair.
> Do you want to take that on?

Yeh, that sounds fun. I'll have a look tonight (CET).

PR #483, opened 2026-01-07
feat(core): improve stdin-buffer emoji/IME handling with Kitty protocol

GreyElaina:
## Summary

Add proper grapheme segmentation and Kitty emoji reassembly for correct handling of emoji, CJK characters, and other multi-codepoint sequences.

### StdinBuffer (grapheme segmentation)
- Add `grapheme-segmenter.ts` using `Intl.Segmenter` for Unicode-correct text segmentation
- Update `StdinBuffer` to emit complete grapheme clusters for non-escape input
- Preserve Kitty keyboard protocol sequences unchanged for downstream parsing

### KeyHandler (Kitty emoji reassembly)
- Add emoji codepoint detection helpers (`isGraphemeExtender`, `canStartGraphemeCluster`)
- Buffer Kitty sequences that form multi-codepoint emoji (ZWJ families, flags, skin tones, keycaps)
- Flush buffer on timeout or when non-emoji input arrives
- Preserve all raw sequences in `KeyEvent.raw` field

## Architecture

```
stdin → StdinBuffer → KeyHandler.processInput() → parseKeypress() → emit
              ↓                    ↓
     grapheme clusters      emoji reassembly
     (for raw input)        (for Kitty protocol)
```

**Key insight**: StdinBuffer handles grapheme segmentation for raw UTF-8 input, while KeyHandler handles emoji reassembly for Kitty protocol sequences. This separation preserves the `raw` field correctly.

## Supported Emoji Types
- Basic emoji: 😀
- ZWJ sequences: 👨‍👩‍👧 (family)
- Flag emoji: 🇺🇸 🇯🇵
- Skin tone modifiers: 👋🏻
- Keycap sequences: #️⃣
- Subdivision flags: 🏴󠁧󠁢󠁥󠁮󠁧󠁿

## Testing
- All 113 stdin-buffer tests pass
- All 60 renderer.input tests pass  
- All 40 KeyHandler tests pass (10 new emoji tests)

### DIFF
--- a/packages/core/src/lib/grapheme-segmenter.ts
+++ b/packages/core/src/lib/grapheme-segmenter.ts
@@ -1,43 +1,9 @@
 let segmenter: Intl.Segmenter | null = null
-let initPromise: Promise<void> | null = null
-let initError: Error | null = null
-
-function initializePolyfill(): Promise<void> {
-  if (initPromise) return initPromise
-
-  initPromise = (async () => {
-    if (typeof Intl === "undefined" || typeof (Intl as any).Segmenter !== "function") {
-      try {
-        await import("@formatjs/intl-segmenter/polyfill-force.js")
-      } catch (e) {
-        const message = e instanceof Error ? e.message : String(e)
-        initError = new Error(
-          `Failed to load Intl.Segmenter polyfill: ${message}. Please ensure @formatjs/intl-segmenter is installed or use a runtime that supports Intl.Segmenter natively.`,
-        )
-      }
-    }
-  })()
-
-  return initPromise
-}
-
-initializePolyfill()
 
 export function getGraphemeSegmenter(): Intl.Segmenter {
   if (segmenter) return segmenter
-
-  if (typeof Intl !== "undefined" && typeof (Intl as any).Segmenter === "function") {
-    segmenter = new Intl.Segmenter(undefined, { granularity: "grapheme" })
-    return segmenter
-  }
-
-  if (initError) {
-    throw initError
-  }
-
-  throw new Error(
-    "Intl.Segmenter is not available. Please ensure your runtime supports it or install @formatjs/intl-segmenter",
-  )
+  segmenter = new Intl.Segmenter(undefined, { granularity: "grapheme" })
+  return segmenter
 }
 
 function isHighSurrogate(code: number): boolean {


kommander:
Interesting, I hadn't thought about it from that perspective. That means full grapheme parsing for input though, which can get out of hand really fast. The stdin buffer was meant to handle incomplete sequences, split at byte level, which would break some behaviours. For input in the Textarea for example, even if the bytes arrive chunked, they would end up forming a joined grapheme at the end. Is this meant to have full grapheme character key events?

GreyElaina:
The grapheme-aware input is already on my PR branch with ASCII fast paths, so performance impact is minimal for typical input.
Scope clarification:

1. StdinBuffer: Only changes how non-escape sequences are split (uses firstGrapheme() instead of single char). This handles raw UTF-8 correctly.
2. KeyHandler: Already has emoji buffering for Kitty protocol (shouldBufferForEmoji() + 10ms timeout + Intl.Segmenter reassembly).

The issue I hit was specifically with Rime IME emoji input - it sends input through some keystroke-based mechanism that differs from:

- Clipboard paste (works fine)
- Raycast emoji picker (works fine)
- macOS fn emoji menu (works fine)

These might be using the same underlying mechanism, but I haven't dug deeper.
So this is an edge case fix. But using grapheme-based checking (isSingleGrapheme(s) instead of s.length === 1) is more robust anyway - who knows what weird input methods exist out there (CJK IMEs, etc.). Better to handle grapheme clusters correctly from the start than chase down individual edge cases.

kommander:
It feels like something is more fundamentally wrong with the stdin buffer if it causes such a complexity explosion. Grapheme detection is horribly fragile at the moment and already has a huge complexity in the native part to render them correctly. Doing something similar for input seems like beating a symptom of wrong underlying assumptions in the stdin buffer.

GreyElaina:
zig/native side grapheme is used for **rendering**, the stdin buffer is for input **parsing**, i must say they are two different things.

the stdin buffer itself is simple. these complexity is brought by the terminal protocol and Unicode itself (see libghostty, utf8 everywhere etc.), anyway not a design issue. it's not correct to break UTF-8 by bytes or to break emoji by codepoints if not according to grapheme.

kommander:
Okay, I think I understand better now. This is specifically a Kitty Keyboard issue, which emits separate input events for a ZWJ emoji for example, right? And even though inserting them one by one it would be rendered correctly, but not emitted as single input event, but separate ones for each byte?

GreyElaina:
Yeah exactly. Kitty protocol emits separate escape sequences per codepoint, not per byte. So 👨‍👩‍👧 (family emoji, 7 codepoints joined by ZWJ) comes as 7 separate `\x1b[<codepoint>u` sequences.
The fix buffers these with a 10ms timeout and reassembles using `Intl.Segmenter`. StdinBuffer itself didn't change much for this: the Kitty reassembly logic is in KeyHandler.

GreyElaina:
it seems to be done by #682

PR #484, opened 2026-01-07
test(core): add regression tests for CJK line boundary and emoji handling

GreyElaina:
Adds regression tests for CJK and emoji handling edge cases.

## Tests Added
- CJK character at line boundary tests (text-buffer-view)
- CJK wrapping and streaming append tests (Text.test.ts)
- Zig native test for streaming append with CJK
- Textarea keybinding tests for ZWJ emoji, flags, skin tones

## Additional
- Add `appendText()` method to `TextBufferRenderable` for streaming use case


kommander:
Were these tests supposed to fail without the ime input fix?

GreyElaina:
🤔 uh, this pr have problem and i donot care about this.

btw it donot work because `createKeyEvent` bypasses stdin-buffer & KeyHandler.

PR #505, opened 2026-01-09
chore: downgrade xtversion detection log from info to debug

assagman:
In the latest _opencode_  dev release, "0.0.0-dev-202601091759" , and I started getting this log while TUI is being loaded:

info(terminal): Terminal detected via xtversion: tmux next-3.5

https://github.com/user-attachments/assets/61835133-aab6-420a-91c9-6b759386de9c

---

## reproducing

```sh
git checkout v0.1.71
```

```
bun run build
```

```sh
bun run packages/react/examples/basic.tsx
```

output:

image


kommander:
Ah that's because now it actually sends the xtversion query.

kommander:
I just removed the log completely, seemed to have been some leftover.

PR #507, opened 2026-01-09
perf(utf8): optimize wrap breaks

simonklee:
Refactored `findWrapBreaks` to remove data dependencies. The old implementation chained `@select` calls serially. The new one uses bitwise ORs on boolean vectors, allowing the CPU to check all character classes (whitespace, punctuation, brackets) in parallel instructions.


kommander:
I assume you've run your local bench comparison. Would be awesome to get that into the pipeline.

simonklee:
> I assume you've run your local bench comparison. Would be awesome to get that into the pipeline.

Yeh, sorry. I forgot to add it to the PR description. I didn't really find this one in a benchmark, it just caught my eye when going through adjacent code for https://github.com/anomalyco/opentui/pull/506

PR #508, opened 2026-01-10
Fix(solid/examples): line-number-demo type errors and hidden header

jhroemer:
Few type errors: 
- Using strings instead of `RGBA` for the style definitions
- Using nonexistent `bold` prop on `text` component - changed to using `attributes` prop and `TextAttributes`

Additionally: 
The "Line numbers demo" header wasn't showing when I was testing. Its container seems to be squashed by the other content. Fixed by not allowing it to shrink. 

line-numbers-demo


jhroemer:
Btw, I'm currently starting a new personal project, which is why I'm looking at the examples. Did you discuss having documentation for the solid package specifically? (might've missed it, only found for the core package). 
In any case, I'd be happy to put in a bit of work.

kommander:
I have been holding back on documentation because the API is very unstable. Might have stabilized somewhat though. There have been suggestions of using (AI) tooling for documentation, I've never used any of these though and can't judge how helpful that would be. 

What I would like to have when starting out documentation would be a GitHub CI action/bot that checks a PR and if changes to the documentation are needed, with suggestions to documentation changes so they can be included or not, depending on if they make sense. 

That would at least somewhat help keeping the docs up to date.

There have also been requests for a website, which would have to be set up with automatic documentation updates for releases etc. 

Overall a huge undertaking. Would be happy to have help on that, because I am under water with all the other things going on. 

I think by now it would be okay to start properly documenting, at least the parts that are somewhat stable.

kommander:
Defining colors as hex string should work fine though, if not in a hot path where it has to be parsed all the time.

jhroemer:
@kommander thanks a lot!

Good to know your thoughts, I think all of it makes sense. I might just start w. a proposal for a simple markdown file for the solid package, but I agree about the keeping-things-in-sync part. Might make sense to link to some of the existing Core docs for some of it.

PR #511, opened 2026-01-11
fix: add .js extension to react-reconciler/constants imports for Node.js ESM

nathanclevenger:
## Problem

Node.js ESM mode requires explicit file extensions for subpath imports from packages. The current code imports from `react-reconciler/constants` without the `.js` extension, which works in Bun but fails in Node.js:

```
Error: Cannot find module 'react-reconciler/constants' 
imported from @opentui/react/index.js
Did you mean to import "react-reconciler/constants.js"?
```

This breaks projects using `@opentui/react` with Node.js-based test runners like Vitest running in ESM mode.

## Solution

Add the `.js` extension to the two `react-reconciler/constants` imports:
- `packages/react/src/reconciler/host-config.ts`
- `packages/react/src/reconciler/reconciler.ts`

This is backward compatible - the `.js` extension works correctly in both Bun and Node.js.

## Testing

Tested with Vitest in a pnpm monorepo - the import errors are resolved.


kommander:
@msmps you can probably better judge the impact.

kommander:
I just realized "which works in Bun but fails in Node.js", but there is no way currently to use opentui with nodejs as it is bun FFI only. So I don't see what this would solve?

remorses:
Yeah the user probably tried running it in Node.js and this was the first visible issue

kommander:
Maybe it should detect if tried to be run on node very early and fail hard with an error message that it is Bun exclusive currently. Closing this as node support is far away I think.

kommander:
Maybe it should detect if tried to be run on node very early and fail hard with an error message that it is Bun exclusive currently. Closing this as node support is far away I think.

PR #512, opened 2026-01-11
fix(utf8): use col_offset for CJK word wrap boundary detection

### LINKED ISSUE #255: bug: CJK char corruption in text rendering

kskang:
CJK characters have display width 2. But the code counted them as width 1 when calculating word wrap boundaries. This caused incorrect wrap positions.

## Changes
- Add `col_offset` field to `WrapBreak` struct. This field tracks display column position.
- Use `col_offset` instead of `char_offset` in wrap boundary detection.
- Use dynamic width calculation for non-ASCII characters in buffer drawing.

## Test
- Add native test for CJK word wrap with mixed scripts (Korean, English, Chinese, Japanese).

Fixes #255

### DIFF
--- a/packages/core/src/zig/buffer.zig
+++ b/packages/core/src/zig/buffer.zig
@@ -865,7 +865,7 @@ pub const OptimizedBuffer = struct {
             } else {
                 if (byte_offset >= text.len) break;
                 grapheme_bytes = text[byte_offset .. byte_offset + 1];
-                g_width = 1;
+                g_width = @intCast(utf8.getWidthAt(text, byte_offset, tab_width, self.width_method));
                 byte_offset += 1;
             }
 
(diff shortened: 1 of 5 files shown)


kommander:
Good find. I've got something similar on a branch, but I wasn't happy with the performance impact on `measureForDimensions`. Maybe that's now negligible with its caching. I'll bench some more and bring that in somehow the coming days.

kommander:
I think I fixed that with a little bit different approach in #529, it is a little bit more performance aware. Thank you for looking into this!

PR #515, opened 2026-01-12
fix(renderer): disable mouse tracking on destroy

### LINKED ISSUE #509: Destroy of renderer doesn't call disableMouse
I'm writing the TUI that finishes with a clear log of what is done, so I have a TUI that after destroying is logging everything important to a stdout/stderr.

The error that I have is that after I destroy the renderer, mouse events are leaking into input of the next command. After investigate I found that `destroy` and `finalizeDestroy` doesn't call `disableMouse`, so mouse escape sequence is not sent.

simonklee:
The destroy path was not explicitly disabling mouse tracking before tearing down native resources. 

Best-effort mouse cleanup now lives in native terminal shutdown instead of
TypeScript destroy state. Track whether mouse was ever enabled so teardown
still emits disable sequences after state drift, while skipping unnecessary
resets when mouse was never active.# Please enter the commit message for your
changes.

Fixes: #509

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -2219,6 +2219,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
       },
       true,
     )
+    this._useMouse = false
     this.setCapturedRenderable(undefined)
 
     this.stdin.removeListener("data", this.stdinListener)
--- a/packages/core/src/tests/renderer.useMouse.test.ts
+++ b/packages/core/src/tests/renderer.useMouse.test.ts
@@ -44,4 +44,18 @@ describe("useMouse configuration", () => {
 
     renderer.destroy()
   })
+
+  test("destroy disables mouse tracking", async () => {
+    const { renderer } = await createTestRenderer({
+      useMouse: true,
+      exitOnCtrlC: false,
+      useAlternateScreen: false,
+    })
+
+    expect(renderer.useMouse).toBe(true)
+
+    renderer.destroy()
+
+    expect(renderer.useMouse).toBe(false)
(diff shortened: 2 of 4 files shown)


kommander:
Interesting. Just out of curiosity, do you know in what edge cases that would happen? I wonder because other states might be out of sync then as well.

simonklee:
> Interesting. Just out of curiosity, do you know in what edge cases that would happen? I wonder because other states might be out of sync then as well.

No specific repro beyond #509, but the class of edge cases is 'write didn't reach the TTY.' errors are swallowed, so if stdout is closed/piped or a write fails, the terminal can still be in mouse mode while state says false. This just sends a best‑effort disable before teardown. Other state flags could drift under the same failure mode, but i haven't looked into it extensively.

kommander:
Got it. But the native teardown is called always, so it could just unconditionally disable mouse?

simonklee:
> Got it. But the native teardown is called always, so it could just unconditionally disable mouse?

Let me go through the call paths again to make sure i understand it.

erikjoachim:
hey! found this while tracking and figuring a mouse escape sequence issue in opencode.

noticed the fix checks `this._useMouse` which only covers click tracking (?1000). since `enableMouse()` passes both `_useMouse` AND `enableMouseMovement` to the native lib, should the destroy cleanup also handle both? 

mouse movements also leak escape sequences after exit, so maybe just call disableMouse() unconditionally?

```diff
- if (this._useMouse) {
-   this.disableMouse()
- }
+ this.disableMouse()
```

also looks like this is in draft - any plans to move it forward? 😄

PR #517, opened 2026-01-13
docs: update solid README to provide some base level documentation, similarly to the react package README

bryceosterhaus:
I was using solid with opentui for the first time and had a hard time finding docs for some of the basic functionality and components. These basic docs might help some people onboard a bit quicker. I copied the general format of the docs in the react package.

If this isn't needed, feel free to close.

### DIFF
--- a/packages/solid/README.md
+++ b/packages/solid/README.md
@@ -53,3 +53,136 @@ await Bun.build({
   },
 })
 ```
+
+## Table of Contents
+
+- [Core Concepts](#core-concepts)
+  - [Components](#components)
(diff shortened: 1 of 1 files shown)


kommander:
If it makes sense for @Adictya and @fezproof, I think it helps.

Adictya:
Hey @bryceosterhaus, thanks for this! It looks pretty inline with what matt did for react and makes sense to me.

bryceosterhaus:
FYI: Format was failing so I ran format and squashed the changes into my original commit.

Adictya (approved):

PR #520, opened 2026-01-14
feat: detect image events on paste with ctrl +v or command v

tyrellshawn:
Tested using the examples but havent done any integration tests with opencode yet.


Hona:
I think we currently have paste detection in a few places - including in opencode repo.

How does this differ to that?

kommander:
@Hona I told them that I want to provide that on a library level, as apps would have to re-implement that and pasting binary data like images seems quite common.

packages/core/src/lib/KeyHandler.ts
```diff
@@ -61,13 +62,23 @@ export class KeyEvent implements ParsedKey {
...
 
+export type PasteEventInit = {
+  data: Buffer
+  text?: string
+  fileType?: string
+}
+
 export class PasteEvent {
-  text: string
+  data: Buffer
```
kommander, line 70: I wonder if the PasteEvent should maybe only have `data` and we drop `text` as for large pastes like images it would always toString duplicate the data, while consumers of the paste event can do the toString when it is actually text content and not a file?

tyrellshawn: I agree but how would we define the data interface. As long as we know it can do toString we are good? I feel like having a getFileType method in the interface would be good too?

kommander: Just as `data: Buffer`? The PasteEvent has the optional fileType.

tyrellshawn: updated

Issue #518, opened 2026-01-13
Feature request: select-all (Cmd+A) for Textarea


danolekh:
browser-like "select all" in textarea - press cmd+a to select everything

i  have a working implementation:
- Added `"select-all"` to TextareaAction
- Added keybinding: `{ name: "a", super: true, action: "select-all" }`
- Added `selectAll()` method

i mostly used claude to navigate this repo. i don't have much terminal dev experience - i usually work in browsers - so I wanted this familiar behavior

What I tested (macOS only)

| Terminal | Works? |
|----------|--------|
| Kitty | ✅ yes |
| Alacritty | ✅ yes |
| Ghostty | ❌ no - intercepts Cmd+A for terminal buffer selection |

Questions

1. Does this make sense for the project?
2. On Windows/Linux, `super+a` maps to Win+A which is weird. So I'm not sure about this feature
3. What's the tui philisophy regarding keybindings overall - I see some emcas bindings, and, in the same components browser-like bindings

kommander:
1. Totally
2. Mappings can be overridden by applications, I'd add maybe another default mapping for windows? Not sure what it is there. The default mappings should maybe be platform dependent.
3. People have all kinds of conflicting expectations, so it will work fine for some while others expect something else. There is no defacto standard for the terminal, it depends on the platform, terminal emulator and the philosophy the users comes from. Impossible to meet them all with a default. It's configurable/customizable, that's one philosophy I'd say.

danolekh:
@kommander 
have a question regarding arrow keys behaviour when something is selected

in browsers, if something is selected, and you click arrow left / right you go to respective edges of that selection
in terminals, your selection doesn't affect this behiour - cursor just moves from the place it was at that moment

is this something we shouldn't modify?

https://github.com/user-attachments/assets/f26319af-96c9-4463-9e12-238ac74a2be2

kommander:
It could have a similar behaviour like that, yes.

PR #522, opened 2026-01-14
fix: select buffer home/end behaviour when textarea's content overflows

### LINKED ISSUE #521: bug: weird select behaviour when textarea is scrollable
when trying to go through different edge cases for #518 feature 
i've stumbled on annoying behaviour: when you are in select mode (shift is toggled) and move your cursor to position off the visible textarea content, it doesn't behave normally
here is an example with shift and cmd+arrow keys:
when textarea is scrollable, it doesn't work as expected
when the whole content is fully contained within visible textarea content, it it works as expected

https://github.com/user-attachments/assets/204d14f2-66fa-4cb1-8647-b70fa3efe942

danolekh:
fixes #521 and makes possible to properly implement #518 

overview from clod:

Problem: Shift+cmd+arrow keys (select to buffer home/end) didn't work correctly in scrollable Textareas because selection anchor coordinates were captured relative to the wrong viewport position.

Root Cause: updateSelectionForMovement stored selection anchor using viewport-relative visual coordinates. When the viewport scrolled during a large cursor jump, the anchor became invalid.

Solution:
1. Store selection anchor using logical coordinates (row, col) which are document-absolute
2. Added scrollToCursor() to explicitly scroll viewport to cursor (bypassing the normal scroll-skip during selection)
3. After movement, convert anchor logical coords to visual coords based on current viewport

fix tested in alacritty & kitty terminals

preview
https://github.com/user-attachments/assets/7a61e7f2-6a87-4cbf-843f-b760d9619f76

### DIFF
--- a/packages/core/src/renderables/EditBufferRenderable.ts
+++ b/packages/core/src/renderables/EditBufferRenderable.ts
@@ -63,7 +63,6 @@ export abstract class EditBufferRenderable extends Renderable implements LineInf
   private _scrollSpeed: number = 16
 
   private _selectionAnchorLogical: { row: number; col: number } | null = null
-  private _cursorRowBeforeMovement: number | null = null
 
   public readonly editBuffer: EditBuffer
   public readonly editorView: EditorView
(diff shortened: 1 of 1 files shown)


kommander:
Closed in favor of #542

PR #529, opened 2026-01-15
Initial text truncation + CJK++ word wrap fix

kommander:


zenyr:
This should fix :
- #255 
Checking if I can still replicate the issue on the main branch...

PR #532, opened 2026-01-16
Add support for PuTTY ALT+arrow key sequences

tobwen:
Adds recognition for double ESC sequences generated by PuTTY terminal when ALT+arrow keys are pressed (format: ESC ESC [A-D).

Updates the keypress parser to properly identify these sequences and mark them with meta/option flags while ensuring regular arrow keys still work correctly.

Enhances terminal compatibility for users of PuTTY in default mode.


kommander:
This only tests parseKeypress, but does the stdin-buffer handle doubled escapes?

simonklee:
Solved by #770

PR #539, opened 2026-01-16
feat(input): migrate to extend textarea

msmps:
## Summary

Refactor `InputRenderable` to extend `TextareaRenderable`, eliminating ~200 lines of duplication.

**Key changes:**
- Input now inherits text editing, cursor management, and key binding logic from Textarea
- Action names aligned with Textarea conventions (`line-home`, `line-end`, `backspace`, `delete`)
    - **BREAKING CHANGE?**

## Test Plan

- [x] Existing Input tests updated and passing
- [x] Manual test

### DIFF
--- a/packages/core/src/renderables/Input.test.ts
+++ b/packages/core/src/renderables/Input.test.ts
@@ -181,7 +181,7 @@ describe("InputRenderable", () => {
       })
 
       input.focus()
-      input.cursorPosition = 1 // Move cursor after 'e'
+      input.cursorOffset = 1 // Move cursor after 'e'
 
       mockInput.pressKey("DELETE")
       expect(input.value).toBe("hllo")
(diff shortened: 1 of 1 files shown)


Hona:
This would be a good pr to try my snapshot testing on 👀

msmps:
> This would be a good pr to try my snapshot testing on 👀 

Take it away @Hona 🙌

packages/core/src/renderables/Input.ts
```diff
@@ -63,379 +27,188 @@ export enum InputRenderableEvents {
...
+    return true
   }
 
-  public set textColor(value: ColorInput) {
-    const newColor = parseColor(value ?? this._defaultOptions.textColor)
-    if (this._textColor !== newColor) {
-      this._textColor = newColor
-      this.requestRender()
-    }
+  /** @deprecated Use `cursorOffset` instead */
```
kommander: I think we can make it breaking while still alpha officially? Just have to make sure opencode uses it right.

msmps: removed entirely! 👍

hansvdorno7 (approved):

PR #545, opened 2026-01-16
refactor(border): add parseBorderStyle function for border style vali…

### LINKED ISSUE #543: Bug: Invalid borderStyle value causes segmentation fault
## Environment
- macOS (Apple Silicon) - Darwin arm64
- Bun 1.3.5
- @opentui/core 0.1.73
- @opentui/react 0.1.73

## Description

When using an invalid `borderStyle` value (e.g., `"round"` instead of the valid `"rounded"`), the application crashes with a segmentation fault instead of showing an error or falling back to a default style.

## Reproduction

```tsx
import { createRoot } from "@opentui/react";

// This crashes with segfault:
createRoot(renderer).render(
  <box border borderStyle="round" width={30} height={5}>
    <text>Invalid borderStyle crashes</text>
  </box>
);
```

## Expected Behavior

Either:
1. TypeScript should catch invalid values at compile time (it does, but runtime should still handle it gracefully)
2. Runtime should throw a clear error message like "Invalid borderStyle: 'round'. Valid values are: single, double, rounded, heavy"
3. Fall back to the default border style

## Actual Behavior

Segmentation fault at address 0x10 or 0x48:

```
================================================================================
Bun v1.3.5 (a6ba3035) macOS Silicon

Segmentation fault at address 0x10
Segmentation fault

oh no! Bun has crashed. This indicates a bug in Bun, not your code.
```

## Analysis

The crash occurs in the native Zig library (`libopentui.dylib`). The invalid borderStyle value `"round"` likely causes an out-of-bounds lookup in the `BorderChars` record which expects only: `"single" | "double" | "rounded" | "heavy"`.

## Workaround

Use only valid borderStyle values: `"single"`, `"double"`, `"rounded"`, or `"heavy"`.

## Suggestion

Add input validation for borderStyle in the JavaScript layer before passing to the native Zig code:

```typescript
const validStyles = ["single", "double", "rounded", "heavy"];
if (!validStyles.includes(borderStyle)) {
  console.warn(\`Invalid borderStyle "${borderStyle}", using "single"\`);
  borderStyle = "single";
}
```

msmps:
## Description
- fix segfault when invalid `borderStyle` values are passed to box components
- added `parseBorderStyle()` validation that gracefully falls back to "single" with a warning
- closes #543

### DIFF
--- a/packages/core/src/renderables/Input.test.ts
+++ b/packages/core/src/renderables/Input.test.ts
@@ -181,7 +181,7 @@ describe("InputRenderable", () => {
       })
 
       input.focus()
-      input.cursorPosition = 1 // Move cursor after 'e'
+      input.cursorOffset = 1 // Move cursor after 'e'
 
       mockInput.pressKey("DELETE")
       expect(input.value).toBe("hllo")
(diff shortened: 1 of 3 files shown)


kommander:
Maybe one regression test that just makes sure it falls back and does not crash when setting something invalid via constructor options or the setter?

kommander:
@remorses something wrong with the preview setup or cloudflare outage? Going to merge that anyway.

PR #546, opened 2026-01-17
feat(clipboard): add OSC 52 clipboard support for SSH/remote sessions

Son0fSun:
## Summary

- Add cross-platform clipboard support using OSC 52 escape sequences
- Enable copy/paste operations to work over SSH connections
- Replace macOS-only `pbcopy` in console-demo with portable OSC 52

## Problem

Clipboard operations don't work when running OpenTUI applications over SSH because the app doesn't have access to the local clipboard. This breaks copy/paste workflows for remote development.

Fixes https://github.com/sst/opencode/issues/6111
Fixes https://github.com/anomalyco/opencode/issues/2773

## Solution

OSC 52 is an ANSI escape sequence that allows terminal applications to interact with the system clipboard through the terminal emulator. The terminal handles the clipboard operation locally, so it works seamlessly over SSH.

### New API

```typescript
import { copyToClipboard, clearClipboard, isOsc52Supported } from "@opentui/core"

// Copy text to clipboard
copyToClipboard("Hello, world!")

// With options
copyToClipboard(text, { 
  target: "clipboard",  // or "primary", "secondary"
  silent: true          // suppress errors
})

// Check support
if (isOsc52Supported()) {
  // OSC 52 likely supported
}
```

### Supported Environments

- **Direct terminals**: iTerm2, Kitty, Alacritty, Windows Terminal, foot, WezTerm
- **tmux**: Full passthrough support including nested sessions
- **GNU Screen**: Proper DCS passthrough format

## Technical Details

OSC 52 format: `ESC ] 52 ; <target> ; <base64-content> ST`

For terminal multiplexers, the sequence must be wrapped in DCS passthrough:
- tmux: `ESC P tmux; ESC ESC <seq> ESC \`
- screen: `ESC P ESC ESC <seq> ESC \`

## Changes

| File | Description |
|------|-------------|
| `packages/core/src/lib/clipboard.ts` | New OSC 52 clipboard module |
| `packages/core/src/lib/clipboard.test.ts` | Comprehensive test suite (11 tests) |
| `packages/core/src/lib/index.ts` | Export clipboard functions |
| `packages/core/src/examples/console-demo.ts` | Use OSC 52 instead of pbcopy |

## Test plan

- [x] Unit tests pass (`bun test packages/core/src/lib/clipboard.test.ts`)
- [x] Tested copy operation over SSH session
- [x] Verified paste works on local machine (macOS)
- [ ] Test on Windows Terminal
- [ ] Test inside tmux
- [ ] Test inside GNU Screen

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/packages/core/src/zig/terminal.zig
+++ b/packages/core/src/zig/terminal.zig
@@ -706,21 +706,17 @@ pub fn writeClipboard(self: *Terminal, tty: anytype, target: ClipboardTarget, pa
 
     const osc52 = stream.getWritten();
 
-    // Check for tmux/screen and wrap accordingly
     // Use self.in_tmux which is set by checkEnvironmentOverrides() considering
     // env vars, xtversion response, and remote option
     const is_tmux = self.in_tmux or self.isXtversionTmux();
 
     if (is_tmux) {
-        // tmux requires DCS passthrough with tmux; prefix
         // For nested tmux, we use a fixed level of 1 as we don't have access
         // to env vars here (by design - detection already happened in checkEnvironmentOverrides)
         // In practice, single-level wrapping works for most cases
         var wrapped_buf: [4096]u8 = undefined;
         var wrapped_stream = std.io.fixedBufferStream(&wrapped_buf);
         const wrap_writer = wrapped_stream.writer();
-
-        // Double all ESC characters
         for (osc52) |c| {
             if (c == '\x1b') {
                 try wrap_writer.writeByte('\x1b');
@@ -729,25 +725,20 @@ pub fn writeClipboard(self: *Terminal, tty: anytype, target: ClipboardTarget, pa
         }
         const doubled = wrapped_stream.getWritten();
 
-        // Wrap in DCS with tmux; prefix
         try tty.writeAll(ansi.ANSI.tmuxDcsStart);
         try tty.writeAll(doubled);
         try tty.writeAll(ansi.ANSI.tmuxDcsEnd);
     } else if (self.opts.remote) {
-        // In remote mode, don't check STY - assume no passthrough needed
         try tty.writeAll(osc52);
     } else {
-        // Check for GNU Screen (STY env var)
         var env_map = std.process.getEnvMap(std.heap.page_allocator) catch return;
         defer env_map.deinit();
 
         if (env_map.get("STY")) |_| {
-            // GNU Screen requires DCS passthrough without the tmux; prefix
             var wrapped_buf: [2048]u8 = undefined;
             var wrapped_stream = std.io.fixedBufferStream(&wrapped_buf);
             const wrapped_writer = wrapped_stream.writer();
 
-            // Double all ESC characters
             for (osc52) |c| {
                 if (c == '\x1b') {
                     try wrapped_writer.writeByte('\x1b');
@@ -756,7 +747,6 @@ pub fn writeClipboard(self: *Terminal, tty: anytype, target: ClipboardTarget, pa
             }
             const doubled = wrapped_stream.getWritten();
 
-            // Wrap in DCS
             try tty.writeAll(ansi.ANSI.screenDcsStart);
             try tty.writeAll(doubled);
             try tty.writeAll(ansi.ANSI.screenDcsEnd);


packages/core/src/lib/clipboard.ts
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Check if the current environment likely supports OSC 52.
+ *
+ * This is a heuristic check based on environment variables and terminal type.
+ * It cannot guarantee OSC 52 support, but can help determine if it's worth trying.
+ *
+ * @returns true if OSC 52 is likely supported
+ */
+export function isOsc52Supported(): boolean {
```
kommander: I would add that to native capabilities detection in `src/zig/terminal.zig`.

packages/core/src/lib/clipboard.ts
```diff
@@ -0,0 +1,192 @@
...
+ * import { copyToClipboard } from "@opentui/core"
+ *
+ * // Copy text to clipboard
+ * copyToClipboard("Hello, world!")
+ *
+ * // Copy to primary selection (X11)
+ * copyToClipboard("Hello, world!", { target: "p" })
+ * ```
+ */
+export function copyToClipboard(text: string, options: ClipboardOptions = {}): boolean {
```
kommander: I'd probably like to do that in native as well, writing to the buffer there. Perspectively it will support writing to any JS stream through native to support multiple ssh sessions etc. like multiple renderers in the same process.

CookiC:
Bug Report: OSC 52 clipboard not working in SSH sessions
Issue: OSC 52 clipboard functionality introduced in PR #546 does not work when using opencode over SSH, even though the local terminal supports OSC 52.
Affected Version: opencode 1.1.49 (regression - works in 1.1.46)
Environment:
- SSH session to remote server
- Local terminal: iTerm2 / WezTerm / Kitty (all support OSC 52)
- Remote server: running opencode
Root Cause:
In packages/core/src/zig/terminal.zig, the checkEnvironmentOverrides() function has an early return when opts.remote is true:
```
fn checkEnvironmentOverrides(self: *Terminal) void {
    self.in_tmux = false;
    self.skip_graphics_query = false;
    self.caps.bracketed_paste = true;
    if (self.caps.rgb) {
        self.caps.hyperlinks = true;
    }
    if (self.opts.remote) {
        return;  // ❌ PROBLEM: Early return skips OSC 52 detection!
    }
    var env_map = std.process.getEnvMap(std.heap.page_allocator) catch return;
    defer env_map.deinit();
    // OSC 52 detection below (lines 405-437) never executes in SSH:
    if (!self.caps.osc52 and !self.term_info.from_xtversion) {
        if (env_map.get("WT_SESSION") != null) {
            self.caps.osc52 = true;
        }
        if (!self.caps.osc52 and (self.in_tmux or env_map.get("STY") != null)) {
            self.caps.osc52 = true;
        }
        // ... more OSC 52 terminal detection ...
    }
}
```
Why This Breaks OSC 52:
The early return at line ~268-271 causes all environment variable checks to be skipped, including OSC 52 capability detection (lines 405-437). This means self.caps.osc52 is never set to true in SSH sessions, even when:
- The local terminal supports OSC 52
- The SSH environment variables are present
- Modern terminal emulators (iTerm2, WezTerm, Windows Terminal, etc.) are being used
Evidence:
1. Neovim's OSC 52 implementation works in the same SSH environment (from ~/.config/nvim/lua/config/options.lua):
```
if vim.env.SSH_CONNECTION then
  vim.g.clipboard = {
    name = "OSC 52",
    copy = {
      ["+"] = require("vim.ui.clipboard.osc52").copy("+"),
      ["*"] = require("vim.ui.clipboard.osc52").copy("*"),
    },
    paste = {
      ["+"] = require("vim.ui.clipboard.osc52").paste("+"),
      ["*"] = require("vim.ui.clipboard.osc52").paste("*"),
    },
  }
end
```
Neovim detects SSH_CONNECTION and enables OSC 52 without checking if it's in a "remote" environment.
3. The clipboard wrapper (packages/core/src/lib/clipboard.ts) checks isOsc52Supported() which returns false because caps.osc52 was never enabled.
Expected Behavior:
When using opencode over SSH with a terminal that supports OSC 52 (iTerm2, WezTerm, Kitty, Windows Terminal, etc.), the copy-to-clipboard function should work by sending OSC 52 sequences that the local terminal handles.
Actual Behavior:
Copy-to-clipboard does not work in SSH sessions, even with OSC 52-capable terminals.
Suggested Fix:
Remove or move the early return check so that OSC 52 detection can run even in remote/SSH environments:
```
fn checkEnvironmentOverrides(self: *Terminal) void {
    // ... existing code ...
    // ❌ REMOVE THIS EARLY RETURN
    // if (self.opts.remote) {
    //     return;
    // }
    var env_map = std.process.getEnvMap(std.heap.page_allocator) catch return;
    defer env_map.deinit();
    // Allow OSC 52 detection even in SSH/remote environments
    // because modern terminals support OSC 52 over SSH
    if (!self.caps.osc52 and !self.term_info.from_xtversion) {
        // ... OSC 52 detection code ...
    }
}
```
Or, add specific SSH environment variable detection (like Neovim does):
```
if (env_map.get("SSH_CONNECTION") != null or env_map.get("SSH_CLIENT") != null) {
    // SSH environment - still check for OSC 52 capable terminals
    self.caps.osc52 = true; // Assume modern terminals support it
}
```
Discovered by: Kimi 2.5 (AI assistant)
Verified: Not verified

Issue #528, opened 2026-01-15
bug: terminal is not properly reset if the process terminates abruptly.


rwese:
Currently when an OpenTui application crashes, exits unclean, the terminal does not properly respond to escape sequences.

I gave the ai a shot to fix it, since I only barely know zig.

If this is close to what is needed I will clean this up, but might need some guidance.

Before fix:

https://github.com/user-attachments/assets/cc992215-4123-4814-9bfe-172c643a3663

After fix:

https://github.com/user-attachments/assets/c09b1621-348e-43de-ac4e-8000f5f93192

Commit/Changes:

https://github.com/rwese/opentui/commit/3b918893d2c2ef705ac8df158868011c284d94ac

kommander:
The additionally handled signals make sense. 

As for the exit handler, it did that before, but it was interfering with effect.ts shutdown and the consense was that shutdown should be handled on application level and the application has to call `renderer.destroy` properly.

Similarly for error handling, if for an unhandled error opentui decides to destroy the renderer or reset the terminal, but the application wants to handle the error and move on, the app would be broken.

rwese:
I've tried to clean it up and simplify. Now it does listen only on Exit, which should be the very last event bun gives us.

https://github.com/rwese/opentui/commit/3ebe2ad3850a79d2b6321cfb5c5c5d89f0aebe68

Another way to implement this, without breaking any other handling, which applications may already employ, would be to expose it as a function which applications may use. 

Maybe I am just not well versed in how bun vs zig vs terminal emulators work.. 

What i have forgot to mention, i run into this problem frequently with opencode. Everytime when bun fails to install a plugin, _read: when i break something again_, the `opencode` dies and the terminal is borked. I resorted to use `opencode <log etc.>; reset` to get around this.

kommander:
Cool. The `renderer.destroy()` method is responsible for terminal cleanup and should be the method called by applications for shutdown.

I think opencode needs proper `uncaughtError` handling then and call `renderer.destroy()` before it exits.

rwese:
I will look at the way opencode does it and, if it still happens, attempt a fix there.

When you consider this a won't fix, or unrelated issue, we can close this issue.

kommander:
I would add the additional default signals to handle at least.

simonklee:
Are we missing something to close this issue @kommander ?

kommander:
[...]

PR #547, opened 2026-01-17
feat(core): add SIGPIPE, SIGBUS, SIGFPE to default exit signals

rwese:
See #528 

Adds couple more of possible exit signals that might occur.


kommander:
I don't think Bun supports all of these yet, but should be fine to just hook them up if it does.

Issue #553, opened 2026-01-19
Proposal: The Grid™


necropheus:
# Idea
This is an idea I've been working on over the weekend, a grid component that has baked-in logic for keyboard-based layout navigation.

# Problem: navigating the UI with your keyboard
Right now, if you want to navigate through the UI with your keyboard, you have to do 2 things:

1. Implement your own focus system so you know *where* you're focusing
2. Navigate the UI with `useKeyboard`

But that comes with some problems (in my opinion):

## Tracking
Tracking your own focus system is not trivial. The more things you have in your UI, the more complex it gets: `if focus is "foo" but not "bar" and flag "qux" === "mom"`, etc. It's basically a global state that grows in complexity as your app grows.

## Key handling
You have to have multiple scattered `useKeyboard` calls through your project to manage their respective UIs, but that potentially becomes a performance issue, and it's hard to track (if you change something regarding your focus system, you have to remember to change the `useKeyboard` calls that use said system), not to mention you probably have to repeat logic everywhere, meaning, each `useKeyboard` has to go `if (globalFocus === myContext)`.
An alternative would be to have just one giant `useKeyboard` call. That would solve the scattered calls and the repeated logic, but it would grow in complexity as your app grows, making it again hard to maintain.

## Logical UI bugs
It's prone to bugs. The combination of multiple `useKeyboard` calls and the custom focus system has a high potential for logical bugs where a specific combination of keys and focus states breaks your whole UI

# Partial solution (that should also be implemented in general, not just for the grid): focusable boxes
Now, there's one thing that could solve some of these problems: boxes being able to be focused, thus being able to receive key events.

Why would this solve some of these issues?

## No more `useKeyboard`
No need for multiple `useKeyboard` calls. Each box has `onKeyDown`, and acts as its own focus system. Let's imagine this:

These are boxes. Arrays represent the fact that the UI has multiple elements:
```
A => [B, ...] => [C, ...] => [D, ...]
```
- **D** has `onKeyDown` to capture left and right key events, because its UI goes left to right
- **C** has `onKeyDown` to capture up and down events, because its UI goes up and down
- **B** has `onKeyDown` to capture left and right key events, same as **D**, because its UI goes left to right too
- **A** has `onKeyDown` to capture ctrl+{arrowKey} because its UI goes in all directions

1. Now, if I'm focusing **D** and press the left key, **D** captures that event and moves its internal component's focus (which would be yet other focusable boxes), and stops propagation.
[...]

PR #554, opened 2026-01-19
Draft: The Grid™

necropheus:
This is just to show the files for this proposal https://github.com/anomalyco/opentui/issues/553

### DIFF
--- a/packages/core/src/renderables/grid/constants.ts
+++ b/packages/core/src/renderables/grid/constants.ts
@@ -1,4 +1,4 @@
-import { GridKeys } from "./types"
+import type { GridKeys } from "./types"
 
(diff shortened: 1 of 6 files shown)


remorses:
Given this renderable does not touch other opentui code you can put this in your own codebase and publish as an npm package. If the API stabilizes and others find it useful later on then it makes sense merging into core

necropheus:
> Given this renderable does not touch other opentui code you can put this in your own codebase and publish as an npm package. If the API stabilizes and others find it useful later one then it makes sense merging into core

I could do that too yes

simonklee:
https://github.com/anomalyco/opentui/issues/553#issuecomment-4026094787

PR #555, opened 2026-01-19
fix(core): recognize Ctrl+Backspace on Windows Terminal and WSL

Info-Cado:
## Summary

- Fix Ctrl+Backspace not triggering `delete-word-backward` on Windows Terminal and WSL

## Problem

Windows Terminal and WSL terminals send different byte sequences for backspace:
- **Regular Backspace**: `\x7f` (ASCII 127, DEL)
- **Ctrl+Backspace**: `\b` (ASCII 8)

The current parser treats both as plain backspace without setting `ctrl: true`, so keybindings like `ctrl+backspace` → `delete-word-backward` never trigger.

### Additional WSL Complexity

In WSL, `process.platform` returns `'linux'`, not `'win32'`, even though Windows Terminal sends Windows-style key sequences. This required detecting WSL via environment variables (`WSL_DISTRO_NAME`, `WSL_INTEROP`).

## Solution

Added `isWindowsTerminal()` helper that detects:
1. Native Windows (`process.platform === 'win32'`)
2. WSL (`WSL_DISTRO_NAME` or `WSL_INTEROP` environment variables)

| Platform | `\b` Handling | Result |
|----------|-----------------|--------|
| Windows/WSL | Ctrl+Backspace | `{ name: "backspace", ctrl: true }` |
| macOS/Linux | Ctrl+H (unchanged) | `{ name: "h", ctrl: true }` |

## Testing

- All 41 tests pass
- Tests use the same `isWindowsTerminal()` helper for platform-aware assertions
- Manually verified in WSL with Windows Terminal

Fixes anomalyco/opencode#6991

### DIFF
--- a/packages/core/src/lib/parse.keypress.test.ts
+++ b/packages/core/src/lib/parse.keypress.test.ts
@@ -2,6 +2,13 @@ import { test, expect } from "bun:test"
 import { parseKeypress, nonAlphanumericKeys, type ParsedKey, type KeyEventType } from "./parse.keypress"
 import { Buffer } from "node:buffer"
 
+const isWindowsTerminal = (): boolean => {
+  if (process.platform === "win32") return true
+  if (process.env.WSL_DISTRO_NAME) return true
+  if (process.env.WSL_INTEROP) return true
+  return false
+}
+
 test("parseKeypress - basic letters", () => {
   expect(parseKeypress("a")).toEqual({
     name: "a",
@@ -133,7 +140,7 @@ test("parseKeypress - special keys", () => {
   expect(backspaceB.sequence).toBe("\b")
   expect(backspaceB.raw).toBe("\b")
   expect(backspaceB.source).toBe("raw")
-  if (process.platform === "win32") {
+  if (isWindowsTerminal()) {
     expect(backspaceB.name).toBe("backspace")
   } else {
     expect(backspaceB.name).toBe("h")
@@ -769,7 +776,7 @@ test("parseKeypress - backspace key with modifiers (modifyOtherKeys format)", ()
 
 test("parseKeypress - raw backspace sequences (platform-aware)", () => {
   const ctrlHOrBackspace = parseKeypress("\b")!
-  if (process.platform === "win32") {
+  if (isWindowsTerminal()) {
     expect(ctrlHOrBackspace.name).toBe("backspace")
     expect(ctrlHOrBackspace.ctrl).toBe(true)
   } else {
@@ -779,7 +786,7 @@ test("parseKeypress - raw backspace sequences (platform-aware)", () => {
   expect(ctrlHOrBackspace.meta).toBe(false)
 
   const altCtrlHOrBackspace = parseKeypress("\x1b\b")!
-  if (process.platform === "win32") {
+  if (isWindowsTerminal()) {
     expect(altCtrlHOrBackspace.name).toBe("backspace")
     expect(altCtrlHOrBackspace.ctrl).toBe(true)
   } else {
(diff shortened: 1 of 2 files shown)


kommander:
I think that could make sense, it does not play well with existing expectations though it seems.

Info-Cado:
Thanks for the feedback @kommander!

You're right - I investigated the CI failures and I see the issue. The change breaks existing expectations around how `\b` (ASCII 8) is interpreted:

**Current behavior**: `\b` → `{ name: "h", ctrl: true }` (via the generic ctrl+letter handler)
- This then matches keybindings like `{ name: "h", ctrl: true, action: "backspace" }` for single-char deletion

**My change**: `\b` → `{ name: "backspace", ctrl: true }`
- This matches `{ name: "backspace", ctrl: true, action: "delete-word-backward" }` and triggers word deletion

The root cause is **terminal semantics divergence**:
- **Unix/macOS**: `\b` = Ctrl+H (traditionally "h" with ctrl modifier)
- **Windows Terminal**: `\b` = Ctrl+Backspace (a distinct key combination)

### Options I see:

1. **Platform detection**: Only interpret `\b` as Ctrl+Backspace on Windows (`process.platform === 'win32'`). This preserves existing behavior on macOS/Linux but enables the fix for Windows/WSL users.

2. **Keep existing behavior**: Revert this change and document that Ctrl+Backspace requires Kitty keyboard protocol or modifyOtherKeys mode to work. The trade-off is that Windows Terminal users in WSL won't get word-deletion with Ctrl+Backspace.

3. **Add explicit keybinding**: Keep `\b` as Ctrl+H, but add a default keybinding `{ name: "h", ctrl: true, action: "delete-word-backward" }` to make Ctrl+H also delete words. This would be a behavior change but wouldn't affect key parsing.

What approach would you prefer? I'm happy to implement whichever makes most sense for the project.


BTW this comment was written by AI

Info-Cado:
I've pushed a fix that uses **platform detection** (Option 1).

**Changes:**
- On Windows (`process.platform === 'win32'`): `\b` → `{ name: "backspace", ctrl: true }` (enables Ctrl+Backspace word deletion)
- On macOS/Linux: `\b` → `{ name: "h", ctrl: true }` (preserves existing Ctrl+H behavior)

This should:
1. Pass all existing tests on macOS CI (behavior unchanged)
2. Enable Ctrl+Backspace on Windows Terminal/WSL users

The tests now use platform detection to validate the expected behavior on each platform.

Info-Cado:
I discovered an issue with the previous platform detection approach: **WSL reports `process.platform` as `'linux'`, not `'win32'`**, even though Windows Terminal sends Windows-style key sequences to WSL applications.

Updated the fix to also check for WSL environment variables:
- `WSL_DISTRO_NAME` - set by WSL to the distribution name
- `WSL_INTEROP` - set by WSL2 for Windows interop

This ensures Ctrl+Backspace works correctly for:
1. Native Windows (`process.platform === 'win32'`)
2. WSL1/WSL2 (detected via environment variables)
3. macOS/Linux (unchanged behavior - `\b` → Ctrl+H)

Tests updated to use the same detection logic.

packages/core/src/lib/parse.keypress.test.ts
```diff
@@ -2,6 +2,13 @@ import { test, expect } from "bun:test"
 import { parseKeypress, nonAlphanumericKeys, type ParsedKey, type KeyEventType } from "./parse.keypress"
 import { Buffer } from "node:buffer"
 
+const isWindowsTerminal = (): boolean => {
```
kommander, line 5: The tests should probably use the same method as `parse.keypress` if it's not mocking anything?

simonklee:
https://github.com/anomalyco/opentui/pull/797

PR #556, opened 2026-01-19
fix: don't re-capture console in restoreOriginalConsole

murrayju:
## Summary

The `restoreOriginalConsole` method was calling `setupConsoleCapture()` after restoring the original console, which immediately re-captured it. This prevented `console.log` output from appearing after `renderer.destroy()` was called.

## Problem

When using `createCliRenderer` and then calling `renderer.destroy()`, any `console.log` statements after the destroy call would not appear in the terminal. This is because:

1. `renderer.destroy()` calls `this._console.destroy()`
2. Which calls `deactivate()`
3. Which calls `restoreOriginalConsole()`
4. `restoreOriginalConsole` restores `global.console` but then immediately calls `setupConsoleCapture()` which re-captures it

## Solution

Remove the errant `setupConsoleCapture()` call from `restoreOriginalConsole()` so that deactivating the console capture actually restores normal console behavior.

## Testing

After this fix, the following pattern works correctly:

```typescript
const renderer = await createCliRenderer({ exitOnCtrlC: true });
// ... use renderer ...
renderer.destroy();
console.log('This message now appears!'); // Previously was captured and lost
```


kommander:
@msmps I remember darkly that there was a weird reason for this. Looking at it again it doesn't make sense. I know that there is no test coverage for this, so even though it makes sense, I am not sure what it would break. Only thing I see is that console activate/deactivate was meant for the "debug console view" and if it should capture console log calls, which is different from the "capture", which makes sure console.* calls do not interfere with stdout. Maybe it would make sense to have a `destroy` method for the console/cache etc. that resets the capture? That could be triggered when all renderers are destroyed to reset capture.

simonklee:
image

Issue #549, opened 2026-01-18
bug: selection changes on textarea viewport change


danolekh:
https://github.com/user-attachments/assets/b85ae67f-1282-4aaf-9659-ad6fb1221e2f

kommander:
Darn, that was probably not test covered and the other changes broke that. Can you reproduce that in a failing test and I'll fix it?

danolekh:
yes sir, I added it to my PR #558 (hope its ok)
I wasnt able to reproduce it using renderOnce, it feels like something in the render loop updates selection based on cursor position when it shouldn't

danolekh:
also, I dont think its a regression,
I was able to reproduce it on a79281d69efe79e8da81c68e16523e6f420eaac3 commit (0.1.63 release, 1mo old)

kommander:
This should be fixed now then with #558 ?

PR #558, opened 2026-01-20
Feat/select all textarea

danolekh:
- added ability to do select all (cmd+a / ctrl+a behaviour) for textareas #518 
- browser-like arrow left/right [behaviour with selections](https://github.com/anomalyco/opentui/issues/518#issuecomment-3765306216)
- added fix for #549

### DIFF
--- a/packages/core/src/renderables/EditBufferRenderable.ts
+++ b/packages/core/src/renderables/EditBufferRenderable.ts
@@ -761,12 +761,6 @@ export abstract class EditBufferRenderable extends Renderable implements LineInf
       return
     }
 
-    this._ctx.updateSelection(this, cursorX, cursorY)
-
-    const selection = this._ctx.getSelection()
-    if (selection) {
-      selection.isDragging = false
-      this.onSelectionChanged(selection)
-    }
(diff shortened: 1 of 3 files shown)


kommander:
@danolekh will you fix the tests or is there something more to consider?

danolekh:
@kommander yes, i will fix

danolekh:
should be fixed now

packages/core/src/renderables/EditBufferRenderable.ts
```diff
@@ -762,5 +762,11 @@ export abstract class EditBufferRenderable extends Renderable implements LineInf
     }
 
     this._ctx.updateSelection(this, cursorX, cursorY)
+
+    const selection = this._ctx.getSelection()
+    if (selection) {
+      selection.isDragging = false
+      this.onSelectionChanged(selection)
+    }
```
kommander: I need to consider this extra block for a moment, as `ctx.updateSelection` should trigger `onSelectionChanged`.

danolekh: yeah, that makes sense, i've updated the code

Issue #254, opened 2025-11-02
[DISCUSSION] should opentui.com be a website?


Vicentesan:
this issue aims to understand why the current domain [opentui.com](http://opentui.com/) redirects to the gh repo instead of an website. looking at the repo, it doesn't seem like we have a website. for that reason, i would rephrase the question: is it on the opentui roadmap to create a website? if so, i would like to help build it

0xjuanma:
Even a minimal landing page could be fine for now, with links to gh, etc if needed.

kommander:
This is up for grabs now.

We'd like to align design with www.opencode.ai, opentui will get its own proper logo. The website and documentation should live as packages in this repo and be built and deployed from there. 

As soon as we have something to publish, a target to deploy to will be provided.

remorses:
I have been working on a npm package that can render opentui TUIs to the browser via websocket, would you like to use it for the website? I can prepare a quick demo to show you how it looks like. We could even let users play with the core examples script directly in the browser

remorses:
Here is an example website created using opentui and deployed to Fly. Each request is a WebSocket connection to the server, TUI output is returned as json over WebSocket. It can handle a lot of concurrent users because it does not spawn a new process per connection.

https://opentui.fly.dev/

Vicentesan:
i will try to design something as well in the coming days

kommander:
Cool

@remorses I would like to keep it simple and pretty much static for now to get going, but I am open to trying things going forward.

remorses:
Here is an experimental website created using opentui and deployed to Fly. Each request is a WebSocket connection to the server, TUI output is returned as json over WebSocket. It can handle a lot of concurrent users because it does not spawn a new process per connection.

https://opentui.fly.dev/

Code is [here](https://github.com/remorses/opentui/blob/web-and-tunnel-and-vterm-spans/packages/website/src/server.tsx)

kommander:
@remorses it's very laggy and the 3D stuff doesn't seem to work? But it tries to render 200-1200fps somehow. I think it just crashed and restarted. 

<img width="1101" height="741" alt="Image" src="https://github.com/user-attachments/assets/e64addbe-0855-40f4-8e30-784a892c9cdb" />

simonklee:
the live TUI rendering is interesting and could work well as a playground feature down the line.

[...]

PR #561, opened 2026-01-20
Web

simonklee:
Something for #254 

https://github.com/user-attachments/assets/b245ec24-2214-416b-9e4b-bac8cbde7125

Tried to keep it simple and align with opencode.ai. Keep it simple so it's not too much to maintain. Probably remove most of the items in the docs menu until there is contents for it, added it just to show what I was thinking in terms of scope.

### DIFF
--- a/packages/web/src/components/TuiSurface.astro
+++ b/packages/web/src/components/TuiSurface.astro
@@ -37,6 +37,15 @@
       fgWeak: '#999',  // --color-text-weak
       border: '#e5e5e5', // --color-border
     };
+    
+    syntax = {
+      keyword: '#6366f1',
+      string: '#059669',
+      function: '#0891b2',
+      component: '#9333ea',
+      punctuation: '#71717a',
+      text: '#3f3f46',
+    };
 
     constructor(canvas: HTMLCanvasElement) {
       this.canvas = canvas;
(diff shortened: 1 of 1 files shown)


msmps:
super nice @simonklee

simonklee:
One of these classic 8k lines added PRs 😉

kommander:
@simonklee this is all static, right? So would it be enough to host on github pages for now and point the domain there? We can also have aws/cf.

simonklee:
> @simonklee this is all static, right? So would it be enough to host on github pages for now and point the domain there? We can also have aws/cf.

It’s only astro so GitHub pages should work. https://docs.astro.build/en/guides/deploy/github/  

Need to setup the gh action and change the astro configuration a little and make sure internal links work.

simonklee:
> > @simonklee this is all static, right? So would it be enough to host on github pages for now and point the domain there? We can also have aws/cf.
> 
> It’s only astro so GitHub pages should work. https://docs.astro.build/en/guides/deploy/github/
> 
> Need to setup the gh action and change the astro configuration a little and make sure internal links work.

Created .github/workflows/deploy.yml.

you'll need to (after merging):

1. Go to repo Settings > Pages
2. Set Source to GitHub Actions

Issue #566, opened 2026-01-21
bug: ctrl+w on a line with tabs deletes to a wrong char


CruelAddict:
Every 2+ col wide grapheme in a line makes ctrl + w / deleteWordBackward delete extra characters.

I've recorded a video where you can see me pressing ctrl + w against an indented string on the current master, vs on a branch where i fixed the issue

https://github.com/user-attachments/assets/27c326cc-02a3-4144-b2b3-6f911c6bc47b

The bug is caused by code looking for word boundaries not converting char offset to display column

kommander:
#567 should fix this.

PR #567, opened 2026-01-21
fix(core): adjust for graphemes width when calculating word boundaries

CruelAddict:
PR on issue #566 

Added a helper calculating column delta and applied it to getPrevWordBoundary and getNextWordBoundary. The algorithm is basically reused from segment_callback

Disclaimer: I don't really know zig and this is my first time using it :)

### DIFF
--- a/packages/core/src/renderables/Code.test.ts
+++ b/packages/core/src/renderables/Code.test.ts
@@ -1123,6 +1123,227 @@ test("CodeRenderable - streaming mode with drawUnstyledText=false waits for new 
   currentRenderer.stop()
 })
 
+test("CodeRenderable - onHighlight callback receives highlights and context", async () => {
+  const syntaxStyle = SyntaxStyle.fromStyles({
+    default: { fg: RGBA.fromValues(1, 1, 1, 1) },
+    keyword: { fg: RGBA.fromValues(0, 0, 1, 1) },
+  })
+
+  const mockClient = new MockTreeSitterClient()
+  mockClient.setMockResult({
+    highlights: [[0, 5, "keyword"]] as SimpleHighlight[],
+  })
+
+  let callbackInvoked = false
+  let receivedHighlights: SimpleHighlight[] | null = null
+  let receivedContext: { content: string; filetype: string | undefined; syntaxStyle: SyntaxStyle } | null = null
+
+  const codeRenderable = new CodeRenderable(currentRenderer, {
+    id: "test-code",
+    content: "const message = 'hello';",
+    filetype: "javascript",
+    syntaxStyle,
+    treeSitterClient: mockClient,
+    onHighlight: (highlights, context) => {
+      callbackInvoked = true
+      receivedHighlights = [...highlights]
+      receivedContext = { ...context }
+      return highlights
+    },
+  })
+
+  currentRenderer.root.add(codeRenderable)
+  await renderOnce()
+
+  mockClient.resolveHighlightOnce(0)
+  await new Promise((resolve) => setTimeout(resolve, 10))
+  await renderOnce()
+
+  expect(callbackInvoked).toBe(true)
+  expect(receivedHighlights).not.toBeNull()
+  expect(receivedHighlights?.length).toBe(1)
+  expect(receivedHighlights?.[0]).toEqual([0, 5, "keyword"])
+  expect(receivedContext?.content).toBe("const message = 'hello';")
+  expect(receivedContext?.filetype).toBe("javascript")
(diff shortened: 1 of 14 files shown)


kommander:
Currently `findWrapBreaks` does not resolve col offsets, that is right. Having it do that is very much overhead. My plan here was:

- let `getGraphemes` do grapheme and word break parsing in one go for a TextChunk, as it iterates anyway and can then cache both.
- `getGraphemes` is currently triggered in word wrapping logic anyway
- word breaks can then have col offsets, as the grapheme parsing iterates graphemes and col offsets anyway

This should overall be more efficient as currently doing graphemes and word breaks separately (two loops) and would make additional conversion/translation obsolete. 

cc @simonklee wdyt?

simonklee:
> cc @simonklee wdyt?

I think the perf regression is localized to getNextWordBoundary/getPrevWordBoundary, not the main render/wrap loop. From that perspective it should be okay. 

Maybe add tests for emoji/CJK and tabs in next/prev word boundary? Something like.

```diff
diff --git i/packages/core/src/edit-buffer.test.ts w/packages/core/src/edit-buffer.test.ts
index 1a54e1a..d36776d 100644
--- i/packages/core/src/edit-buffer.test.ts
+++ w/packages/core/src/edit-buffer.test.ts
@@ -417,6 +417,18 @@ describe("EditBuffer", () => {
       const next1 = buffer.getNextWordBoundary()
       expect(next1.col).toBeGreaterThan(0)
     })
+
+    it("should handle wide graphemes before boundaries", () => {
+      buffer.setText("你 好")
+      buffer.setCursorToLineCol(0, 0)
+
+      const nextBoundary = buffer.getNextWordBoundary()
+      expect(nextBoundary.col).toBe(3)
+
+      buffer.setCursorToLineCol(0, 5)
+      const prevBoundary = buffer.getPrevWordBoundary()
+      expect(prevBoundary.col).toBe(3)
+    })
   })
 
   describe("native coordinate conversion methods", () => {
diff --git i/packages/core/src/zig/tests/edit-buffer_test.zig w/packages/core/src/zig/tests/edit-buffer_test.zig
index 4b2ab74..09c33b9 100644
--- i/packages/core/src/zig/tests/edit-buffer_test.zig
+++ w/packages/core/src/zig/tests/edit-buffer_test.zig
@@ -120,6 +120,38 @@ test "EditBuffer - multiple word boundaries" {
     try std.testing.expectEqual(@as(u32, 16), cursor.col);
 }
 
+test "EditBuffer - next word boundary with wide grapheme" {
+    const pool = gp.initGlobalPool(std.testing.allocator);
+    defer gp.deinitGlobalPool();
+
+    var eb = try EditBuffer.init(std.testing.allocator, pool, .wcwidth);
+    defer eb.deinit();
+
+    // "你" = 2 cols, " " = 1 col, "好" = 2 cols
+    try eb.insertText("你 好");
+    try eb.setCursor(0, 0);
+
+    const next_cursor = eb.getNextWordBoundary();
+    try std.testing.expectEqual(@as(u32, 0), next_cursor.row);
+    try std.testing.expectEqual(@as(u32, 3), next_cursor.col);
+}
+
+test "EditBuffer - prev word boundary with wide grapheme" {
+    const pool = gp.initGlobalPool(std.testing.allocator);
+    defer gp.deinitGlobalPool();
+
+    var eb = try EditBuffer.init(std.testing.allocator, pool, .wcwidth);
+    defer eb.deinit();
+
+    // "你" = 2 cols, " " = 1 col, "好" = 2 cols
+    try eb.insertText("你 好");
+    try eb.setCursor(0, 5);
+
+    const prev_cursor = eb.getPrevWordBoundary();
+    try std.testing.expectEqual(@as(u32, 0), prev_cursor.row);
+    try std.testing.expectEqual(@as(u32, 3), prev_cursor.col);
+}
+
 test "EditBuffer - word boundary at end of line" {
     const pool = gp.initGlobalPool(std.testing.allocator);
     defer gp.deinitGlobalPool();
```

CruelAddict:
Added tests as requested.
I've also found a problem with line width calculation in segment callback (fix [here](https://github.com/anomalyco/opentui/pull/567/files#diff-ca9b30db0c25eba0062f2a40d0b01a7ce0540a5c48a5e94a5317efd6ee50244bR1137))
Recorded a little repro:
https://github.com/user-attachments/assets/012a55ae-c3c7-4af0-90bf-48df30a73d6b

And added a test catching the issue. It also felt right to add similar tests on emoji & CJK there too

kommander:
Ah cool, I've been looking for that CJK line break bug. Github has some issue running the core build and test job currently it seems. Will try again in a bit.

PR #569, opened 2026-01-21
fix(react/solid): add missing markdown element

simonklee:


msmps:
@simonklee could you add it for solid too? i think it might've been missed by @remorses

remorses:
yeah sorry about that

simonklee:
> @simonklee could you add it for solid too? i think it might've been missed by @remorses

Done.

PR #574, opened 2026-01-23
add captureSpans and getSpanLines for styled terminal output capture

remorses:
Uses the same VTerm data structures from #440 to expose terminal buffer content as structured span data.

Adds `getSpanLines()` to `OptimizedBuffer` and `captureSpans()` to test renderer.

Use case: render opentui output to web pages or images. You can render a TUI in a test renderer, extract the spans, and serialize them for web rendering or image generation.

```ts
import { createTestRenderer } from "@opentui/core/testing"

const { renderer, renderOnce, captureSpans } = await createTestRenderer({ 
  width: 80, 
  height: 24 
})

// render your components...
await renderOnce()

const data = captureSpans()
// data.lines contains styled spans with text, fg, bg, and style flags
// serialize to JSON, render to canvas, generate SVG, etc.
```


packages/core/src/buffer.ts
```diff
@@ -153,6 +164,62 @@ export class OptimizedBuffer {
     return outputBuffer.slice(0, bytesWritten)
   }
 
+  public getSpanLines(): VTermLine[] {
+    this.guard()
+    const { char, fg, bg, attributes } = this.buffers
+    const lines: VTermLine[] = []
+
+    const rgbaToHex = (r: number, g: number, b: number, a: number): string | null => {
```
kommander: There is a `rgbToHex` method already in `RGBA.ts` and I wouldn't redefine such a method inline.

kommander:
I am a little confused as to what this is being used for? The FrameBuffer has a format that allows each cell to be rendered as such with all the info needed to generate ANSI or an image.

packages/core/src/buffer.ts
remorses: Fixed, using RGBA class and rgbToHex from lib now.

remorses:
The raw buffer data is available via `renderer.currentRenderBuffer.buffers`, but using it requires knowing internal memory layout:

```ts
const { char, fg, bg, attributes } = buffer.buffers
for (let i = 0; i < buffer.width * buffer.height; i++) {
  const codepoint = char[i]
  const fgR = fg[i * 4], fgG = fg[i * 4 + 1], fgB = fg[i * 4 + 2], fgA = fg[i * 4 + 3]
  // etc...
}
```

This layout could change in the future. `getSpanLines()` provides a stable, easier-to-use API that also handles span grouping for you.

remorses:
Another use case is tests - quickly finding styled text without manual buffer iteration:

```ts
import { VTermStyleFlags } from "@opentui/core"

const { captureSpans, renderOnce } = await createTestRenderer({ width: 80, height: 24 })

// render components...
await renderOnce()

const { lines } = captureSpans()

// Find all bold text in the frame
const boldSpans = lines.flatMap(line => 
  line.spans.filter(span => span.flags & VTermStyleFlags.BOLD)
)

expect(boldSpans.map(s => s.text.trim())).toContain("Important")
```

kommander:
I see, that makes sense. Any way we could re-use the styled-text types for this, like for the flags, instead of introducing additional types? It seems like the spans are close to what styled text/styled chunks describe kind of.

remorses:
Refactored to use existing types: removed `VTermStyleFlags`, now uses `TextAttributes` directly and `RGBA` objects instead of hex strings. Added `RGBA.equals()` method. Net -19 lines.

packages/core/src/lib/RGBA.ts
```diff
@@ -64,6 +64,11 @@ export class RGBA {
   toString() {
     return `rgba(${this.r.toFixed(2)}, ${this.g.toFixed(2)}, ${this.b.toFixed(2)}, ${this.a.toFixed(2)})`
   }
+
+  equals(other?: RGBA): boolean {
```
kommander, line 68: Probably needs some epsilon for comparing floats moving forward, but fine for now.

PR #575, opened 2026-01-23
fix(core): skip stdout writes in testing mode for writeOut functions

remorses:
Without this change, the TUI was being rendered to the terminal when using the test renderer, polluting test output.

- Add early return in `writeOut()` and `writeOutMultiple()` when `self.testing` is true
- Use passed dimensions as fallback for `terminalWidth`/`terminalHeight` in non-TTY environments

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -496,8 +496,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     this.stdout = stdout
     this.realStdoutWrite = stdout.write
     this.lib = lib
-    this._terminalWidth = stdout.columns
-    this._terminalHeight = stdout.rows
+    this._terminalWidth = stdout.columns ?? width
+    this._terminalHeight = stdout.rows ?? height
     this.width = width
     this.height = height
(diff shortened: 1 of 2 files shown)


packages/core/src/renderer.ts
```diff
@@ -496,8 +496,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     this.stdout = stdout
     this.realStdoutWrite = stdout.write
     this.lib = lib
-    this._terminalWidth = stdout.columns
-    this._terminalHeight = stdout.rows
+    this._terminalWidth = stdout.columns ?? width
```
simonklee, line 499: Just checking, but does this handle correctly when stdout.columns is `0`? I don't write enough typescript/js to know how this is handled.

simonklee:
For what it's worth. Checked out the PR locally, and it solves the escape sequences in terminal problem 🚀

packages/core/src/renderer.ts
remorses: Yes, ?? only evaluates on null or undefined

PR #581, opened 2026-01-24
feat(zig): add --json and --bench flags to benchmark runner

simonklee:
Add two new CLI options for benchmark filtering:

- `--json`: Output results in machine-readable JSON format
- `--bench, -b NAME`: Filter individual benchmarks by name (case-insensitive)

This allows filtering on specific benchmark names (not just categories) and is
what enables me to measure individual benchmarks for opentui-bench.fly.dev as
well as their pprof profile.

### DIFF
--- a/packages/core/src/zig/bench.zig
+++ b/packages/core/src/zig/bench.zig
@@ -154,40 +154,56 @@ pub fn main() !void {
     if (!json_output and filter != null) {
         try stdout.print("Filtering benchmarks by: \"{s}\"\n", .{filter.?});
     }
+    if (!json_output and bench_filter != null) {
+        try stdout.print("Filtering individual benchmarks by: \"{s}\"\n", .{bench_filter.?});
+    }
 
     var ran_any = false;
 
     for (benchmarks) |bench| {
-        if (matchesFilter(bench.name, filter)) {
-            if (!json_output) {
-                try stdout.print("\n=== {s} Benchmarks ===\n\n", .{bench.name});
-                try stdout.flush();
-            }
+        if (!matchesFilter(bench.name, filter)) continue;
 
-            // Use arena for results only - benchmark modules manage their own temp memory
-            var results_arena = std.heap.ArenaAllocator.init(allocator);
-            defer results_arena.deinit();
+        // Use arena for results only - benchmark modules manage their own temp memory
+        var results_arena = std.heap.ArenaAllocator.init(allocator);
+        defer results_arena.deinit();
 
-            const start_time = std.time.nanoTimestamp();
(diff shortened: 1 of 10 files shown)


packages/core/src/zig/bench-utils.zig
```diff
@@ -393,3 +415,32 @@ pub fn printResults(writer: anytype, results: []const BenchResult) !void {
     try writer.writeAll("\x1b[0m\n");
     try writer.flush();
 }
+
+// JSON output for machine-readable results
+pub fn printResultsJson(writer: anytype, results: []const BenchResult, bench_name: []const u8) !void {
```
kommander: Maybe we can use the std lib json capabilities to write out json? The bench results are pretty defined, but this would need to be adapted and become brittle when we refactor/add results in the benchmarks. Here is an example of how to use the std lib:

```zig
const std = @import("std");

const Address = struct {
    street: []const u8,
    city: []const u8,
    zip: u32,
};

const Profile = struct {
    id: u32,
    name: []const u8,
    active: bool,
    tags: []const []const u8,
    address: Address,
    scores: []const u16,
    note: ?[]const u8,
};

pub fn main() !void {
    const tags = [_][]const u8{ "zig", "json", "example" };
    const scores = [_]u16{ 98, 87, 91 };

    const profile = Profile{
        .id = 42,
        .name = "Ada Lovelace",
        .active = true,
        .tags = tags[0..],
        .address = Address{
            .street = "123 Byte Lane",
            .city = "Turington",
            .zip = 40404,
        },
        .scores = scores[0..],
        .note = null,
    };

    var file = try std.fs.cwd().createFile("sample.json", .{ .truncate = true });
    defer file.close();

    var buffer: [4096]u8 = undefined;
    var writer = file.writer(buffer[0..]);
    try std.json.Stringify.value(profile, .{ .whitespace = .indent_2 }, &writer.interface);
    try writer.interface.writeAll("\n");
    try writer.interface.flush();
}
```

simonklee:
Looking into a slightly better approach for this. Realized that when running with --bench i'm still doing way too much so I want to filter earlier.

simonklee:
> Looking into a slightly better approach for this. Realized that when running with --bench i'm still doing way too much so I want to filter earlier.

I moved all bench filter logic to the top of each benchmark. That added quite a few indention changes so suppress whitespace is helpful: https://github.com/anomalyco/opentui/pull/581/changes?w=1

In the future I can imagine it would be helpful to have a function similar to Go's `t.Benchmark(name string, func(b) { ...})`.

kommander:
I think it's a little odd that benches themselves have to apply the filter, but as long this works for the tracking I am fine with this.

Issue #576, opened 2026-01-23
feat: Auto-focus on click (browser-like behavior)


eitanalka:
## Problem

Currently, clicking on a focusable element (input, scrollbox, etc.) doesn't automatically focus it. Users must manually handle `onMouseDown` to set focus state, which is different from browser behavior.

## Proposed Solution

Add browser-like auto-focus behavior to the renderer:
- Left-click triggers `focus()` on the clicked element
- If the element isn't focusable, bubble up to find the closest focusable ancestor
- Uses the same parent traversal pattern as `isWithinContainer`

## Benefits

- Matches expected browser behavior
- Simplifies component code (no manual `onMouseDown` handlers needed)
- Works seamlessly with `focusedBorderColor`/`focusedBackgroundColor` style props

## Implementation

I have a working implementation with tests ready:
- **Branch:** [feature-auto-focus-on-click](https://github.com/eitanalka/opentui/tree/feature-auto-focus-on-click)
- **Diff:** [View changes](https://github.com/eitanalka/opentui/compare/main...feature-auto-focus-on-click)

### Code changes:
1. `packages/core/src/renderer.ts` - ~10 lines to handle auto-focus with parent bubbling
2. `packages/core/src/tests/renderer.focus.test.ts` - 3 tests covering the behavior

Happy to open a PR if this approach looks good!

kommander:
makes sense and looks simple enough

eitanalka:
> makes sense and looks simple enough

Thank you. I opened a PR.

viralcodex:
hi @eitanalka, in your PR please add fixes: <issue> to link it to this issue automatically

PR #584, opened 2026-01-25
feat: Auto-focus on click (browser-like behavior)

eitanalka:
Fixes [Issue #576](https://github.com/anomalyco/opentui/issues/576)

## Problem

Currently, clicking on a focusable element (input, scrollbox, etc.) doesn't automatically focus it. Users must manually handle `onMouseDown` to set focus state, which is different from browser behavior.

## Proposed Solution

Add browser-like auto-focus behavior to the renderer:
- Left-click triggers `focus()` on the clicked element
- If the element isn't focusable, bubble up to find the closest focusable ancestor
- Uses the same parent traversal pattern as `isWithinContainer`

## Benefits

- Matches expected browser behavior
- Simplifies component code (no manual `onMouseDown` handlers needed)
- Works seamlessly with `focusedBorderColor`/`focusedBackgroundColor` style props

### DIFF
--- a/packages/core/src/renderables/Textarea.ts
+++ b/packages/core/src/renderables/Textarea.ts
@@ -121,16 +121,7 @@ const defaultTextareaKeybindings: KeyBinding[] = [
   { name: "right", super: true, shift: true, action: "select-visual-line-end" },
   { name: "up", super: true, shift: true, action: "select-buffer-home" },
   { name: "down", super: true, shift: true, action: "select-buffer-end" },
-
-  ...(process.platform === "darwin"
-    ? [
-        { name: "a", ctrl: true, action: "line-home" as const },
-        { name: "a", super: true, action: "select-all" as const },
-      ]
-    : [
-        { name: "a", ctrl: true, action: "select-all" as const },
-        { name: "a", super: true, action: "select-all" as const },
-      ]),
+  { name: "a", super: true, action: "select-all" },
 ]
 
 export interface SubmitEvent {}


packages/core/src/renderer.ts
```diff
@@ -1147,6 +1147,19 @@ export class CliRenderer extends EventEmitter implements RenderContext {
...
+        let current: Renderable | null = maybeRenderable ?? null
+        while (current) {
+          if (current.focusable) {
+            current.focus()
+            break
+          }
+          current = current.parent
+        }
+      }
+
```
kommander: This is a default behaviour that should be cancelable with `preventDefault` so I think this logic should happen after the event is triggered and potentially handled, then a renderable event handler or even a global event handler could use `event.preventDefault` and this logic can check `event.defaultPrevented`.

eitanalka: I updated the code. Please let me know if this approach works or if there is a better way to handle it.

eitanalka: I worked a little on refactoring the handleMouseData function to make it easier to handle defaults and prevent them. Please let me know what you think. https://github.com/eitanalka/opentui/compare/main...refactor/mouse-event-dispatch

kommander: I don't think `handleMouseData` is thoroughly test covered enough yet. I am even anxious about moving the `processMouseEvent` calls here. I would like to try and test cover all the edge cases it currently handles, because they are there for a reason. It might be some of the behaviour was introduced before there were proper tests. I would let a model dig through the history of that method and collect all the cases that were introduced and why to derive proper behavioural tests, get that merged first and from there start refactoring.

eitanalka: Okay I understand. How should we handle the auto-focus then?

kommander: I'd try to test cover the edge cases first. Then instead of moving around the `processMouseEvent` calls lets have a proxy method that checks if the event to be processed is a `down` event and default wasn't prevented. Let me suggest something...

kommander:
I added test coverage for the mouse data handler on main. Can you merge that in or let me push to your branch?

eitanalka:
> I added test coverage for the mouse data handler on main. Can you merge that in or let me push to your branch?

I updated the branch and gave permission to push to the branch just in case you need to modify it still.

PR #591, opened 2026-01-25
fix(core): handle special char markers in getSpanLines

remorses:
getSpanLines() now decodes special char markers (box-drawing, graphemes) correctly using getRealCharBytes().


packages/core/src/buffer.ts
kommander, line 156: We should probably document that this is only meant for testing/debugging/dev and not to implement features on top of, as it is quite inefficient.

PR #600, opened 2026-01-28
fix(input): set printable sequence for modifyOtherKeys chars

simonklee:
modifyOtherKeys (CSI 27;mod;code~) currently decodes name and
modifiers but leaves sequence as the raw ESC sequence. Textarea
inserts from sequence and rejects control chars (< 32), so sequences
like ESC [27;2;49~ (Shift+1 on AZERTY in WezTerm) never insert text.

Terminals that send plain "1" take the single-byte digit path and
work.

For printable codes, set sequence to the decoded character and mark
digits as number=true. This mirrors the single-byte digit behavior, so
text insertion now sees "1" instead of ESC and passes the guard.

Raw is preserved and special keys (return/escape/tab/space/backspace)
still keep their raw sequences, so control/navigation handling does
not change.

Keybinding lookup continues to use name+modifiers, so shortcuts are
unaffected.

This addresses an issue reported in opencode https://github.com/anomalyco/opencode/issues/10577

### DIFF
--- a/packages/core/src/lib/parse.keypress.test.ts
+++ b/packages/core/src/lib/parse.keypress.test.ts
@@ -1501,6 +1501,20 @@ test("parseKeypress - fallback to raw parsing when Kitty option is enabled but s
   expect(normalCtrl?.ctrl).toBe(true)
 })
 
+test("parseKeypress - modifyOtherKeys digits", () => {
+  const shiftOne = parseKeypress("\x1b[27;2;49~")!
+  expect(shiftOne.name).toBe("1")
+  expect(shiftOne.shift).toBe(true)
+  expect(shiftOne.ctrl).toBe(false)
+  expect(shiftOne.meta).toBe(false)
+  expect(shiftOne.option).toBe(false)
+  expect(shiftOne.number).toBe(true)
+  expect(shiftOne.sequence).toBe("1")
+  expect(shiftOne.raw).toBe("\x1b[27;2;49~")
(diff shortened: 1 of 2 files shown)


simonklee:
There are a couple of tests that regress due to this change. Looking into that before considering if the fix is sound.

simonklee:
> There are a couple of tests that regress due to this change. Looking into that before considering if the fix is sound.

Well, i have 4 failing tests on my local machine, but not in ci:

```

4 tests failed:
✗ InputRenderable > Key Bindings and Aliases > should support Emacs-style bindings by default [1.00ms]
✗ Textarea - Keybinding Tests > Selection with ctrl+shift+a/e (line home/end) > should not interfere with ctrl+a (without shift) [2.00ms]
✗ Textarea - Keybinding Tests > Visual line navigation with meta+a/e > should differ from ctrl+a/e when wrapping is enabled [2.00ms]
✗ Textarea - Editing Tests > Keyboard Input - Control Commands > should move to line start with Ctrl+A [2.00ms]

 3114 pass
 6 skip
 4 fail
 201 snapshots, 26820 expect() calls
Ran 3124 tests across 86 files. [63.73s]
error: script "test:js" exited with code 1
error: script "test" exited with code 1
```

simonklee:
> There are a couple of tests that regress due to this change. Looking into that before considering if the fix is sound.

Was unrelated to this PR: https://github.com/anomalyco/opentui/pull/601

PR #601, opened 2026-01-28
textarea: align ctrl+a defaults and tests with platform behavior

simonklee:
Remove the redundant ctrl+a line-home default entry (the platform block
already handles line-home on macOS and select-all elsewhere). Update
Input/Textarea ctrl+a assertions to branch on process.platform and keep
follow-up expectations stable by clearing selection.

This fixes 4 test failures on linux/windows where ctrl+a was expected
to move to line start but instead selected all text.

```
> bun run test

...

4 tests failed:
✗ InputRenderable > Key Bindings and Aliases > should support Emacs-style bindings by default [1.00ms]
✗ Textarea - Keybinding Tests > Selection with ctrl+shift+a/e (line home/end) > should not interfere with ctrl+a (without shift) [2.00ms]
✗ Textarea - Keybinding Tests > Visual line navigation with meta+a/e > should differ from ctrl+a/e when wrapping is enabled [1.00ms]
✗ Textarea - Editing Tests > Keyboard Input - Control Commands > should move to line start with Ctrl+A [2.00ms]

 3113 pass
 6 skip
 4 fail
 201 snapshots, 26814 expect() calls
Ran 3123 tests across 86 files. [64.20s]
error: script "test:js" exited with code 1
error: script "test" exited with code 1
```

### DIFF
--- a/packages/core/src/renderables/Textarea.ts
+++ b/packages/core/src/renderables/Textarea.ts
@@ -67,6 +67,7 @@ const defaultTextareaKeybindings: KeyBinding[] = [
   { name: "end", action: "buffer-end" },
   { name: "home", shift: true, action: "select-buffer-home" },
   { name: "end", shift: true, action: "select-buffer-end" },
+  { name: "a", ctrl: true, action: "line-home" },
   { name: "e", ctrl: true, action: "line-end" },
   { name: "a", ctrl: true, shift: true, action: "select-line-home" },
   { name: "e", ctrl: true, shift: true, action: "select-line-end" },
@@ -120,15 +121,7 @@ const defaultTextareaKeybindings: KeyBinding[] = [
   { name: "right", super: true, shift: true, action: "select-visual-line-end" },
   { name: "up", super: true, shift: true, action: "select-buffer-home" },
   { name: "down", super: true, shift: true, action: "select-buffer-end" },
-  ...(process.platform === "darwin"
-    ? [
-        { name: "a", ctrl: true, action: "line-home" as const },
-        { name: "a", super: true, action: "select-all" as const },
-      ]
-    : [
-        { name: "a", ctrl: true, action: "select-all" as const },
-        { name: "a", super: true, action: "select-all" as const },
-      ]),
+  { name: "a", super: true, action: "select-all" },
 ]
 
 export interface SubmitEvent {}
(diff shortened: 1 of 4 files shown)


remorses:
I think this was added because on macOS terminals like Zed Terminal cmd-left is sent as ctrl-a

simonklee:
> I think this was added because on macOS terminals like Zed Terminal cmd-left is sent as ctrl-a

I've changed the fix to branch on tests instead. Thanks for having a look at it.

packages/core/src/renderables/Textarea.ts
```diff
@@ -67,7 +67,6 @@ const defaultTextareaKeybindings: KeyBinding[] = [
   { name: "end", action: "buffer-end" },
   { name: "home", shift: true, action: "select-buffer-home" },
   { name: "end", shift: true, action: "select-buffer-end" },
-  { name: "a", ctrl: true, action: "line-home" },
```
kommander, line 70: I think users of opencode expect this very much. Tricky. It should not conflict with what opencode currently expects. Maybe we cannot provide a default binding for select-all and leave it none, or just super+a for now.

kommander:
There was a PR/branch to setup a test matrix for all platforms. Should revive that and run tests properly cross platform at some point.

packages/core/src/renderables/Textarea.ts
simonklee: That is duplicate, right.  The ternary adds that.

remorses:
This PR has support for that but currently only enables linux because some tests were failing

https://github.com/anomalyco/opentui/pull/440/files#diff-bc0066fcae057ae6403090cbd04189fdbfc929aa5c51e165175469de302701ef

packages/core/src/renderables/Textarea.ts
kommander: Yes, but I mean on other platforms than Darwin I think opencode users currently expect ctrl+a to also do line-home, because they expect readline/emacs.

simonklee: That's fair, it's what i did in the first commit but changed to the current impl after a comment from Tommy. Thinking about it more I lean towards the "no platform split" version.

simonklee: https://github.com/anomalyco/opentui/pull/601/changes/132288662a814290091c80657126757aa3856214

kommander: Yes no platform split seems more sane for the lib, platform splits can be done for keybindings on app level.

PR #604, opened 2026-01-31
Experimental solid babel transform insert order change fix

Adictya:


simonklee:
The current code manages child nodes by object identity, so this workaround is
no longer required. The proposed compiler change can also break dynamic child
insertion. Close this outdated draft.

PR #610, opened 2026-02-01
fix(core): use byte offset for line_starts in text wrap

### LINKED ISSUE #609: lineStarts returns column offset instead of byte offset when wrapping multi-byte text
### Problem

After #255 fix, CJK/emoji text renders correctly in most cases. However, `lineInfo.lineStarts` still returns **column offsets** instead of **byte offsets** when text wraps. This causes character corruption at line boundaries.

<details open>
<summary><b>Korean text wrap (representative case)</b></summary>

```ts
buffer.setStyledText(stringToStyledText("흐름도")) // 9 bytes
view.setWrapWidth(4) // wraps after "흐름"
```

|       | `lineStarts` | Line 0 | Line 1 |
| ----- | ---------- | ------ | ------ |
| **AS-IS** | `[0, 4]`     | `"흐�"`  | `"��도"` |
| **TO-BE** | `[0, 6]`     | `"흐름"` | `"도"`   |

</details>

<details>
<summary><b>Chinese text wrap</b></summary>

```ts
buffer.setStyledText(stringToStyledText("你好世界")) // 12 bytes
view.setWrapWidth(4)
```

|       | `lineStarts` | Line 0 | Line 1   |
| ----- | ---------- | ------ | -------- |
| **AS-IS** | `[0, 4]`     | `"你�"`  | `"��世界"` |
| **TO-BE** | `[0, 6]`     | `"你好"` | `"世界"`   |

</details>

<details>
<summary><b>Korean with leading ASCII</b></summary>

```ts
buffer.setStyledText(stringToStyledText(" 안녕a")) // 8 bytes
view.setWrapWidth(5)
```

|       | `lineStarts` | Line 0  | Line 1 |
| ----- | ---------- | ------- | ------ |
| **AS-IS** | `[0, 5]`     | `" 안�"`  | `"��a"`  |
| **TO-BE** | `[0, 7]`     | `" 안녕"` | `"a"`    |

</details>

<details>
<summary><b>Flag emoji</b></summary>

```ts
buffer.setStyledText(stringToStyledText("🇰🇷🇯🇵🇨🇳")) // 24 bytes
view.setWrapWidth(4)
```

|       | `lineStarts` | Line 0 | Line 1  |
| ----- | ---------- | ------ | ------- |
| **AS-IS** | `[0, 4]`     | `"🇰"`    | `"🇷🇯🇵🇨🇳"` |
| **TO-BE** | `[0, 16]`    | `"🇰🇷🇯🇵"` | `"🇨🇳"`    |

</details>

<details>
<summary><b>Skin tone emoji</b></summary>

```ts
buffer.setStyledText(stringToStyledText("👋🏻👋🏿hi")) // 18 bytes
view.setWrapWidth(4)
```

|       | `lineStarts` | Line 0 | Line 1 | Line 2 |
| ----- | ---------- | ------ | ------ | ------ |
| **AS-IS** | `[0, 4, 8]`  | `"👋"`   | `"🏻"`   | `"👋🏿hi"` |
| **TO-BE** | `[0, 8, 16]` | `"👋🏻"`   | `"👋🏿"`   | `"hi"`   |

AS-IS splits skin tone modifier from base emoji.

</details>

<details>
<summary><b>Mixed keycap + CJK</b></summary>

```ts
buffer.setStyledText(stringToStyledText("1️⃣한글")) // 13 bytes
view.setWrapWidth(4)
```

|       | `lineStarts` | Line 0 | Line 1   |
| ----- | ---------- | ------ | -------- |
| **AS-IS** | `[0, 3]`     | `"1�"`   | `"�⃣한글"` |
| **TO-BE** | `[0, 10]`    | `"1️⃣한"` | `"글"`     |

</details>

<details>
<summary><b>Long Korean (multiple wraps)</b></summary>

```ts
buffer.setStyledText(stringToStyledText("가나다라마바사")) // 21 bytes
view.setWrapWidth(6)
```

|       | `lineStarts` | Line 0 | Line 1 | Line 2   |
| ----- | ---------- | ------ | ------ | -------- |
| **AS-IS** | `[0, 6, 12]` | `"가나"` | `"다라"` | `"마바사"` |
| **TO-BE** | `[0, 9, 18]` | `"가나"` | `"다라"` | `"마바사"` |

Happens to display correctly but byte offsets are still wrong (6≠9, 12≠18).

</details>

### Root Cause

In `text-buffer-view.zig`, `cached_line_starts` was populated using `char_offset` (display column width) instead of tracking actual byte position through the wrap logic.

### Related

- #255

zenyr:
## Summary

- Fix `lineInfo.lineStarts` to return byte offsets instead of column offsets
- Prevents character corruption when slicing wrapped multi-byte text (CJK, emoji)

Fixes #609

## Changes

- Add `byte_offset` tracking to `VirtualLine` and wrap context in `text-buffer-view.zig`
- Update `cached_line_starts` to use byte position instead of char/column position
- Add regression tests for CJK text wrapping

## Test

```ts
buffer.setStyledText(stringToStyledText("흐름도"))
view.setWrapWidth(4)
view.lineInfo.lineStarts // [0, 6] ✅ (was [0, 4] ❌)
```

---

### Alternatives Considered

| Approach | Trade-off |
|----------|-----------|
| Compute byte offset on-demand at read time | O(n) per `lineInfo` access — defeats caching purpose |
| Store only in `cached_line_starts` without `VirtualLine` field | Would require extra params in `commitVirtualLine`, harder to maintain |
| Derive from chunk byte lengths | Still needs global traversal; chunks are buffer-relative, not document-relative |

**Chosen approach:** Track `byte_offset` during wrap (O(1) amortized). Memory overhead is ~4 bytes per wrapped line (~80KB for 10K lines), which is acceptable.

### Relation to Prior Work

- **#529** (merged): Introduced `VirtualLine` structure with `char_offset`. This PR builds on that by adding `byte_offset` alongside it.
- **#512** (closed): Addressed wrap *boundary detection* with `col_offset` in `WrapBreak`. Different problem — this PR fixes `lineStarts` *return values*, not boundary calculation.

Both prior PRs left `cached_line_starts` using character/column offsets, which corrupts multi-byte characters when used for slicing.

### Known Limitation

The no-wrap code path still uses `char_offset` for `lineStarts`. This is out of scope for this PR, as wrapping is the primary use case for `lineStarts`.

---

cc: @DonKongPaPa @kskang @soomtong @stefanahman — you showed interest in #255; this PR addresses a remaining edge case with line boundary handling. Feedback welcome!


zenyr:
Hi, it looks like the failing test `idle() resolves immediately when called on paused renderer` might be a flaky timing-based test.

Expected: < 50ms
Actual: 56.31ms

I ran this test locally multiple times and it passed consistently. The changes in this PR only affect `text-buffer-view.zig` (byte offset tracking), which should be unrelated to the renderer idle logic.

Would it be possible to re-run the CI when you have a moment? Please let me know if there's anything else I should address. Thank you!

zenyr:
**Update from Discord discussion:**

Addressed the "Known Limitation" mentioned in the PR body—no-wrap mode now uses `byte_offset` too.

Also optimized hot paths based on feedback about avoiding repeated measure calls:

| Commit | Change |
|--------|--------|
| `00336eb` | Fix no-wrap to use byte_offset (was still using char_offset) |
| `089f334` | Refactor to proactive tracking—no more retrospective `getBytes()` loop |
| `e344f44` | Add `byte_start`/`byte_len` to `VirtualChunk`—eliminates `findWrapPosByWidth` recalculation during word-wrap reflow |

The bigger refactor (merging `findGraphemeInfo`/`findWrapBreaks` into a single loop) is worth a separate issue.

<details>
<summary>Future optimization: unified grapheme/wrap-break loop</summary>

Currently `findGraphemeInfo` and `findWrapBreaks` in `utf8.zig` iterate over text separately:
- `findGraphemeInfo` → collects `byte_offset`, `byte_len`, `width`, `col_offset`
- `findWrapBreaks` → collects `byte_offset`, `char_offset` (no width info)

Both use similar SIMD-accelerated loops with ASCII fast paths. Merging them would:
1. Single pass instead of two
2. `WrapBreak` gets `byte_len`/`width` → word-wrap can use cached breaks directly
3. Eliminates grapheme lookups in hot path entirely

This could make `grapheme_start` in `VirtualChunk` redundant (replaced by `byte_start`), and remove the need for `charOffsetToColumn` conversions during wrap.

Estimated ~250 LOC across 4 files.
</details>

simonklee:
I think you still need to recompute the byte window to match the new column slice. Otherwise when the prefix ends mid‑chunk, byte_len still equals the full chunk length.

kommander:
I need to check if this also solves a word-split bug I found, will do that asap.

kommander:
The multiple `findWrapPosByWidth` calls in the hotpath are expensive. I was hping that the byte alignment would fix the "word split bug", which currently splits english words wrongly as well. The word wrap logic currently is a mess, it needs to be cleanup and simplified, doing `findGraphemes` and `findWrapBreaks` in one pass, allowing the wrap breaks to have byte and actual grapheme offsets. The main complexity that remains would be word breaks across fragmented chunks. Much of it is test covered.

I would use the tests in this branch here as a base to refactor. cc @simonklee

simonklee:
> I would use the tests in this branch here as a base to refactor. cc @simonklee

I'll try to get it work and use this as a test-case.

simonklee:
I don't see anything blow up: 

- https://opentui-bench.fly.dev/compare?base=97&curr=98&category=UTF-8+Operations
- https://opentui-bench.fly.dev/compare?base=97&curr=98&category=TextBuffer+Wrapping

kommander:
@simonklee I don't remember where we landed on this one?

simonklee:
> @simonklee I don't remember where we landed on this one?

I dont think we landed anywhere. I ran benchmarks, that's all. But looking at it - it is adjacent to my recent bug fixes so i can pick this up.

simonklee:
Thanks for working on this. I spent time testing and trying to optimize this implementation because the correctness goal is important.

The changes do improve correctness in several paths, but this approach is still not mergeable. I'm seeing material regressions in hot wrapping/layout paths, even after optimization attempts. In my run, measureForDimensions (layout streaming wrap) is still about 3x slower than baseline, and word wrap multi-line/static-wrap cases are also slower.

There is also a remaining correctness issue in wcwidth + word wrap with combining marks that can land inside a UTF-8 codepoint boundary.

Given this is core hot-path code, I need both correctness and performance. So I'm closing this PR.

This is not a no on the goal, but I think we need to explore a different approach that doesn't lead to regressions in hot paths

PR #614, opened 2026-02-01
fix: add .js extensions for NodeNext module resolution

### LINKED ISSUE #504: TypeScript NodeNext resolution fails with extensionless re-exports
### Summary
TypeScript NodeNext module resolution can’t resolve OpenTUI’s extensionless re-exports from its declaration files, which hides exported APIs like `createCliRenderer` and `createRoot`.

### Details
- @opentui/core `index.d.ts` contains `export * from "./renderer"`
- @opentui/react `src/index.d.ts` contains `export * from "./reconciler/renderer"`

With TS 5.9 and `moduleResolution: "NodeNext"`, `tsc --traceResolution` reports “Module name './renderer' was not resolved” even though `renderer.d.ts` exists. This results in TS2305 (“no exported member”).

### Repro
```json
// tsconfig.json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext"
  }
}
```

```ts
import { createCliRenderer } from '@opentui/core'
import { createRoot } from '@opentui/react'
```

### Workaround
Use `moduleResolution: "Bundler"` in the consumer tsconfig so TS accepts extensionless re-exports.

### Request
Consider adding explicit extensions in declaration re-exports (e.g. `./renderer.js` / `./reconciler/renderer.js`) or adjusting the package `exports` so NodeNext resolution works without the Bundler override.

---
_Opened on behalf of @schickling via Codex CLI._

schickling:
## Summary

This PR fixes TypeScript module resolution for consumers using `moduleResolution: "NodeNext"`.

- Add `.js` extensions to all relative imports/exports in TypeScript source files
- Convert directory imports like `./lib` to `./lib/index.js`
- Convert `from "."` imports to `from "./index.js"`

## Problem

When TypeScript processes the package with `moduleResolution: "NodeNext"`, it cannot resolve re-exports like `export * from "./renderer"` because NodeNext requires explicit file extensions. This causes TypeScript to report errors like:

```
Module '"@opentui/core"' has no exported member 'BoxRenderable'.
Module '"@opentui/react"' has no exported member 'createRoot'.
```

## Solution

Add explicit `.js` extensions to all relative imports/exports. TypeScript understands that `./renderer.js` refers to `./renderer.ts` during compilation, and the generated `.d.ts` files will have the correct extensions for NodeNext resolution.

## Test plan

- [x] Verified the fix with the reproduction at https://github.com/schickling-repros/opentui-nodenext-issue
- [x] TypeScript trace resolution shows all relative imports are successfully resolved
- [x] No new TypeScript errors introduced (existing errors in examples/tests remain)

Fixes #504

---
Generated with [Claude Code](https://claude.com/claude-code)


msmps:
@schickling mind running a quick `prettier:write` from the root?

schickling:
@msmps Thanks for the review! I've resolved the merge conflict and run `prettier:write` as requested.

**Resolved:**
- Merged latest `main` branch changes
- Fixed conflict in `Markdown.test.ts` by combining both the `.js` extensions (for NodeNext compatibility) and the new `TextRenderable` import from upstream
- Applied prettier formatting

Ready for another look!

simonklee:
@msmps Is this generally a good approach - i don't write ts enough to know. I guess you could also rewrite the extensions during compile for the `.d.ts` files since js is only required here.

msmps:
> @msmps Is this generally a good approach - i don't write ts enough to know. I guess you could also rewrite the extensions during compile for the `.d.ts` files since js is only required here.

it's a good and uniform approach! i think we need to introduce a linter though like eslint, biome or oxlint to enforce the js extension going forward so this doesn't become a regression! i have a branch locally for that but would like your thoughts on this @simonklee @kommander

simonklee:
oxlint (and oxfmt) seem to work well in the projects i've tried them in.

kommander:
We should add a linter with that rule. Not a big fan of linters otherwise. Maybe just that one rule, otherwise it will diverge for sure.

inakineitor:
> We should add a linter with that rule. Not a big fan of linters otherwise. Maybe just that one rule, otherwise it will diverge for sure.

If you want to avoid a linter maybe we can switch the tsconfig to use NodeNext module resolution and the LSP should pick up if the import is done wrong without the .js extension (because it won't recognize the import).

remorses:
Agree, switching to NodeNext for module and resolver would be best way.

inakineitor:
Is there anything I could do to help move this PR along? It would be great to have OpenTUI be on NodeNext!

kommander:
We'll go forward with this asap.

schickling:
excited-lets-go-gif

inakineitor:
@kommander I just realized that the `moduleResolution` was not changed in the tsconfig as part of this pull request. Would you like me to open a new one to do it? This means that there are some uncaught [imports](https://github.com/anomalyco/opentui/blob/24cc57e7c7023e8be2809fee9bd76149aa8d41a7/packages/core/src/NativeSpanFeed.ts) without the `.js` extension.

simonklee:
> @kommander I just realized that the `moduleResolution` was not changed in the tsconfig as part of this pull request. Would you like me to open a new one to do it? This means that there are some uncaught [imports](https://github.com/anomalyco/opentui/blob/24cc57e7c7023e8be2809fee9bd76149aa8d41a7/packages/core/src/NativeSpanFeed.ts) without the `.js` extension.

If you want to open one that would be awesome

inakineitor:
> > @kommander I just realized that the `moduleResolution` was not changed in the tsconfig as part of this pull request. Would you like me to open a new one to do it? This means that there are some uncaught [imports](https://github.com/anomalyco/opentui/blob/24cc57e7c7023e8be2809fee9bd76149aa8d41a7/packages/core/src/NativeSpanFeed.ts) without the `.js` extension.
> 
> If you want to open one that would be awesome

I have draft PR for this ready [here](https://github.com/anomalyco/opentui/pull/843) but I realized that there are two upstreams by anomaly that are exporting NodeNext incompatible types, so I wanted to fix those first. Would you mind looking at https://github.com/anomalyco/bun-ffi-structs/pull/3 and telling me if this works for your setup? If that works I can do the same for https://github.com/kommander/bun-webgpu.

PR #617, opened 2026-02-02
feat(DiffRenderable): expose line highlighting API

3000-2:
## Summary

Add public methods to `DiffRenderable` for controlling line colors from external code:

## Motivation

These methods enable use cases like highlighting the "current hunk" in a diff viewer application. The internal `LineNumberRenderable` already supports line coloring - this PR simply exposes that capability through the public `DiffRenderable` API.

## Example Usage

```typescript
// Highlight lines 10-20 with a custom color
diffRenderable.highlightLines(10, 20, "#3a3a5c")

// Clear the highlight
diffRenderable.clearHighlightLines(10, 20)
```

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -34,7 +34,7 @@
         "@opentui/core-linux-x64": "0.1.79",
         "@opentui/core-win32-arm64": "0.1.79",
         "@opentui/core-win32-x64": "0.1.79",
-        "bun-webgpu": "0.1.4",
+        "bun-webgpu": "0.1.5",
         "planck": "^1.4.2",
         "three": "0.177.0",
       },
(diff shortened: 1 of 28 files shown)


kommander:
That would make sense, yes. The `setLineColor` method on `LineNumberRenderable` calls `gutter.setLineColors`, which requests a re-render. Maybe the `highlightLines` method could be implemented on the `LineNumberRenderable` directly to streamline that and only update the gutter once for all the line colors being set.

3000-2:
@kommander Thanks for the suggestion! I've moved highlightLines and clearHighlightLines to LineNumberRenderable so it batches all line color changes and only calls gutter.setLineColors() once at the end.

PR #619, opened 2026-02-02
Add marginX/marginY and paddingX/paddingY to the style attributes

### LINKED ISSUE #618: Small Suggestion: have paddingX/paddingY and marginX/marginY
Hi Team,
I am using OpenTUI for one of my projects and while writing padding and margin values for components (solid js), I felt I have been writing left and right a lot.

Since I have used tailwind for my projects, I felt like X and Y attributes for both margin and padding should be a simple addition to make writing these values a little easier instead of left right and top and down separately if the values used are same.

I have already linked the PR to this issue if you guys find it acceptable to merge it.

Would be more than happy to discuss more on this.

Thanks for building a great library to use, it's really cool 😄

viralcodex:
Fixes: #618 

Hi team,

Here's a small change that I have added.
X and Y attributes for both padding and margin just to make writing these values a little easier.

This is my first PR on this project and I am using this project for my own CLI tool I am building.

We can discuss and see if it fits with the project requirements.

But from a Dev experience point of view, I felt like I should add it to make it better, Also this aligns with people accustomed to tailwindCSS as well.

Here's a small demo:

https://github.com/user-attachments/assets/42d04e53-2006-435e-8e59-ef7d641ffe2f


kommander:
Hi, I think that's fine to add as convenience and it's only additive to the API, so shouldn't be an issue. Now that we have proper docs for the website, maybe we could directly document that there as well?

viralcodex:
Hi @kommander, sure I can add it there as well, not an issue

viralcodex:
@kommander, added in the docs, please let me know if I have missed something or something else is needed/changed.

Thanks 😄

simonklee (approved):
lgtm

PR #621, opened 2026-02-03
feat(diff): add SlottableDiff

ishaksebsib:
## Summary
Adds a new `SlottableDiffRenderable` component that renders each diff line as a separate interactive element, enabling slots for inline content like comment boxes.

## Why
The existing `DiffRenderable` component renders the entire diff as a monolithic `CodeRenderable`, which provides excellent performance for viewing but doesn't allow for interactivity. Each line is just part of a large text buffer with no individual identity or click handling.

## Changes
- **DiffLine.ts**: New `DiffLineRenderable` component for individual diff lines with click support, line numbers, and syntax highlighting
- **SlottableDiff.ts**: New `SlottableDiffRenderable` that extends `DiffRenderable` with `insertSlot()`/`removeSlot()`/`onLineClick` methods for injecting arbitrary content between lines
- **Diff.ts**: Made private properties/methods `protected` to enable inheritance by SlottableDiff
- **Demo**: Added `interactive-diff-demo.tsx` examples showing click-to-comment functionality *interactive using the portal pattern

## Features
- Click any diff line to trigger `onLineClick` callback
- Insert custom UI elements (input boxes, comments) between lines via slots
- Works with both unified and split views

## Performance
- **Virtualization**: When `virtualize` is enabled, only the visible range is rendered with configurable `overscan` padding.
- **Surgical slot updates**  No full rebuilds on slot changes Instead of rebuilding all lines when adding/removing slots (O(n) operation), the component tracks visual rows via `_visualRows` array and uses `insertBefore()` to place slots at the exact position (O(1) operation)
- **Slot management**: `insertSlot()` checks if the slot already exists and only performs manipulation when necessary
- **Efficient slot removal**: Removing a slot doesn;t triggers a full view rebuild - only the slot is removed and render is requested

## Example Use Case
Interactive code review where users can click on diff lines to add inline comments, similar to GitHub's PR review interface.


kommander:
I have some performance concerns with having each line be its own stacked Renderable tree. It's a lot of boxes and texts, which for a large diff and especially a split diff can become super heavy on the layout, which already has some performance penalties.

ishaksebsib:
Yeah, agreed. I added virtualization for this exact reason, we only parse and render the visible lines/slots, so performance stays stable even with large diffs.

PR #624, opened 2026-02-04
fix(core): enable threading on Linux

simonklee:
I stumbled over this, but don't see any reason why threading should be disabled on Linux.

Probably wait on https://github.com/anomalyco/opentui/pull/623 to be merged (mostly about testing)


simonklee:
> (fail) TestRecorder > should capture timestamps in increasing order [37.45ms]

Seems like a concurrency issue. WIll look into.

simonklee:
> > (fail) TestRecorder > should capture timestamps in increasing order [37.45ms]
> 
> Seems like a concurrency issue. WIll look into.

Nvm, think its a timing issue.

PR #626, opened 2026-02-04
ci(core): run tests on both macOS and Linux

simonklee:
Add matrix strategy for build core to run tests on ubuntu-latest in addition to macos-latest. Since it's a GH Actions change i have no idea if it works or fails a random way 😉

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
-*                   @kommander @msmps @Hona @simonklee
-/packages/react/    @msmps @Adictya @fezproof @kommander @Hona @simonklee
-/packages/solid/    @Adictya @fezproof @msmps @kommander @Hona @simonklee
+*                   @kommander @msmps @Hona @simonklee @kitlangton
+/packages/react/    @msmps @Adictya @fezproof @kommander @Hona @simonklee @kitlangton
+/packages/solid/    @Adictya @fezproof @msmps @kommander @Hona @simonklee @kitlangton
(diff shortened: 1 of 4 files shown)


simonklee:
> Core - Build and TestExpected — Waiting for status to be reported

... guess i need to change the gate or add a new one.

kommander:
I was wondering if we could re-use a similar setup as we have for the release CI flow, building the native libs once, then load the artifacts in a matrix, because for the release we use a cross compiled version, then the typescript tests would run with the cross compiled version on different platforms as well. 

The native tests can be run built and run per platform. Cross compiling once could slightly improve the run duration as well I think. 

But mainly to actually test the cross compiled binaries, to be close to what we would release. WDYT?

simonklee:
> I was wondering if we could re-use a similar setup as we have for the release CI flow, building the native libs once, then load the artifacts in a matrix, because for the release we use a cross compiled version, then the typescript tests would run with the cross compiled version on different platforms as well.
> 
> The native tests can be run built and run per platform. Cross compiling once could slightly improve the run duration as well I think.
> 
> But mainly to actually test the cross compiled binaries, to be close to what we would release. WDYT?

I looked more at the current pipelines and there are a couple ways you could go, but for testing core native i think you need compile on that test binary anyway so cross compiling doesn't help here.

I ended up with this

```
build (macos-latest)
├── Cross-compile all platforms
└── Upload artifacts
        │
        ▼
test (matrix)
├── macos-latest (darwin-arm64)
│   ├── Download artifacts
│   ├── Run native tests
│   └── Run TS tests
│
└── ubuntu-latest (linux-x64)
    ├── Download artifacts
    ├── Run native tests
    └── Run TS tests
        │
        ▼
build-complete (gate)
```

simonklee:
I think maybe what we can do to optimize slightly is to also test native on the macos machine that does cross compilation so that that you avoid booting another macos machine for test native.

PR #628, opened 2026-02-04
fix(core): correct colors in Apple Terminal and add 256-color fallback 

Harry-zklcdc:
## Summary
- Disable truecolor and explicit width probing for Apple Terminal to avoid incorrect color output.
- Add 256-color fallback rendering when truecolor is not supported.
- Route renderer color output through capability-aware selection (24-bit vs 256).

### Before

image
image

### After

image
image

### DIFF
--- a/packages/core/src/zig/renderer.zig
+++ b/packages/core/src/zig/renderer.zig
@@ -33,6 +33,20 @@ fn rgbaComponentToU8(component: f32) u8 {
     return @intFromFloat(@round(clamped * 255.0));
 }
 
+fn rgbToAnsi256(r: u8, g: u8, b: u8) u8 {
+    if (r == g and g == b) {
+        if (r < 8) return 16;
+        if (r > 248) return 231;
+        return @intCast(232 + ((@as(u16, r) - 8) / 10));
+    }
+
+    const r6: u16 = (@as(u16, r) * 5 + 127) / 255;
+    const g6: u16 = (@as(u16, g) * 5 + 127) / 255;
+    const b6: u16 = (@as(u16, b) * 5 + 127) / 255;
+
+    return @intCast(16 + (36 * r6) + (6 * g6) + b6);
+}
+
 pub const DebugOverlayCorner = enum {
     topLeft,
     topRight,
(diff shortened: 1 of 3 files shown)


packages/core/src/zig/renderer.zig
```diff
@@ -33,6 +33,20 @@ fn rgbaComponentToU8(component: f32) u8 {
     return @intFromFloat(@round(clamped * 255.0));
 }
 
+fn rgbToAnsi256(r: u8, g: u8, b: u8) u8 {
+    if (r == g and g == b) {
+        if (r < 8) return 16;
+        if (r > 248) return 231;
+        return @intCast(232 + ((@as(u16, r) - 8) / 10));
```
simonklee, line 40: If `r` is 248 this overflows:

```
232 + (248 - 8) / 10 = 232 + 24 = 256
```

and you can't fit 256 in an u8

simonklee: A minimal fix is `if (r >= 248) return 231;`

Harry-zklcdc: I fixed it at https://github.com/anomalyco/opentui/pull/628/commits/79eb5f3f55acd88c57fd57cac8377a54fe67205b

kommander:
See discussion in #559

Harry-zklcdc:
> See discussion in #559

I would like to first merge this fix to address the issue where the downstream product OpenCode displays incorrectly in the macOS 15 terminal.

kommander:
The problem I have is rgbToAnsi in the hot path for all cells all the time. The RGBA+A solution discussed in the other PR would avoid a lot of recalculations, as it basically caches the index in the RGBA data and allows further improvements, avoiding many recalculations for each frame.

Harry-zklcdc:
> The problem I have is rgbToAnsi in the hot path for all cells all the time. The RGBA+A solution discussed in the other PR would avoid a lot of recalculations, as it basically caches the index in the RGBA data and allows further improvements, avoiding many recalculations for each frame.

This is indeed an issue where performance optimization is possible, and I look forward to another PR being merged as soon as possible to fix the display issues in existing macOS terminals.

PR #631, opened 2026-02-05
WIP: core: switch to yoga ffi bindings

simonklee:
WIP pr so i can test this e2e with opencode.


remorses:
Another approach to make testing easier in opencode is to make dependencies external so the yoga dependency is not bundled into opentui and you can use package.json overrides

Example PR doing this: https://github.com/anomalyco/opentui/pull/416

It has conflicts I can fix them if there is will to merge it

simonklee:
> Another approach to make testing easier in opencode is to make dependencies external so the yoga dependency is not bundled into opentui and you can use package.json overrides
> 
> Example PR doing this: #416
> 
> It has conflicts I can fix them if there is will to merge it

Thanks, let me have a look at that later — this was more just to get the bundled package. ~I realize now i need to recreate this pr on https://github.com/anomalyco/opentui, because i created it on my own handle.~ nvm. looked the wrong place.

PR #632, opened 2026-02-05
fix(keyboard): correct kitty numpad key code mappings

eyenalxai:
fix off-by-one error in kitty keyboard protocol numpad key mappings.

according to [kitty documentation](https://sw.kovidgoyal.net/kitty/keyboard-protocol/), the numpad key codes start at 57399 for kp0.

the previous mapping started at 57400, causing all keys to be off by one.
this meant pressing numpad 2 (code 57401) would incorrectly report as kp1.

changes:
shifted all keypad codes down by 1 to match kitty spec.
updated tests with correct key codes.

`bun test packages/core/src/lib/parse.keypress-kitty.test.ts` passes

### DIFF
--- a/packages/core/src/tests/renderer.input.test.ts
+++ b/packages/core/src/tests/renderer.input.test.ts
@@ -994,10 +994,10 @@ test("Kitty keyboard emoji via keyInput events", async () => {
 })
 
 test("Kitty keyboard keypad keys via keyInput events", async () => {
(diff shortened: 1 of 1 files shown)


simonklee (approved):
Lgtm

eyenalxai:
> 1 tests failed:
> (fail) Kitty keyboard keypad keys via keyInput events [21.85ms]

oops missed this one brb

Issue #92, opened 2025-08-27
Support for rendering images (Kitty Graphics, SIXEL, iterm2 image)


remorses:
SIXEL allows you to render images in the terminal, this would be great for a project I am working on, [Termcast](https://github.com/remorses/termcast): reimplementing Raycast in the terminal, to show extensions icons.

Most popular terminals already support SIXEL:

| Terminal                          | SIXEL?          | Notes                                                                                                       |
| --------------------------------- | --------------- | ----------------------------------------------------------------------------------------------------------- |
| **xterm**                         | **Yes**         | Native SIXEL + XTSMGRAPHICS to query max geometry.                                                          |
| **WezTerm**                       | **Yes**         | Also supports iTerm2/kitty image protocols; default max ~25M px/image via `imgcat`.                         |
| **iTerm2 (macOS)**                | **Yes**         | Supported since 3.3; includes a SIXEL feature test (e.g., print `snake.six`). Also supports IIP.           |
| **GNOME Terminal (VTE)**          | **No** (varies) | Upstream VTE has work/flags; many distro builds ship **without** SIXEL. Some Flatpaks/bundles may enable it.|
| **Konsole (KDE)**                 | **Yes**         | Since KDE Gear **22.04**.                                                                                   |
| **Windows Terminal**              | **Yes**         | Added in Preview **1.22**, mainline in **1.23**.                                                            |
| **VS Code (integrated terminal)** | **Yes**         | Uses xterm.js addon; supports SIXEL & IIP. Enabled since v**1.80** (toggle: `terminal.integrated.enableImages`). |
| **kitty**                         | **No**          | Won’t implement SIXEL; uses its own graphics protocol instead.                                              |
| **Alacritty**                     | **No**          | No built-in SIXEL (there are third-party/fork experiments).                                                 |
| **foot** (Wayland)                | **Yes**         | Supported (e.g., since 1.2.0).                                                                              |

Here is a website with more in depth support for SIXEL: https://www.arewesixelyet.com/

Also Alacritty has an active PR to add support for it [here](https://github.com/alacritty/alacritty/pull/4763)

Ghostty still [does not support SIXEL](https://github.com/ghostty-org/ghostty/discussions/2496#discussioncomment-11353475) but it supports [Kitty graphics protocol](https://sw.kovidgoyal.net/kitty/graphics-protocol/)

kevcube:
[...]

PR #633, opened 2026-02-05
Add ImageRenderable for pixel based images

iamlemec:
This adds an `ImageRenderable` element to render pixel images. The rendering closely mirrors the current/next buffer approach from `OptimizedBuffer` but using `PixelBuffer`, which is basically a collection of `PixelPatch` objects. This allows for flicker-free, non-reduntant rendering.

The main interface is a new `renderPixels` function on `Renderables` that allows any renderable (not just `ImageRenderable`) to display arbitrary numbers of images. Images are hashed to prevent redundant writes. The actual output is buffered with regular text output on the native side to prevent flicker.

I tried to keep the changes as minimal as possible. I did need to increase the size of the output buffer from 2MB to 8MB to accomodate images. The only additional dependency is `pngjs` for decoding PNG files. There is an example with animated images called `image.ts`. This currently only outputs kitty format, but other formats should not be difficult. The data input is just a `Uint8Array` of RGBA pixels, but could push file loading from example to core library.

In terms of z-index issues, so far as I can tell, images are always placed above text, even if the text is written afterwards (at least in kitty/ghostty). So doing truly interleaved text and pixels would be non-trivial, but not impossible. You'd basically have to go in and transparent blit out parts of the images if they are occluded. This doesn't do that, so the pixel images are always just an overlay on top.

See issue #92 for more discussion.


simonklee:
Image support added in main

PR #636, opened 2026-02-06
feat(ssh): add @opentui/ssh package with stream-mode renderer migration

msmps:
# `@opentui/ssh` + stream-mode renderer migration

## Summary
- Adds a new `@opentui/ssh` package for running OpenTUI apps over SSH with per-session renderer isolation.
- Migrates core non-stdout transport from callback output to `outputMode: "stream"` backed by `NativeSpanFeed` in Zig.
- Applies follow-up hardening from review: auth decision guarding, session admission race fixes, authorized_keys option safety, PTY dimension normalization, and renderer stdin teardown correction.

### 1) New package: `@opentui/ssh`
- `createSSHServer()` with lifecycle events (`listening`, `session`, `error`, `close`).
- `SSHSession` wrapper with stream output wiring, input forwarding, resize handling, and idempotent close.
- Built-in middleware:
  - `logging()`
  - `publicKey()` (OpenSSH `authorized_keys` parsing/matching)
  - `allowAll()` (production-ready open server, no warnings)
  - `devMode()` (development-only, prints security warning)
- Host key utilities and docs/examples/tests.

### 2) Core renderer stream architecture
- `outputMode: "stream"` in `@opentui/core` with required `onOutput`, `width`, and `height` validation.
- NativeSpanFeed/Zig integration for feed-backed output dispatch and backpressure-safe async delivery.
- Renderer and test utilities updated for stream-mode creation and validation.

### 3) Reviewer follow-up fixes
- Prevents auth double-decision races via guarded accept/reject flow in SSH auth handling.
- Enforces `maxSessions` against active + pending session creation to close async admission race.
- Normalizes invalid PTY dimensions (e.g. `0x0`) to safe defaults before renderer creation.
- Treats `authorized_keys` option-prefixed entries as unsupported and ignores them (with warning) to avoid silently weakening restrictions.
- Fixes global `process.stdin.pause()` behavior so stream-only renderer teardown does not pause stdin.

## Documentation and Testing
- Expanded `packages/ssh/README.md` with middleware phases/order, auth/session patterns, and compatibility notes.
- Added/expanded tests across:
  - `packages/ssh/tests/*` (server, middleware, authorized-keys, session, integration, host-key)
  - `packages/core/src/tests/renderer.stream-mode.test.ts`
  - `packages/core/src/tests/renderer.stream-validation.test.ts`

## Manual Verification (SSH Counter Example)
1. Start the example server:
   - `cd packages/ssh`
   - `bun run example`
2. In another terminal, connect with SSH:
   - `ssh -p 2222 localhost`
3. Validate behavior in one or more sessions:
   - `+` / `-` updates the counter
   - terminal resize updates displayed dimensions
   - `q` cleanly closes the session
4. (If host key changed) clear stale host key entry:
   - `ssh-keygen -R "[localhost]:2222"`

### 1) Install Pilotty
```bash
npm install -g pilotty
```
Verify with `pilotty --version`.

### 2) Install the Pilotty OpenCode skill
```bash
npx skills add msmps/pilotty
```

### 3) AI-assisted review prompt
Use this in OpenCode to run the flow autonomously:

```text
use pilotty to test out the ssh impl with the ssh counter example;
spawn concurrent sessions and validate
```

Expected validation flow: start server, open concurrent SSH sessions, verify per-session counter isolation, test resize propagation, and quit sessions cleanly.

## Notes
- `publicKey()` currently supports key matching only; OpenSSH option prefixes (`from=`, `command=`, `no-pty`, etc.) are intentionally not enforced yet and are ignored.

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -1493,9 +1493,9 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     ) {
       const canStartSelection = Boolean(
         maybeRenderable &&
-        maybeRenderable.selectable &&
-        !maybeRenderable.isDestroyed &&
-        maybeRenderable.shouldStartSelection(mouseEvent.x, mouseEvent.y),
+          maybeRenderable.selectable &&
+          !maybeRenderable.isDestroyed &&
+          maybeRenderable.shouldStartSelection(mouseEvent.x, mouseEvent.y),
       )
 
       if (canStartSelection && maybeRenderable) {
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -161,6 +161,9 @@ function createNullStdin(): NodeJS.ReadStream {
 /**
  * Creates a no-op WriteStream used for stream mode when stdout is omitted.
  * Carries terminal dimensions while all actual output flows through onOutput.
+ * Sets isTTY=true so existing renderer internals work without null checks.
+ * Debug features like dumpOutputCache() write to this no-op stream and are
+ * intentionally discarded in stream mode.
  */
 function createNullStdout(width: number, height: number): NodeJS.WriteStream {
   const stdout = new Writable({
@@ -337,14 +340,20 @@ export enum MouseButton {
 
 const rendererTracker = singleton("RendererTracker", () => {
   const renderers = new Set<CliRenderer>()
+  let processStdinUsers = 0
   return {
     addRenderer: (renderer: CliRenderer) => {
       renderers.add(renderer)
+      if (renderer.stdin === process.stdin) processStdinUsers++
     },
     removeRenderer: (renderer: CliRenderer) => {
       renderers.delete(renderer)
+      if (renderer.stdin === process.stdin) processStdinUsers = Math.max(0, processStdinUsers - 1)
       if (renderers.size === 0) {
-        process.stdin.pause()
+        // Only pause process.stdin if a renderer was actually using it
+        if (processStdinUsers === 0) {
+          process.stdin.pause()
+        }
 
         if (hasSingleton("tree-sitter-client")) {
           getTreeSitterClient().destroy()
@@ -367,6 +376,16 @@ export async function createCliRenderer(config: CliRendererConfig = {}): Promise
   let feed: NativeSpanFeed | null
 
   if (config.outputMode === "stream") {
+    if (typeof config.onOutput !== "function") {
+      throw new Error('outputMode "stream" requires onOutput')
+    }
+    if (!Number.isInteger(config.width) || config.width < 1) {
+      throw new Error('outputMode "stream" requires width to be a positive integer')
+    }
+    if (!Number.isInteger(config.height) || config.height < 1) {
+      throw new Error('outputMode "stream" requires height to be a positive integer')
+    }
+
     // Stream mode has no process-backed tty, so width/height come from config.
     // Null-object streams let existing renderer internals run without null checks.
     width = config.width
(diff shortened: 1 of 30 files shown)


msmps:
cc @kommander @kitlangton

packages/core/src/renderer.ts
```diff
@@ -278,20 +368,51 @@ export async function createCliRenderer(config: CliRendererConfig = {}): Promise
...
+    stdin = config.stdin ?? createNullStdin()
+    stdout = config.stdout ?? createNullStdout(width, height)
+    feed = NativeSpanFeed.create(config.feedOptions ?? {})
+  } else {
+    stdin = config.stdin || process.stdin
+    stdout = config.stdout || process.stdout
+    width = stdout.columns || 80
+    height = stdout.rows || 24
+    feed = null
+  }
```
kommander, line 402: I can see where the outputMode is coming from. However, I would try to keep the interface simpler, `stdin` and `stdout` are already options, so these `streams` can be overridden. The `CliRenderer` can "just" check if the given streams are `process` streams and if not setup the `NativeSpanFeed` and pipe all data to the given `stdout` stream. Dimensions `width/height` are also already options for the renderer and either stdout is a tty stream and we can get that info or use the options available (with fallback defaults). 

TL;DR: Opentui should be dumb about the streams, it takes stdin/stdout and reads what it gets and pipes out what it has.

packages/core/src/renderer.ts
```diff
@@ -1720,30 +1949,33 @@ export class CliRenderer extends EventEmitter implements RenderContext {
...
+      this.stdin.pause()
     }
-
-    this.stdin.pause()
   }
 
   public resume(): void {
-    if (this.stdin.setRawMode) {
-      this.stdin.setRawMode(true)
+    // Skip stdin setup in stream mode — input comes via renderer.input()
```
kommander, line 1965: If stdin is just overridden it should just still do:
```ts
if (this.stdin.setRawMode) {
    this.stdin.setRawMode(true)
}
```
And use methods as available. Again just being "dumb" about streams at take them at face value mostly.

packages/core/src/renderer.ts
```diff
@@ -1876,10 +2111,40 @@ export class CliRenderer extends EventEmitter implements RenderContext {
...
+
+    // Tear down stream feed callbacks before destroying native renderer.
+    //
+    // Memory ownership: the renderer's native struct (buffers, hit grids) is
+    // freed by destroyRenderer(). The feed's chunk memory is owned independently
+    // by NativeSpanFeed and freed in its own finalizeDestroy(), which defers
+    // until all pending async onOutput handlers have settled. This means
+    // destroyRenderer() does not free the memory that onOutput views
+    // reference — those views point into feed-owned chunks, not renderer memory.
+    if (this._outputMode === "stream" && this._feed) {
```
kommander, line 2130: These branches should not be necessary, streams are just streams that are process.stdin/stdout or not.

packages/core/src/zig/renderer.zig
```diff
@@ -212,6 +274,19 @@ pub const CliRenderer = struct {
...
+        else
+            try allocator.alloc(u8, 0);
+        errdefer allocator.free(instanceOutputA);
+        const instanceOutputB = if (uses_instance_buffers)
+            try allocator.alloc(u8, OUTPUT_BUFFER_SIZE)
+        else
+            try allocator.alloc(u8, 0);
+        errdefer allocator.free(instanceOutputB);
+
         self.* = .{
```
simonklee, line 290: `renderer.zig` is currently mixing frame generation with output transport concerns. Can we extract an `OutputBackend` and make it the single owner of strategy-specific state/behavior?

`OutputBackend` should own:

- stdout double buffers (`instanceOutputA/B`, active buffer, lengths)
- span feed handle + commit/backpressure rules
- thread support/handoff details

Then `CliRenderer` can call a small backend API and keep one render path.

This removes repeated `switch (self.outputStrategy)` branches and gives one place to enforce output invariants.

msmps:
splitting this out and addressing implementation in a separate pull request

PR #637, opened 2026-02-07
fix(solid,core): support Windows path separators and improve native package diagnostics

RocketBoyMusic:
## Summary

I was running into issues in a separate repo while using OpenTUI on Windows, and wanted to fix this here so other users don’t hit the same problems.

This PR makes two focused improvements:
1. Fixes `@opentui/solid` Bun plugin path matching so Solid server runtime rewrites work with both Unix (`/`) and Windows (`\\`) separators.
2. Improves `@opentui/core` diagnostics when platform-native packages are missing, and adds a Windows CI smoke test to catch this earlier.

## Changes

- Updated Solid plugin `onLoad` server-runtime filters to be path-separator agnostic (`[\\/]`).
- Kept rewrite behavior unchanged (`server.js -> solid.js` and `server.js -> store.js`).
- Added regression tests for Unix and Windows paths, plus a negative non-server case.
- Added clearer error messaging in `packages/core/src/zig.ts` when native package import fails, including recovery commands.
- Added a `windows-smoke` job in `build-core.yml` that runs `bun -e "await import('@opentui/core')"`.

## Testing

- `bun test packages/solid/tests/solid-plugin.test.ts`
  - `3 pass, 0 fail`


packages/solid/tests/solid-plugin.test.ts
Adictya, line 1: Feels a bit extra to test this

Adictya (dismissed):
Solid side impl looks good otherwise

packages/solid/tests/solid-plugin.test.ts
RocketBoyMusic: Yeah, agreed honestly. I’m probably over testing because I just kept hitting this repeatedly on Windows (outside WSL) and wanted to avoid regression

Adictya: Yeah we can keep it then, @kommander  can take a better call on that. Anyway thanks for the contrib!

kommander:
But now it doesn't prove to be working on windows?

RocketBoyMusic:
Fair point. I removed the Windows smoke gate because it was failing here, but I tested this on Windows and this fix resolved the issue for me. Plus I agree with Adictya that it was a bit extra

packages/core/src/zig.ts
```diff
@@ -22,7 +22,23 @@ import {
 import { isBunfsPath } from "./lib/bunfs"
 import { attributesWithLink } from "./utils"
 
-const module = await import(`@opentui/core-${process.platform}-${process.arch}/index.ts`)
+const nativePackageName = `@opentui/core-${process.platform}-${process.arch}`
+let module: { default: string }
+try {
+  module = (await import(`${nativePackageName}/index.ts`)) as { default: string }
```
kommander: I think this might break standalone executable compilation for Opencode. I remember this was cumbersome to get right. Needs to be tested with Opencode to build the executable and run it isolated from the opencode repo to make sure bundled deps works.

RocketBoyMusic: Yeah, I agree, testing this it doesn't seem to work and breaks the Bun standalone bundling. I’ll revert to the literal import pattern and keep it focused on the Solid Windows path-separator fix and tests

kommander:
Does this still solve an issue? Opencode works with main on windows.

PR #640, opened 2026-02-07
fix(buffer): add errdefer for allocations in OptimizedBuffer init

simonklee:
Pre-allocate buffers before struct assignment so errdefer can free them on partial failure.

Prev, when one of the later allocations failed with OOM, earlier successful allocations were not released before returning, so if the caller retried it it would leak.


simonklee:
Forgot to mention, this is mostly pedantic, an OOM on a few KB at process start means the system is in catastrophic failure and you have bigger problems.

kommander:
Buffers are used throughout and often destroyed and recreated, so it can definitely happen.

PR #645, opened 2026-02-07
Add Renderable.scrollIntoView()

cevr:
## Problem / Intent

There's no way to scroll a renderable into view from a ref. Consumers (e.g. Solid/React) need to call `ref.scrollIntoView()` to reveal a focused or selected item inside a ScrollBox — mirroring the DOM API.

## Approach

`Renderable.scrollIntoView()` walks up the parent chain looking for a ScrollBox ancestor via duck typing (`'scrollChildIntoView' in current`), avoiding a circular dependency between Renderable and ScrollBox. The private `ScrollBox.scrollChildIntoView(child)` does the viewport math — scrolls the minimum amount to make the child fully visible, no-ops if already in view.

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -429,6 +429,21 @@ export abstract class Renderable extends BaseRenderable {
     this.emit(RenderableEvents.BLURRED)
   }
 
+  /**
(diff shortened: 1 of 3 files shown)


simonklee:
Superseded-by: #724

PR #653, opened 2026-02-09
feat(extmarks): add onEncounter and onDeletion hooks for virtual extmarks

### LINKED ISSUE #652: Add onEncounter and onDeletion hooks for virtual extmarks
## Summary

Virtual extmarks currently hard-code cursor skipping and whole-range deletion. Consumers have no way to customize what happens when the cursor hits a virtual extmark or when the user tries to delete one. This adds per-extmark `onEncounter` and `onDeletion` callbacks to `ExtmarkOptions`, enabling patterns like "select then act" — where the first arrow press stops at the extmark and a second press skips past it.

**Motivation:** [opencode#8501](https://github.com/anomalyco/opencode/issues/8501) — pasted text summarized into virtual extmarks needs a way to expand inline. Without hooks, consumers must intercept keyboard events before the controller, duplicate private adjacency-detection logic, and fight with the wrapped `setCursorByOffset`. Hooks at the controller level are the clean solution.

## API

```typescript
interface ExtmarkEncounter {
  extmark: Readonly<Extmark>
  direction: "left" | "right" | "up" | "down" | "set"
  skip(): void           // default behavior — jump past
  setCursor(offset: number): void  // place cursor at specific offset
}

interface ExtmarkDeletionEncounter {
  extmark: Readonly<Extmark>
  direction: "backward" | "forward"
  deleteExtmark(): void  // default behavior — delete entire range
  prevent(): void        // block deletion
}

// New optional fields on ExtmarkOptions:
onEncounter?: (encounter: ExtmarkEncounter) => void
onDeletion?: (encounter: ExtmarkDeletionEncounter) => void
```

## Behavior

- Only fires for **virtual** extmarks (non-virtual don't trigger cursor skipping)
- Fully backwards compatible — omitting callbacks preserves current behavior
- Callback must call exactly one action method; if none called, default behavior runs as fallback
- Callbacks stored in separate maps (not on `Extmark` interface), following the existing `metadata` pattern — excluded from undo/redo snapshots

## Scope

**In scope:**
- `onEncounter` hooks wired into all 5 wrapped cursor methods (`moveCursorLeft`, `moveCursorRight`, `moveUpVisual`, `moveDownVisual`, `setCursorByOffset`)
- `onDeletion` hooks wired into `deleteCharBackward` and `deleteChar`
- Callback lifecycle management in `create()`, `delete()`, and `clear()`
- Tests for both hooks
- Updated extmarks demo with "select then act" interaction

**Out of scope:**
- Mutable extmark properties (e.g., changing `styleId` in place)
- Controller-level hooks (vs per-extmark)
- Insertion hooks
- Consumer-side changes (opencode)

jorgeraad:
Closes #652

## Summary

Adds per-extmark `onEncounter` and `onDeletion` callbacks to `ExtmarkOptions`, letting consumers
control what happens when the cursor hits a virtual extmark or the user tries to delete one.

## Changes

**Core hooks** (`packages/core/src/lib/extmarks.ts`):
- `ExtmarkEncounter` and `ExtmarkDeletionEncounter` interfaces with action methods (`skip()`, `setCursor()`, `deleteExtmark()`, `prevent()`)
- `onEncounter` and `onDeletion` optional fields on `ExtmarkOptions`
- Callback maps stored alongside existing `metadata` map, cleaned up in `delete()` and `clear()`
- `onEncounter` wired into all 5 wrapped cursor methods (left, right, up, down, setCursorByOffset)
- `onDeletion` wired into `deleteCharBackward` and `deleteChar`
- Fallback to default behavior when callback doesn't call any action method

**Demo** (`packages/core/src/examples/extmarks-demo.ts`):
- Updated to use the new hooks — arrowing into an extmark stops and highlights it, then you can
  press Enter to expand it, Backspace to delete it, or arrow again to skip past

## Demo

https://drive.google.com/file/d/1w83taLW36FnsmEeZGCj-49q9H00FuN12/view?usp=sharing

## Backwards compatibility

Fully backwards compatible. Omitting `onEncounter`/`onDeletion` preserves existing skip and delete behavior unchanged.

### DIFF
--- a/packages/core/src/examples/extmarks-demo.ts
+++ b/packages/core/src/examples/extmarks-demo.ts
@@ -53,7 +53,12 @@ function clearSelection(): void {
 }
 
 function createOnEncounter(label: string) {
-  return (encounter: { extmark: { id: number; start: number; end: number }; direction: string; skip: () => void; setCursor: (offset: number) => void }) => {
+  return (encounter: {
+    extmark: { id: number; start: number; end: number }
+    direction: string
+    skip: () => void
+    setCursor: (offset: number) => void
+  }) => {
     if (selectionState.id === encounter.extmark.id) {
       clearSelection()
       hookState.message = `Skipped past ${label}`
--- a/packages/core/src/lib/extmarks.ts
+++ b/packages/core/src/lib/extmarks.ts
@@ -578,10 +578,7 @@ export class ExtmarksController {
     return true
   }
 
-  private invokeDeletionHook(
-    extmark: Extmark,
-    direction: ExtmarkDeletionEncounter["direction"],
-  ): boolean {
+  private invokeDeletionHook(extmark: Extmark, direction: ExtmarkDeletionEncounter["direction"]): boolean {
     const onDeletion = this.onDeletionCallbacks.get(extmark.id)
     if (!onDeletion) return true
 


kommander:
Needs to make sure it doesn't break current opencode behaviour. It doesn't seem to interfere, but need someone to check with opencode.

PR #656, opened 2026-02-09
text-buffer: encapsulate internals behind accessors

simonklee:
Stop view, render, and edit layers from reaching into rope, mem_registry,
allocator, tab_width, and width_method fields directly. All external access
now goes through accessor methods.

This decouples the read/render paths from the rope data structure, which is
the prerequisite for making UnifiedTextBuffer generic over backend type.

### DIFF
--- a/packages/core/src/examples/focus-restore-demo.ts
+++ b/packages/core/src/examples/focus-restore-demo.ts
@@ -1,0 +1,310 @@
+#!/usr/bin/env bun
+
+// Interactive demo to test the focus restore fix on Windows Terminal.
+//
+// How to test:
+//   1. Run from the example selector, or: bun src/examples/focus-restore-demo.ts
+//   2. Move the mouse around - you should see the mouse position update live
+//   3. Alt-tab away from the terminal, then alt-tab back
+//   4. Move the mouse again - if the fix works, mouse tracking resumes immediately
+//   5. Try minimizing and restoring the window too
+//   6. Press Escape to return to menu, Ctrl+C to exit
+
+import {
+  type CliRenderer,
+  createCliRenderer,
+  BoxRenderable,
+  TextRenderable,
+  RGBA,
+  TextAttributes,
+  type MouseEvent,
+} from "../index"
+import { setupCommonDemoKeys } from "./lib/standalone-keys"
+
+let container: BoxRenderable | null = null
+let mouseArea: BoxRenderable | null = null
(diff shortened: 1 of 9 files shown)


packages/core/src/zig/tests/edit-buffer_test.zig
```diff
@@ -349,14 +349,14 @@ test "EditBuffer - type and move around single tab" {
 
     const cursor1 = eb.getCursor(0).?;
     try std.testing.expectEqual(@as(u32, 0), cursor1.row);
-    _ = iter_mod.lineWidthAt(&eb.tb.rope, 0);
+    _ = iter_mod.lineWidthAt(eb.tb.rope(), 0);
 
-    _ = iter_mod.getGraphemeWidthAt(&eb.tb.rope, &eb.tb.mem_registry, 0, cursor1.col, eb.tb.tab_width, eb.tb.width_method);
+    _ = iter_mod.getGraphemeWidthAt(eb.tb.rope(), eb.tb.memRegistry(), 0, cursor1.col, eb.tb.tabWidth(), eb.tb.widthMethod());
```
kommander: Couldn't a method like this be `tb.getGraphemeWidthAt(0, cursor1.col)` instead of passing all the `tb` props? It can just use whatever it has at its disposal to solve that.

Similar for other methods as well, then we don't have to pass the rope around and no need to abstract the rope.

packages/core/src/zig/text-buffer-view.zig
```diff
@@ -277,7 +277,7 @@ pub const UnifiedTextBufferView = struct {
         if (total_width == 0) return .{ .char_count = 0, .width = 0 };
         if (total_width <= max_width) return .{ .char_count = total_width, .width = total_width };
 
-        const wrap_offsets = chunk.getWrapOffsets(&self.text_buffer.mem_registry, self.text_buffer.allocator, self.text_buffer.width_method) catch {
+        const wrap_offsets = chunk.getWrapOffsets(self.text_buffer.memRegistry(), self.text_buffer.getAllocator(), self.text_buffer.widthMethod()) catch {
```
kommander: Maybe not for this PR but I have been thinking for longer already that things like this could be reversed, instead of passing three TextBuffer props, it could be `tb.getWrapOffsetsFor(chunk)` and the chunk becomes more a pure data carrier than logic bearer.

PR #662, opened 2026-02-10
feat(react): support textarea onInput callbacks

remorses:
### What
- Add `onInput` support for React `<textarea>`.
- Wire it directly to textarea content changes in the React renderer.

### Why
- `onInput` was mapped for `<input>` but not `<textarea>`.
- Consumers had to use `onContentChange` + `ref.current?.plainText`, which is easy to make stale in app update flows (we hit this while fixing critique dropdown search).

### Scope
- Minimal change: textarea `onInput` listens to `editBuffer` `content-changed`.
- No behavior change to `onContentChange`.

### Tests
- Added `packages/react/tests/textarea-events.test.tsx`.
- Verifies latest text for both single-line and multiline edits.

### DIFF
--- a/packages/react/src/utils/index.ts
+++ b/packages/react/src/utils/index.ts
@@ -10,30 +10,25 @@ import {
 } from "@opentui/core"
 import type { Instance, Props, Type } from "../types/host"
 
-interface TextareaListenerState {
-  onInput?: (value: string) => void
-  onContentChange?: (event: unknown) => void
-}
+const textareaInputListeners = new WeakMap<TextareaRenderable, () => void>()
 
-const textareaListenerStates = new WeakMap<TextareaRenderable, TextareaListenerState>()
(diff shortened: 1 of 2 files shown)


msmps:
i'll need to look into this and check a few things before merge

kommander:
I was even considering removing that event from Input. What's the difference to onChange?

remorses:
The problem with onContentChange is that it does not pass the textarea value to the callback and using the .plainText field of the renderable can return a stale value in React

kommander:
It does not pass the plainText value in the onContentChanged callback because it's not needed for everything and implies an FFI call to copy the value into the js heap, which would be for every keystroke and too heavy for larger documents.

That sounds like we should fix the stale value in react then.

remorses:
Agree, I will look into that

PR #664, opened 2026-02-10
fix: emit INPUT event for all mutating edit actions in InputRenderable

kitlangton:
InputRenderable now emits INPUT events for all text-mutating operations (Ctrl+W, Ctrl+U, Ctrl+K, undo/redo), not just single-character deletion. Fixes search filters in TUI dialogs that rely on onInput events.

## Test plan
- Two new tests verify Ctrl+W and Alt+Backspace emit INPUT events
- All 49 existing Input tests pass


kitlangton:
> Are we also missing `deleteLine` for Ctrl+Shift+D?

Indeed! Fixed :)

PR #666, opened 2026-02-10
fix: stable grapheme across frames

### LINKED ISSUE #667: Renderer diff uses grapheme pool IDs as identity
Cell.char for graphemes is an allocation handle (class|generation|slot), but diff treats it as content identity; because graphemes are reallocated each frame, IDs churn and diff correctness/perf break.

#666

simonklee:
The renderer diff compares `Cell.char` values to decide what changed between frames. For grapheme cells, the char value embeds a pool ID. Without interning, drawing the same "👋" on two consecutive frames produces two different IDs, so the diff treats every grapheme cell as dirty every frame.

This adds content-addressed interning to GraphemePool: when bytes are already live in the pool, alloc() returns the existing ID instead of allocating a new slot. The intern map is maintained on refcount transitions (interned at 0->1, removed at 1->0), so it stays in sync with the pool's own lifetime tracking.

Maybe most important change is to switch from setRaw() to set(). set() is a bit more expensive, but transfers ownership through the tracker. Its also just called for changed cells, so the perf impact should be acceptable.

Fix #667

### DIFF
--- a/packages/core/src/zig/renderer.zig
+++ b/packages/core/src/zig/renderer.zig
@@ -739,19 +739,9 @@ pub const CliRenderer = struct {
                 }
                 runLength += 1;
 
-                // Update the current buffer with the new cell
-                self.currentRenderBuffer.setRaw(x, y, nextCell.?);
-
-                // If this is a grapheme start, also update all continuation cells
-                if (gp.isGraphemeChar(nextCell.?.char)) {
-                    const rightExtent = gp.charRightExtent(nextCell.?.char);
-                    var k: u32 = 1;
-                    while (k <= rightExtent and x + k < self.width) : (k += 1) {
-                        if (self.nextRenderBuffer.get(x + k, y)) |contCell| {
-                            self.currentRenderBuffer.setRaw(x + k, y, contCell);
-                        }
-                    }
-                }
+                // Update grapheme/link trackers (and continuation cells), so current buffer
+                // retains grapheme ownership after next buffer clear and IDs remain stable.
+                self.currentRenderBuffer.set(x, y, nextCell.?);
 
                 cellsUpdated += 1;
             }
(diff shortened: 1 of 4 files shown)


packages/core/src/zig/renderer.zig
```diff
@@ -739,19 +739,9 @@ pub const CliRenderer = struct {
...
-                    var k: u32 = 1;
-                    while (k <= rightExtent and x + k < self.width) : (k += 1) {
-                        if (self.nextRenderBuffer.get(x + k, y)) |contCell| {
-                            self.currentRenderBuffer.setRaw(x + k, y, contCell);
-                        }
-                    }
-                }
+                // Update grapheme/link trackers (and continuation cells), so current buffer
+                // retains grapheme ownership after next buffer clear and IDs remain stable.
+                self.currentRenderBuffer.set(x, y, nextCell.?);
```
kommander: Interesting, I remember doing the "set" manually explicitly without refcounting, because the `currentRenderBuffer` is never cleared, so never dereferenced and never used to be rendered into another buffer.

simonklee: with interning we need currentRenderBuffer to hold a real ref too. If we copy with setRaw then nextRenderBuffer.clear() drops the only tracked ref to 0, the intern entry gets removed, and the same grapheme allocs a fresh ID next frame (churn again). 

Now we transfers ownership into currentRenderBuffer's grapheme tracker, and clear/overwrite/deinit paths release it, so refs stay balanced. It also keeps bookkeeping in one place instead of reimplementing it in renderer.

packages/core/src/renderables/__tests__/__snapshots__/LineNumberRenderable.test.ts.snap
```diff
@@ -104,7 +104,7 @@ exports[`LineNumberRenderable maintains stable visual line count when scrolling
  │  41 FEATURES:                 │ 
  │  42   ✓ Grapheme-aware        │ 
  │     cursor movement           │ 
- │  43   ✓ Unicode (emoji 🌟     │ 
+ │  43   ✓ Unicode (emoji 🌟      │
```
kommander: This is now off. Is this flaky or why would that change with?

Image

Issue #661, opened 2026-02-10
Mouse moves translate to mouse scrolling under URxvt


demostanis:
when using the rxvt-unicode terminal, when i scroll and move my mouse a little bit, the whole screen moves in unpredictable ways:

https://github.com/user-attachments/assets/f9dc0436-9ffa-47f9-ad68-6cfa88036842

if needed, my whole setup is at https://github.com/demostanis/demolinux,
but ive been able to reproduce with the latest rxvt-unicode in Arch repos with an empty .Xresources

simonklee:
I think we're missing to parse the motion/drag detection and because of it misclassifying things as down events causing the behavior you're seeing.

demostanis:
@simonklee thank you for trying to figure out the issue, however that didn't fix it
(to test out i git clone'd the opencode repo and ran bun install && bun dev)

simonklee:
> [@simonklee](https://github.com/simonklee) thank you for trying to figure out the issue, however that didn't fix it (to test out i git clone'd the opencode repo and ran bun install && bun dev)

I'm not sure we cut a release yet.

kommander:
@demostanis it's only on the beta branch currently. You can test with `bunx opencode-ai@beta`.

demostanis:
@kommander did not resolve it either :/

simonklee:
I’ll have another look and try to reproduce it locally with XWayland.

demostanis:
@simonklee did you try to reproduce it locally? do you need help?

simonklee:
> [@simonklee](https://github.com/simonklee) did you try to reproduce it locally? do you need help?

Thanks - I might send you script later today tbh. I created a minimal thing that just captures the mouse events before parsing (but its on my desktop at home).

I tried to reproduce on arch + sway + Xwayland and my hypothesis for what was going wrong didn't turn out to be correct, but i found some other things. These are sitting in a PR atm but should be merged soon #679.

demostanis:
i tried asking gemini 3 flash to debug the issue and he was able to solve the problem:
https://opncd.ai/share/Lg0TWAsu
(i recorded opencode with `script`, and told gemini to read the raw output with the opencode and opentui sources)
```
From 014e18ac5192b49e89504f7322b75815a59f83f7 Mon Sep 17 00:00:00 2001
From: demostanis <demostanis@protonmail.com>
Date: Mon, 16 Feb 2026 16:15:16 +0100
[...]

PR #669, opened 2026-02-10
fix(mouse): mouse event parse + tests

simonklee:
- Added a bunch of tests for mouse parsing
- Fixed the missing motion/drag detection causing issue in #661

I did not test the 661 in a X11 env since i'm on wayland, lets see if reporter
can confirm the fix otherwise i can try in a VM later


kommander:
Could it break existing behavior?

simonklee:
> Could it break existing behavior?

The current implementation actually is broken. Since we didn’t parse the bit it gets interpreted as down event. Most terminals support the other mode so this issue isn’t noticed unless you have x11 and use urxvt or something similar.

PR #673, opened 2026-02-11
introduce native-span-feed output stream

kommander:
Introduces a stream implementation, to feed data efficiently from native/zig to bun/typescript, using spans of memory in a chunked ring buffer, allowing full zero-copy consumption through to typescript, with and async event interface.

Intentionally named `NativeSpanFeed` to not be confused with a node ReadStream interface. A ReadStream interface is not possible without copy, but the feed is intentionally zero-copy to not cause GC pressure in Bun.

Not actively used in the renderer yet. It will allow to stream all ANSI output to actual WritableStreams in Bun.

### DIFF
--- a/packages/core/src/testing/mock-keys.test.ts
+++ b/packages/core/src/testing/mock-keys.test.ts
@@ -181,10 +181,11 @@ describe("mock-keys", () => {
 
     const startTime = Date.now()
     await mockKeys.pressKeys(["a", "b"], 10) // 10ms delay between keys
+    const totalElapsed = Date.now() - startTime
 
     expect(timestamps).toHaveLength(2)
     expect(timestamps[1] - timestamps[0]).toBeGreaterThanOrEqual(8) // Allow some tolerance
-    expect(timestamps[1] - timestamps[0]).toBeLessThan(20)
+    expect(totalElapsed).toBeGreaterThanOrEqual(16)
   })
 
   test("pressKey with shift modifier", () => {


simonklee:
Tried to repro a couple issues with `close`.

- First) a handler that still holds the `Uint8Array` view after `await` may read freed memory. 
- Second) the wrapper can destroy a stream even though native explicitly refused close.


```ts
import { test, expect } from "bun:test"
import { NativeSpanFeed } from "../NativeSpanFeed"
import { resolveRenderLib } from "../zig"

const lib = resolveRenderLib()

function nextTick(): Promise<void> {
  return new Promise((resolve) => setTimeout(resolve, 0))
}

test("close should not destroy immediately while async handler is still pending", async () => {
  const stream = NativeSpanFeed.create({ chunkSize: 64, initialChunks: 1 })
  const ptr = stream.streamPtr

  let release!: () => void
  const gate = new Promise<void>((resolve) => {
    release = resolve
  })

  let handlerStarted = false
  let handlerSettled = false

  stream.onData(async (_data) => {
    handlerStarted = true
    await gate
    handlerSettled = true
  })

  const payload = new Uint8Array(64).fill(0xaa)
  lib.streamWrite(ptr, payload)
  lib.streamCommit(ptr)
  stream.drainAll()

  expect(handlerStarted).toBe(true)

  stream.close()

  const destroyedImmediately = (stream as any).destroyed === true

  release()
  await nextTick()

  expect(handlerSettled).toBe(true)

  try {
    expect(destroyedImmediately).toBe(false)
  } finally {
    if (!(stream as any).destroyed) {
      lib.destroyNativeSpanFeed(ptr)
    }
  }
})

test("close should not destroy when native close reports Busy", () => {
  const stream = NativeSpanFeed.create({ chunkSize: 64, initialChunks: 1, autoCommitOnFull: false })
  const ptr = stream.streamPtr

  const reserve = lib.streamReserve(ptr, 1)
  expect(reserve.status).toBe(0)

  try {
    stream.close()
  } catch {
    // If close starts throwing on Busy, that's acceptable for this assertion.
  }

  const destroyedAfterBusyClose = (stream as any).destroyed === true

  try {
    expect(destroyedAfterBusyClose).toBe(false)
  } finally {
    if (!(stream as any).destroyed) {
      lib.streamCommitReserved(ptr, 0)
      lib.streamClose(ptr)
      lib.destroyNativeSpanFeed(ptr)
    }
  }
})

```

kommander:
a) is a constraint and design decision that needs to be carefully handled for zero-copy low js GC and memory overhead, that API should _never_ be exposed publicly to users of the lib.
b) That looks like a potential issue, yes. Should wait for handlers to be settled and schedule the close.

packages/core/src/NativeSpanFeed.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  private decrementRefcount(chunkIndex: number): void {
+    if (this.stateBuffer && chunkIndex < this.stateBuffer.length) {
+      const prev = this.stateBuffer[chunkIndex]
+      this.stateBuffer[chunkIndex] = prev > 0 ? prev - 1 : 0
+    }
+  }
+
+  private drainOnce(): number {
+    if (!this.drainBuffer || this.draining) return 0
```
simonklee:
```
if (this.pendingClose) return 0
```

so you avoid starting new draining after close intent.

packages/core/src/NativeSpanFeed.ts
```diff
@@ -0,0 +1,299 @@
...
+    const count = this.lib.streamDrainSpans(this.streamPtr, this.drainBuffer, capacity)
+    if (count === 0) return 0
+
+    this.draining = true
+    const spans = SpanInfoStruct.unpackList(this.drainBuffer.buffer, count)
+    let firstError: unknown = null
+
+    try {
+      for (const span of spans) {
+        if (this.pendingClose) break
```
simonklee: Dront break here.

packages/core/src/NativeSpanFeed.ts
```diff
@@ -0,0 +1,299 @@
...
+            if (result && typeof result.then === "function") {
+              asyncResults ??= []
+              asyncResults.push(result)
+            }
+          } catch (e) {
+            firstError ??= e
+          }
+        }
+
+        if (this.pendingClose) break
```
simonklee: Instead just check if you should stop at the end of the span.
```
const shouldStopAfterThisSpan = this.pendingClose
```

packages/core/src/NativeSpanFeed.ts
```diff
@@ -0,0 +1,299 @@
...
+          const chunkIndex = span.chunkIndex
+          this.pendingAsyncHandlers += 1
+          Promise.allSettled(asyncResults).then(() => {
+            this.decrementRefcount(chunkIndex)
+            this.pendingAsyncHandlers -= 1
+            this.processPendingClose()
+          })
+        } else {
+          this.decrementRefcount(span.chunkIndex)
+        }
```
simonklee, line 281: Then at the end of the span you can break out

```
if (shouldStopAfterThisSpan) break
```

PR #676, opened 2026-02-12
feat: Added HoverCursorStyle to APIs using OSC 22

### LINKED ISSUE #611: feat: custom cursor styles
I saw something very cool recently and was wondering can we get it inside opentui.

Cursor styles on hover.
The TUI shown in the video is ampcode.

https://github.com/user-attachments/assets/16f27d6e-6813-4aa5-aa87-ac952491ceae

viralcodex:
Fixes: #611

Hi @kommander,

Added Custom cursor styles for the openTUI API through MouseCursorStyle.
Includes: pointer, disabled, crosshair, and move.

Used OSC 22 and demo-ed on kitty terminal.

I have added all of them for usage (and future usage) but I can remove them if needed. 

Child elements can override the parent hoverCursorStyle.

Also on shutdown the states reset to default cursors (added shutdown code in both native and typescript code)

https://github.com/user-attachments/assets/c877730f-bada-453a-a6ee-a95f82e47bf2

Let me know if changes are needed, will do accordingly.

Thanks :)


packages/core/src/ansi.ts
```diff
@@ -9,6 +9,9 @@ export const ANSI = {
   moveCursor: (row: number, col: number) => `\x1b[${row};${col}H`,
   moveCursorAndClear: (row: number, col: number) => `\x1b[${row};${col}H\x1b[J`,
 
+  setMousePointer: (shape: string) => `\x1b]22;${shape}\x07`,
```
kommander: I think this should be on the zig side, passing the option down.

packages/core/src/renderer.ts
```diff
@@ -1274,6 +1276,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
           })
           maybeRenderable.processMouseEvent(event)
         }
+
+        this.updateMousePointerOnHover(maybeRenderable)
```
kommander: Not every Renderable is potentially hoverable, but I see where that is coming from. I thought that maybe that could be something that an app controls itself? It would be like:

```ts
r.onMouseOver = (evt) => evt.target.ctx.setCursorStyle(...)
```

packages/core/src/renderer.ts
```diff
@@ -1369,6 +1373,29 @@ export class CliRenderer extends EventEmitter implements RenderContext {
...
+  private updateMousePointerOnHover(renderable: Renderable | undefined): void {
+    const cursorStyle = renderable?.getCurrentHoverCursorStyle()
+    if (cursorStyle) {
+      this.setMousePointer(cursorStyle)
+    } else {
+      this.resetMousePointer()
+    }
+  }
+
+  public setMousePointer(shape: MousePointerStyle): void {
```
kommander: There is a method `setCursorStyle` already, which takes options that could be adjusted. It's currently this:

```ts
 public static setCursorStyle(
    renderer: CliRenderer,
    style: CursorStyle,
    blinking: boolean = false,
    color?: RGBA,
  ): void {
```

It's messy, so we could make that proper like:

```ts
public static setCursorStyle(
    renderer: CliRenderer,
    options: CursorStyleOptions
  ): void {
```

Where the `CursorStyleOptions` become:

```ts
export interface cursorstyleoptions {
  style: "block" | "line" | "underline"
  blinking: boolean,
  color: RGBA,
  cursor: "default" | "pointer" | "text" | "crosshair" | "move" | "not-allowed"
}
```

Then it can be properly packed as FFI Struct in `zig-structs.ts` and passed down to zig, which can hold the cursor state and update the changed values.

kommander (commented):
We can keep this lower level for the lib and let solid/react or the app level take care of how and when to apply the cursor style.

packages/core/src/renderer.ts
viralcodex: there are two setCursorStyles here in the class, one is static one that you mentioned and one instance method, should I update that with this as well?

Then I will have to change the signature at a lot of places

packages/core/src/ansi.ts
viralcodex: ok moving the whole setting the mouse pointer to zig side

viralcodex:
Hi @kommander,
Made some changes, unified the whole styles thing as far as I could understand to CursorStyleOptions using a single struct to zig.

Also, since mouse pointer and Cursor styles were mandatory in the object/struct, I had to pass defaults everytime I was making an update to one of them, so made all props optional, and update as the app required.

Also made change to let the app decide instead of passing a prop to the elements.

I could be missing something or may have written something in a wrong way, so please let me know and I'll fix it.

viralcodex:
here's verifying both cursor and mouse pointer:

https://github.com/user-attachments/assets/cd7932dc-8699-4ecc-a68f-1266642ed272

packages/core/src/renderer.ts
```diff
@@ -1757,6 +1763,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     this._destroyFinalized = true
     this._destroyPending = false
 
+    this.resetMousePointer()
```
kommander: This should happen in `Terminal.zig`, in the shutdown sequence I think.

kommander:
Looks way better 👍 thanks! You think we could update the docs here as well, like the website docs? cc @simonklee maybe you can glimpse over it as well, I think it looks good and can be iterated on.

viralcodex:
@kommander thanks, yes we can, should I add the web docs change here as well, or should I raise a separate PR after merging this one? Whatever works for you guys :)

packages/core/src/renderer.ts
viralcodex: Must have missed it from 1st commit, will do it.

packages/core/src/renderer.ts
```diff
@@ -1571,10 +1572,10 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     this.lib.setCursorPosition(this.rendererPtr, x, y, visible)
   }
 
-  public setCursorStyle(style: CursorStyle, blinking: boolean = false, color?: RGBA): void {
-    this.lib.setCursorStyle(this.rendererPtr, style, blinking)
-    if (color) {
-      this.lib.setCursorColor(this.rendererPtr, color)
+  public setCursorStyle(options: CursorStyleOptions): void {
```
simonklee, line 1575: `setCursorStyle({ style })`  sets blinking to false instead of leaving it unchanged if you omit the blinking senitel, right? because options.blinking == 1 is checked in lib.zig

simonklee: This is also an API change so we need to update docs. @kommander you dont mind breaking this, right?

viralcodex: > `setCursorStyle({ style })` sets blinking to false instead of leaving it unchanged if you omit the blinking senitel, right? because options.blinking == 1 is checked in lib.zig

yes it does, if it breaks the flow, I can try to fix it

kommander: I don't mind breaking the API, nothing is declared stable yet.

kommander:
When @simonklee is fine we can merge this and document in another PR

viralcodex:
Thanks, will raise another PR and link it here for the documentation

packages/core/src/renderer.ts
viralcodex: > `setCursorStyle({ style })` sets blinking to false instead of leaving it unchanged if you omit the blinking senitel, right? because options.blinking == 1 is checked in lib.zig

Have done the fix by checking the current terminal state and only updating the values if they are passed in the object through FFI

simonklee (approved):

PR #680, opened 2026-02-13
fix(solid): handle ScrollBox parent mismatch in getParentNode

IlyaGulya:
## Summary

- ScrollBox delegates `add`/`remove` to its internal `content` wrapper (`scrollbox → wrapper → viewport → content`), so children report `content` as their parent. The reconciler passes `scrollbox` itself — causing the identity check in `cleanChildren` (`getParentNode(el) === parent`) to always fail, leaving stale nodes behind.
- Fix: make `_getParentNode` in `reconciler.ts` detect when the child's parent is the `content` of a `ScrollBoxRenderable` and return the ScrollBox instead. This keeps `universal.js` untouched.

## Root cause

1. `ScrollBox.add(obj)` delegates to `this.content.add(obj)`, which sets `obj.parent = content`
2. The reconciler's `parent` argument is `scrollbox` (the JSX parent)
3. `cleanChildren()` checked `getParentNode(el) === parent` → `content !== scrollbox` → `false`
4. `removeNode()` was skipped → stale children remained in the tree

Only triggered with multiple dynamic siblings (two+ `<For>`, `<For>` + `<Show>`, etc.) because that's when `cleanChildren` uses the marker-based path where the guard exists.

## AI Disclosure

This PR was created with assistance of AI agent (Claude Code + Claude Opus 4.6)

### DIFF
--- a/packages/solid/src/reconciler.ts
+++ b/packages/solid/src/reconciler.ts
@@ -8,6 +8,7 @@ import {
   parseColor,
   Renderable,
   RootTextNodeRenderable,
+  ScrollBoxRenderable,
   SelectRenderable,
   SelectRenderableEvents,
   TabSelectRenderable,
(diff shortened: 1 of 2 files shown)


kommander:
cc @Adictya

packages/solid/src/renderer/universal.js
```diff
@@ -173,9 +173,8 @@ export function createRenderer({
       for (let i = current.length - 1; i >= 0; i--) {
         const el = current[i]
         if (node !== el) {
-          const isParent = getParentNode(el) === parent
```
Adictya, line 176: Can you instead add a special case for scrollbox in packages/solid/src/reconciler.ts:L145. I would rather keep this file as intact as possible

IlyaGulya: Sure, thanks, will do next week

IlyaGulya: Reverted `universal.js`, moved the fix to `_getParentNode` in `reconciler.ts` instead — it detects when a child's parent is ScrollBox's internal `content` and returns the ScrollBox. Also cleaned up redundant tests.

PR #682, opened 2026-02-13
fix(input): handle surrogate pairs in stdin buffer

hugojosefson:
### Bug

`extractCompleteSequences` in `stdin-buffer.ts` iterates the buffer using `remaining[0]` and `pos++`, which operate on UTF-16 code units. But characters above U+FFFF (emoji, CJK Extension B, etc.) are stored as surrogate pairs, two code units.

Existing code splits these into two lone surrogates, which `TextEncoder.encode()` will then convert to U+FFFD (replacement character) downstream.

For example, typing 👍 (U+1F44D) in the input field produces `��` instead.

Note that this doesn't seem to happen with pasted characters, since they probably go through a different code path. It does happen when I type an emoji using compose key in linux. I suppose it could also happen for other characters like CJK etc...

### Fix

Check whether the code unit is a high surrogate (0xD800–0xDBFF) followed by a low surrogate (0xDC00–0xDFFF), and if so keep both as a single sequence entry.

Two regression tests added:
- emoji as sole input
- emoji mixed with ASCII

### DIFF
--- a/packages/core/src/lib/stdin-buffer.test.ts
+++ b/packages/core/src/lib/stdin-buffer.test.ts
@@ -50,6 +50,25 @@ describe("StdinBuffer", () => {
       processInput("hi👍bye")
       expect(emittedSequences).toEqual(["h", "i", "👍", "b", "y", "e"])
     })
+
+    it("should handle emoji split across chunks", () => {
+      processInput("\uD83D")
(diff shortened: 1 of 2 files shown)


michalrus:
Resolves:
- https://github.com/anomalyco/opencode/issues/13182

Thanks!

PR #683, opened 2026-02-13
Deno 🦖

simonklee:
Very much WIP.

https://github.com/user-attachments/assets/cc1bdffc-947c-464f-8275-4f1f35da4548


remorses:
Are you sure this is better than switching to NAPI?

NAPI is more high level so it would be easier to port to both Node.js, Deno and [even more languages like python](https://github.com/remorses/napi-python) while also deleting a lot of code we need now for callbacks, structs, etc

I am also worried about issues with callbacks in the ffi implementations, for example when implementing the yoga ffi bindings I had to work around [platforms ffi limitations on ARM macOS](https://github.com/remorses/bun-yoga/commit/0b3b46e05f1cca96ef68a064f2987ba6e426b876), I think NAPI is probably different in subtle ways that could break and be a nightmare to debug

simonklee:
> Are you sure this is better than switching to NAPI?
> 
> NAPI is more high level so it would be easier to port to both Node.js, Deno and [even more languages like python](https://github.com/remorses/napi-python) while also deleting a lot of code we need now for callbacks, structs, etc
> 
> I am also worried about issues with callbacks in the ffi implementations, for example when implementing the yoga ffi bindings I had to work around [platforms ffi limitations on ARM macOS](https://github.com/remorses/bun-yoga/commit/0b3b46e05f1cca96ef68a064f2987ba6e426b876), I think NAPI is probably different in subtle ways that could break and be a nightmare to debug

No, its very much a test. Trying to get a feel for how it would look going down this route.

remorses:
> No, its very much a test. Trying to get a feel for how it would look going down this route.

Ok, the issue with going with NAPI is that Bun has a lot of bugs with NAPI on windows. For example the yoga ffi napi branch fails on windows even if it's pretty simple

So we would need to solve that first before switching. Or pray Jarred

simonklee:
Also it’s not quite a 1:1 comparison. Parts of this is about ffi another part is about runtime.

PR #686, opened 2026-02-13
fix: discard partial OSC/DCS/APC sequences on flush to prevent garbage text in input

bbartels:
## Summary
Fixes https://github.com/anomalyco/opencode/issues/6147 — Windows Terminal sends unsolicited OSC queries (10/11/4) over SSH whose responses can arrive fragmented and leak into the TUI input field as garbage text like `eeee/eeee/eeee`.
## Problem
When data arrives fragmented over SSH, the `StdinBuffer` timeout can fire mid-sequence. The `flush()` method emits the incomplete buffer as-is, which downstream `parseKeypress()` no longer recognizes as an escape sequence (since it may be missing the `ESC ]` prefix after the first chunk timed out separately). The payload then gets interpreted as regular text input.
## Fix
Added `isPartialEscapeSequence()` check in `flush()` that silently discards incomplete OSC (`ESC ]`), DCS (`ESC P`), and APC (`ESC _`) sequences instead of emitting them as input. These sequence types are never valid keypresses.
CSI (`ESC [`) and SS3 (`ESC O`) are intentionally **not** discarded because a timed-out `ESC [` could be a legitimate `Alt+[` keypress.

### DIFF
--- a/packages/core/src/lib/stdin-buffer.test.ts
+++ b/packages/core/src/lib/stdin-buffer.test.ts
@@ -894,4 +894,124 @@ describe("StdinBuffer", () => {
       expect(emittedSequences).toEqual(["\x1bP>|kitty(0.40.1)\x1b\\"])
     })
   })
+
+  describe("OSC Sequence Handling (#6147)", () => {
+    it("should pass through complete OSC sequences", () => {
+      // Complete OSC with BEL terminator
+      processInput("\x1b]11;rgb:eeee/eeee/eeee\x07")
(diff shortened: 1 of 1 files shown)


kommander:
Mhh, interesting. That means potentially dropping valid input though, like responses to OSC capability queries etc. Maybe increasing the timeout for ssh connections then would help?

gbpdt:
Just to add some support for this change. It seems like the upstream Windows ssh issue might take a long time to be released and filter down to users.

kommander:
Addresed in #770

PR #694, opened 2026-02-16
feat(markdown): render list items as nested containers

remorses:
Split from #597 to make review easier. Depends on #692.

Replaces flat text chunk list rendering with BoxRenderable containers. Each list item gets its own box, with the first inline text on the marker line and remaining child tokens (nested lists, code blocks) as indented children. Supports nested lists, task checkboxes, and code blocks inside list items.


simonklee:
Closing because #1049 already added structural Markdown list rendering, nested
lists, and code blocks.

PR #695, opened 2026-02-16
fix(markdown): only skip incomplete rows in streaming tables

remorses:
Split from #597 to make review easier.

Previously the last table row was always skipped during streaming, even when complete. Now checks if `table.raw.endsWith("\n")` — if yes, the row is complete and shown. Extracts `getTableRowsToRender()` and `getTableRenderableRowCount()` helpers.


kommander:
Obsolete via #732

PR #696, opened 2026-02-16
fix(markdown): rebuild table when streaming ends

mocksoul:
## Summary
- Last table row is permanently missing after streaming completes
- `updateBlocks()` skips table rebuild when token raw hasn't changed, but the table renderable was created with N-1 rows during streaming (`rows.slice(0, -1)`)
- Use `clearCache()` instead of `updateBlocks()` on `streaming: true→false` transition to force full rebuild

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -122,8 +122,8 @@ export class MarkdownRenderable extends Renderable {
   set streaming(value: boolean) {
     if (this._streaming !== value) {
       this._streaming = value
-      // Streaming mode changes table row visibility even when token raw is unchanged.
-      // Force a full rebuild on mode transitions to keep table rendering correct.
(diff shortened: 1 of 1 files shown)


remorses:
It would be better to fix the bug without clearing the full cache to prevent stutters for big markdown content. Can you try telling your agent to try that @mocksoul?

remorses:
Also in opencode streaming is always set to true right now I think.  https://github.com/anomalyco/opencode/blob/bb30e06855fb979b5fd765796a6b7428b9177b91/packages/opencode/src/cli/cmd/tui/routes/session/index.tsx#L1380

simonklee:
See commits, he already did that. But I changed it to the current version. See regressions.

remorses:
Ok I am fine with this. Later we can add a method to only rebuild tables to improve performance (I cannot push to PRs)

PR #706, opened 2026-02-19
fix(markdown): destroy replaced block renderables

simonklee:
`remove()` detaches children from the layout tree but
does not free or clean. You need to destroy recursively.


kommander:
I think I fixed the same thing in #732 ?

PR #707, opened 2026-02-19
fix(markdown): skip redundant href in conceal mode for autolinks

mocksoul:
When conceal mode renders a link where the label text equals the href (e.g. bare URLs and autolinks), the URL was displayed twice:

```
https://example.com (https://example.com)
```

This adds a check to skip the `(href)` suffix when the label matches the URL. Named links like `[docs](https://example.com)` still render as `docs (https://example.com)`.


kommander:
Is this still relevant with recent refactors?

simonklee:
> Is this still relevant with recent refactors?

No, we've already updated the test. It's:

```
test("incomplete link (no closing paren)", async () => {
  const markdown = `Check out [this link](https://example.com`

  expect(await renderMarkdown(markdown)).toMatchInlineSnapshot(`
    "
    Check out this link(https://example.com"
  `)
})

```

in `main`

PR #708, opened 2026-02-19
fix(markdown): render checkbox inline with list item text

mocksoul:
The marked lexer emits a `checkbox` token as the first child of a task list item. The renderer did not handle this token type, so it fell through to `renderTokenToChunks()` which rendered it as raw text on a separate line:

```
- [ ]
Task text here
```

Now the checkbox token is consumed first and rendered inline before the item text:

```
- [ ] Task text here
```

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -352,12 +352,21 @@ export class MarkdownRenderable extends Renderable {
         chunks.push(this.createChunk("- ", "markup.list"))
       }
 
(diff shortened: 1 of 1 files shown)


remorses:
There is already https://github.com/anomalyco/opentui/pull/694 to fix this issue

mocksoul:
You're right, sorry about that — I checked the open issues but missed your PR. Feel free to close this once #694 lands. In the meantime this works as a smaller standalone fix for anyone who needs it sooner.

kommander:
The markdown renderable was refactored, so this no longer applies.

PR #531, opened 2026-01-15
fix(scrollbox): reset _hasManualScroll when user scrolls back to sticky position

bradleat:
## Summary
- Fixes #530 - ScrollBox with `stickyScroll={true}` and `stickyStart="bottom"` randomly scrolls to top after user input
- Resets `_hasManualScroll = false` in `updateStickyState()` when user scrolls back to their sticky position
- Handles edge case where content fits in viewport (`maxScrollTop === 0`)

## Test plan
- [x] All 2941 existing tests pass
- [x] Build succeeds
- [x] Verified locally: terminal emulator stays pinned to bottom after scrolling up then back down

PR #710, opened 2026-02-19
fix(scrollbox): preserve manual scroll state during content size recalculations

### LINKED ISSUE #709: ScrollBox: content size changes reset _hasManualScroll, causing auto-scroll during streaming
When a user scrolls up while content is being streamed (e.g. in opencode TUI), the viewport jumps back toward the bottom. This happens because `recalculateBarProps` (triggered by `content.onSizeChange`) updates `scrollBar.scrollSize`, which can trigger `updateStickyState` → reset `_hasManualScroll = false` — even though the user never scrolled back to the sticky edge.

Root cause: PR #531 added `_hasManualScroll` resets in `updateStickyState`, but didn't guard them with `_isApplyingStickyScroll`. So programmatic size recalculations can falsely reset the flag.

Related opencode issues: https://github.com/anomalyco/opencode/issues/7648, https://github.com/anomalyco/opencode/issues/7659

mocksoul:
## Summary
- Guard `_hasManualScroll = false` in `updateStickyState` with `!this._isApplyingStickyScroll`
- This preserves the user's scroll position when content grows (streaming), while still allowing reset when the user actually scrolls back to the sticky edge

Regression from #531. Fixes #709.

### DIFF
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -167,13 +167,18 @@ export class ScrollBoxRenderable extends BoxRenderable {
     if (this.scrollTop <= 0) {
       this._stickyScrollTop = true
       this._stickyScrollBottom = false
-      if (this._stickyStart === "top" || (this._stickyStart === "bottom" && maxScrollTop === 0)) {
+      // Only reset manual scroll for user-initiated scrolls, not programmatic
+      // recalculations (e.g. content size changes during streaming)
+      if (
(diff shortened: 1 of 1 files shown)


mocksoul:
Closing in favor of #722 which was merged and fixes the same issue with essentially the same approach (plus nicer try/finally guards). Verified it works correctly. Thanks @simonklee!

simonklee:
I didn't realize you had a PR on this already. I could have continued on your branch. Sorry.

PR #712, opened 2026-02-19
fix(markdown): word-wrap table cells instead of truncating content

### LINKED ISSUE #711: Markdown table cells truncate content instead of word-wrapping
## Description

Markdown table cells use `height: 1` and `overflow: "hidden"`, which truncates any content that doesn't fit in a single line. Long cell content is silently cut off instead of word-wrapping.

Silently hiding data from the user is about the worst thing a UI can do — the user has no way to know that content was cut off, and may act on incomplete information.

## Steps to Reproduce

Render a markdown table with cells containing text wider than the column width:

```markdown
| Column A | Column B | Column C |
|----------|----------|----------|
| This is a very long cell content that should wrap | Another long text here | Short |
```

## Expected Behavior

Cell content should word-wrap within the column, and the row height should expand to accommodate the wrapped text.

## Actual Behavior

Content is truncated to a single line. Anything beyond the column width is hidden.

## Screenshots

**Before (content truncated):**

Image

**After (content word-wraps):**

Image

## Root Cause

In `createTableRenderable()`, each cell's `TextRenderable` is created with `height: 1` and `overflow: "hidden"`. The table uses a column-major layout (each column is a separate flex-column container), which makes it impossible to synchronize row heights across columns when cells have different content lengths.

## Proposed Fix

Rewrite the table layout from column-major to row-major:
- Each row becomes a flex-row container with cells as children
- Yoga's `alignItems: stretch` (default) automatically synchronizes cell heights within a row
- Remove `height: 1` and `overflow: "hidden"` — let `measureFunc` auto-calculate height based on wrapped content
- Use `flexBasis` based on natural column widths for compact tables, `flexShrink: 1` for proportional shrinking when the table exceeds viewport width

This also fixes row separators: `│───│` → `├───┼───┤` (proper T-junction characters).

mocksoul:
Fixes #711

## Summary

- Rewrites table layout from column-major to row-major so cells word-wrap instead of silently truncating content
- Fixes row separator characters: `│───│` → `├───┼───┤` (proper T-junctions)

## Problem

Table cells were created with `height: 1` and `overflow: "hidden"`, silently cutting off any content that didn't fit in a single line. The column-major layout (each column as a separate flex-column) made it impossible to synchronize row heights across columns.

## Solution

**Row-major layout:** each row is a flex-row container. Yoga's default `alignItems: stretch` automatically synchronizes cell heights within a row.

- No explicit `height` or `overflow` on `TextRenderable` — `measureFunc` auto-calculates height from wrapped content
- `flexBasis` set to natural column width (max content + padding + borders) for compact tables
- `flexShrink: 1` enables proportional shrinking with word-wrap when the table exceeds viewport width
- `customBorderChars` per cell position for correct T-junction/cross characters

## Testing

- All 84 existing tests pass (25 snapshot updates for improved separator chars)
- 2 new tests for word-wrap behavior

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -430,70 +430,55 @@ export class MarkdownRenderable extends Renderable {
 
     const rowsToRender = this._streaming && table.rows.length > 0 ? table.rows.slice(0, -1) : table.rows
     const colCount = table.header.length
+    const totalRows = 1 + rowsToRender.length // header + data rows
 
-    // Traverse existing table structure: tableBox -> columnBoxes -> cells
-    const columns = (tableBox as any)._childrenInLayoutOrder as Renderable[]
-    for (let col = 0; col < colCount; col++) {
-      const columnBox = columns[col]
-      if (!columnBox) continue
+    // Traverse row-major structure: tableBox -> rows -> cells
+    const rows = (tableBox as any)._childrenInLayoutOrder as Renderable[]
+    for (let row = 0; row < Math.min(totalRows, rows.length); row++) {
+      const isHeaderRow = row === 0
+      const dataRowIndex = row - 1
+      const rowBox = rows[row]
+      const cells = (rowBox as any)._childrenInLayoutOrder as Renderable[]
 
-      // Update column border colors
-      if (columnBox instanceof BoxRenderable) {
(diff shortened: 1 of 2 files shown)


kommander:
Should be fixed via #732

mocksoul:
> Should be fixed via #732

#732 fixes #711 too?

PR #715, opened 2026-02-19
feat: Support multiple TUI renderers in a single process

jettptacek:
This PR is meant to be the start of the conversation of allowing opentui to render multiple TUIs. I have allowed multiple independent `CliRenderer` instances to coexist within a single process. Previously, the Zig renderer used static/global output buffers and always wrote to `stdout`, meaning only one renderer could exist at a time. The timeline engine also only tracked a single renderer.

I understand this is a large change..... that cuts across the native renderer, the TypeScript layer, and the Solid integration and I'm expecting that others might want a different implementation

I am using this right now in ssh server where I can make a different tui per client connected.

## Changes

**Zig renderer (`renderer.zig`)**

- Output buffers moved from static `var` declarations to per-instance heap-allocated fields
- `OutputBufferWriter` now holds a `*CliRenderer` pointer instead of referencing global state
- All hardcoded `std.fs.File.stdout()` calls replaced with a per-instance `output_file` field (defaults to stdout)
- New `setOutputFd` / `fileFromFd` to redirect a renderer's output to an arbitrary file descriptor, with cross-platform handle type support

**FFI bindings (`lib.zig`, `zig.ts`)**

- Exposed `setOutputFd` through the FFI layer

**TypeScript renderer (`renderer.ts`)**

- New `remote` config flag: when true, the renderer skips process-level signal handlers (`SIGWINCH`, `uncaughtException`, `beforeExit`), stdout interception, exit listeners, and `global.requestAnimationFrame` overwriting
- Passes `config.stdout` fd to the native renderer via `setOutputFd`, with Windows `_get_osfhandle` conversion through `msvcrt.dll`
- `processResize` changed from `private` to `public` so callers can trigger resizes on specific renderer instances

**Timeline engine (`Timeline.ts`)**

- `TimelineEngine` changed from tracking a single renderer to a `Map<CliRenderer, AttachedRenderer>` where each renderer gets its own frame callback and live state
- `detach()` can target a specific renderer or detach all

**Solid integration (`index.ts`, `reconciler.ts`, `hooks.ts`, `renderer-stack.ts`)**

- New `renderer-stack.ts` module provides renderer lookup via a synchronous stack (during initial mount) and a `WeakMap` keyed by Solid owner (for effect re-runs)
- `useRenderer()` falls back to `getActiveRenderer()` when `RendererContext` is not available
- `render()` and `testRender()` now call `engine.detach(renderer)` on destroy
- `render()` return type changed to `Promise<CliRenderer>`

**Test renderer (`test-renderer.ts`)**

- Removed `@ts-expect-error` on `processResize` call (now public)

## Tests

- New TypeScript tests (`renderer.multi-instance.test.ts`): independent coexistence, independent resize, independent render buffers, destroy isolation, remote signal handler behavior, timeline engine multi-attach/detach
- New Zig tests (`renderer_test.zig`): `setOutputFd` output redirect via pipe, two renderers writing to separate fds

### DIFF
--- a/packages/solid/src/renderer-stack.ts
+++ b/packages/solid/src/renderer-stack.ts
@@ -3,7 +3,6 @@
 import type { CliRenderer } from "@opentui/core"
 import { getOwner } from "solid-js"
 
-
 const rendererStack: CliRenderer[] = []
 
 export function pushActiveRenderer(renderer: CliRenderer): void {
@@ -37,7 +36,6 @@ function findRendererFromOwner(): CliRenderer | undefined {
   }
   return undefined
 }
-
 
 export function getActiveRenderer(): CliRenderer | undefined {
   const stackRenderer = rendererStack[rendererStack.length - 1]


kommander:
Cool, maybe we could split that so #636 isn't blocked/conflicting and we can introduce things independently? The two parts I see:

- getting rid of globals etc. which would allow running multiple renderers in the same process, with #636 will be able to write out to different write streams via js (NativeSpanFeed)

- introducing custom handles for the native write out, which will allow what #636 does basically, but native.

wdyt?

jettptacek:
Yes I agree. Also, It might be a couple of days before I get those prs made

PR #717, opened 2026-02-20
Fix markdown CodeRenderable  illegible in light theme

NicoDeVeaux:
I noticed that when in light mode, CodeRenderable text was rendered white on white: it was illegible.

This updates markdown.ts to:

* derive code-block fallback colors from markup.raw.block / markup.raw / default
* pass those colors into internal CodeRenderable creation

Includes simple unit tests


packages/core/src/renderables/Markdown.ts
```diff
@@ -406,13 +406,22 @@ export class MarkdownRenderable extends Renderable {
     })
   }
 
+  private getCodeFallbackStyle(): StyleDefinition | undefined {
+    return this.getStyle("markup.raw.block") || this.getStyle("markup.raw") || this.getStyle("default")
```
kommander, line 416: This is language/parser/query specific and not generic enough. There is the `default` style, isn't that enough?

NicoDeVeaux:
Closing, since PR #832 by simonklee already addresses the same bug, but differently: it added fg/bg props to MarkdownRenderable

PR #727, opened 2026-02-22
fix(stdin-buffer): handle split escape sequences

yukukotani:
## Summary

- Extend flush timeout for incomplete escape sequences so split focus sequences (e.g. `ESC[I` arriving as `ESC` then `[I`, or `ESC[` then `I`) reassemble naturally
- Uses existing `isCompleteSequence()` logic — no new fields or state management needed

## Problem

Ghostty (and potentially other terminals) can deliver focus event sequences (`ESC[I` / `ESC[O`) split across multiple chunks. The `StdinBuffer` timeout (5ms) would flush the incomplete sequence before the rest arrived, causing fragments like `[I` to leak into the input as literal text.

## Solution

When the timeout fires and the buffer contains an incomplete escape sequence (as determined by the existing `isCompleteSequence()`), extend the timeout once (50ms) to give the next chunk time to arrive and reassemble through the normal CSI completion logic.

## Changes

1. **`packages/core/src/lib/stdin-buffer.ts`** — 1-line condition change in the timeout handler: `isCompleteSequence(this.buffer) === "incomplete"` triggers a single timeout extension
2. **`packages/core/src/lib/stdin-buffer.test.ts`** — Added 14 test cases for split focus sequences; updated 2 existing tests for new timeout behavior; added buffer cleanup in `beforeEach`

## Testing

All existing tests pass:
- `stdin-buffer.test.ts` (123 tests)
- `renderer.input.test.ts` (60 tests)
- `renderer.focus-restore.test.ts` (9 tests)
- `parse.keypress.test.ts` (41 tests)

### DIFF
--- a/packages/core/src/lib/stdin-buffer.ts
+++ b/packages/core/src/lib/stdin-buffer.ts
@@ -379,11 +379,11 @@ export class StdinBuffer extends EventEmitter<StdinBufferEventMap> {
 
     if (this.buffer.length > 0) {
       this.timeout = setTimeout(() => {
-        // If the buffer is a lone ESC, extend the timeout once so that a
-        // split focus sequence (e.g. ESC arriving separately from [I) can
-        // be reassembled through the normal CSI completion logic when the
+        // If the buffer holds an incomplete escape sequence, extend the
+        // timeout once so that split sequences (e.g. ESC then [I) can be
+        // reassembled through the normal CSI completion logic when the
         // next chunk calls process().
-        if (this.buffer === ESC) {
+        if (isCompleteSequence(this.buffer) === "incomplete") {
           this.timeout = setTimeout(() => {
             const flushed = this.flush()
             for (const sequence of flushed) {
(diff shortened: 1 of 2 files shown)


packages/core/src/lib/stdin-buffer.ts
```diff
@@ -379,8 +379,21 @@ export class StdinBuffer extends EventEmitter<StdinBufferEventMap> {
 
     if (this.buffer.length > 0) {
       this.timeout = setTimeout(() => {
-        const flushed = this.flush()
+        // If the buffer holds an incomplete escape sequence, extend the
+        // timeout once so that split sequences (e.g. ESC then [I) can be
+        // reassembled through the normal CSI completion logic when the
+        // next chunk calls process().
+        if (isCompleteSequence(this.buffer) === "incomplete") {
+          this.timeout = setTimeout(() => {
```
kommander, line 387: I see a risk here of multiple timeouts being active. Starting another timeout here seems odd. The whole point of introducing the stdin-buffer was exactly this case iirc. There must be a better way of handling this, without another arbitrary timeout of 50ms.

kommander:
#770 should address this.

PR #734, opened 2026-02-24
text-buffer-view: lineInfo byte-start tests

simonklee:
Pin UTF-8 boundary, monotonicity, and determinism invariants for lineStarts across CJK, emoji, tab, and newline inputs. A fixed-seed corpus test catches regressions from future
  wrap-offset changes.
  
These tests are expected to fail for now.


simonklee:
Superseded by #747

PR #568, opened 2026-01-21
feat(core): Add highlighting callback for custom inline highlighting

jettptacek:
Adds an `onHighlight` callback to `CodeRenderable` that allows post-processing of syntax highlights before they're rendered. This enables custom inline highlighting
 ### Changes
- Added `onHighlight` option to `CodeRenderable` that receives tree-sitter highlights and can modify them
- Added `HighlightContext` interface providing access to content, filetype, and syntaxStyle

 ### Usage
```typescript

const codeDisplay = new CodeRenderable(renderer, {
  content: code,
  filetype: "typescript",
  syntaxStyle,
  onHighlight: (highlights, ctx) => {
    highlights.push([startPos, endPos, "(registered style) i.e color.FF0000", {}])
    return highlights
  },
})
```
### Solves
I have an open pr in opencode https://github.com/anomalyco/opencode/pull/9016 to close https://github.com/anomalyco/opencode/issues/2586 but my pr monkey patches `treeSitterClient.highlightOnce` to get that feature working and this onHighlight callback change would allow a cleaner implementation of https://github.com/anomalyco/opencode/issues/2586 and/or other custom highlight features

### Examples
I made 2 examples using this feature that I can add to the pull request as well. If it's wanted.

https://github.com/user-attachments/assets/6f41d59c-a5ea-4e9b-b398-43bc4b9914e8

https://github.com/user-attachments/assets/f0aeb59a-5b19-4fd2-948e-e1a2e80d7b1f

PR #736, opened 2026-02-24
fix(renderer): add OSC 8 hyperlink id parameter and onChunks hook for link detection

### LINKED ISSUE #735: Hyperlinks spanning wrapped lines are not grouped (missing 'id' parameter in OSC 8)
## Bug Description

When a hyperlink wraps across multiple terminal lines, Cmd/Ctrl+Click only selects the portion on the clicked line instead of the full URL. This happens because the OSC 8 hyperlink escape sequences emitted by the renderer are missing the `id` parameter, so the terminal treats the link segments on each row as separate hyperlinks rather than one continuous link.

## Video Demonstration

https://github.com/user-attachments/assets/697edfef-6e11-4147-93fa-f7ae1d6f3441

## Root Cause

In `packages/core/src/zig/renderer.zig` at line 670, the OSC 8 open sequence does not include an `id` parameter:

```zig
// Current (line 670):
writer.print("\x1b]8;;{s}\x1b\\", .{url_bytes}) catch {};
```

Per the [OSC 8 hyperlink spec](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda), when a hyperlink is split across multiple rows (e.g., link text ends at end of row N, padding spaces close it, then it reopens on row N+1), the terminal needs the `id` parameter to recognize that both segments belong to the same logical hyperlink.

## Proposed Fix

Add the link ID to the OSC 8 open sequence:

```zig
// Fixed:
writer.print("\x1b]8;id={d};{s}\x1b\\", .{currentLinkId, url_bytes}) catch {};
```

`currentLinkId` is already tracked in the renderer's state and is the correct identifier to use here.

## Rendering Flow (for context)

1. **JS layer**: `textBufferSetStyledText()` receives chunks with optional `link: { url }` property
2. **Link encoding**: `linkAlloc()` allocates URL in Zig `LinkPool`, returns numeric ID; ID is packed into the upper 24 bits of `attributes` u32
3. **Style registration**: `setStyledText()` in `text-buffer.zig` calls `registerStyle()` with full attributes (including link_id bits)
4. **Word wrapping**: `text-buffer-view.zig` `calculateVirtualLinesGeneric()` — wrapping is link-unaware, splits purely on geometry
5. **Cell rendering**: `buffer.zig` `drawTextBufferInternal()` resolves style spans per-cell; link_id flows through `resolved_style.attributes` into cell attributes
6. **Terminal output**: `renderer.zig` `prepareRenderFrame()` emits OSC 8 sequences — **`currentLinkId` persists across rows** (never reset between rows), but the missing `id` param prevents the terminal from grouping the segments

## Test Coverage

Existing hyperlink tests in `packages/core/src/zig/tests/renderer_test.zig` (lines 621-760) cover basic hyperlink rendering but do **not** test multi-line wrapped hyperlinks. A new test should verify that wrapped hyperlinks emit matching `id` parameters.

mugnimaestra:
## Summary

Fixes hyperlinks not working correctly when they wrap across terminal lines, and adds opt-in link detection for tree-sitter rendered code blocks.

Closes #735

### 1. `fix(renderer): add id parameter to OSC 8 hyperlink sequences`

The renderer emitted OSC 8 without the `id` parameter, so terminals could not identify wrapped link segments as the same hyperlink. Now emits:

```
\x1b]8;id={linkId};{url}\x1b\\
```

Links stay open across row boundaries and close naturally when a different link (or no link) is encountered.

### 2. `feat(code): add onChunks hook and detectLinks helper`

The tree-sitter path (`CodeRenderable`) never set `link` on `TextChunk` objects, so OSC 8 was never emitted for code blocks.

Following [maintainer feedback](https://github.com/anomalyco/opentui/pull/736#issuecomment-3955277832), this uses hooks instead of hardcoded scope names:

- **`onChunks`** callback on `CodeRenderable` — runs after `treeSitterToTextChunks`, before `setStyledText`
- **`detectLinks()`** exported helper — scans highlights for URL scopes, sets `chunk.link = { url }` on matches

Uses `content.indexOf()` to find chunk positions in original content, handling concealed markdown syntax where visible text length differs from source byte positions.

```tsx
import { detectLinks } from "@opentui/core"

<code filetype="markdown" onChunks={detectLinks}>
  {content}
</code>
```

## Tests

- 3334 TS tests pass, 1480 native tests pass
- Added renderer test for wrapped hyperlinks
- Added 5 unit tests for `detectLinks` including concealed text edge case

## Known Limitation

Visual hover highlighting covers one row at a time for wrapped links. The [OSC 8 spec](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) defines `id`-based cross-row grouping, but no terminal has been verified to implement it. Cmd/Ctrl+Click works correctly across all rows.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -10,7 +10,7 @@
     },
     "packages/core": {
       "name": "@opentui/core",
-      "version": "0.1.82",
+      "version": "0.1.84",
       "dependencies": {
         "bun-ffi-structs": "0.1.2",
         "diff": "8.0.2",
@@ -28,12 +28,12 @@
       },
       "optionalDependencies": {
         "@dimforge/rapier2d-simd-compat": "^0.17.3",
-        "@opentui/core-darwin-arm64": "0.1.82",
-        "@opentui/core-darwin-x64": "0.1.82",
-        "@opentui/core-linux-arm64": "0.1.82",
-        "@opentui/core-linux-x64": "0.1.82",
-        "@opentui/core-win32-arm64": "0.1.82",
-        "@opentui/core-win32-x64": "0.1.82",
+        "@opentui/core-darwin-arm64": "0.1.84",
+        "@opentui/core-darwin-x64": "0.1.84",
+        "@opentui/core-linux-arm64": "0.1.84",
+        "@opentui/core-linux-x64": "0.1.84",
+        "@opentui/core-win32-arm64": "0.1.84",
+        "@opentui/core-win32-x64": "0.1.84",
         "bun-webgpu": "0.1.5",
         "planck": "^1.4.2",
         "three": "0.177.0",
@@ -44,7 +44,7 @@
     },
     "packages/react": {
       "name": "@opentui/react",
-      "version": "0.1.82",
+      "version": "0.1.84",
       "dependencies": {
         "@opentui/core": "workspace:*",
         "react-reconciler": "^0.32.0",
@@ -72,7 +72,7 @@
     },
     "packages/solid": {
       "name": "@opentui/solid",
-      "version": "0.1.82",
+      "version": "0.1.84",
       "dependencies": {
         "@babel/core": "7.28.0",
         "@babel/preset-typescript": "7.27.1",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -9,7 +9,7 @@
   "types": "src/index.ts",
   "module": "src/index.ts",
   "main": "src/index.ts",
-  "version": "0.1.82",
+  "version": "0.1.84",
   "type": "module",
   "scripts": {
     "build": "bun run build:native && bun run build:lib",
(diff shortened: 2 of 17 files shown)


kommander:
@mugnimaestra the reason why I didn't do the "link detection" in `tree-sitter-styled-text` was that `markup.link.url` is language specific and may be different based on the query that is used, so it's not generic. That should be done via hooks for the `CodeRenderable`. There were hooks introduced in #568 in the `CodeRenderable`, maybe these can be used? I am fine with opentui exporting a method that does link detection and chunk editing via such a hook.

mugnimaestra:
Thanks for the feedback @kommander! I've refactored the approach:

1. **Reverted** the hardcoded `markup.link.url` / `string.special.url` detection from `treeSitterToTextChunks` — you're right that those scope names are language-specific and don't belong in the generic pipeline.

2. **Added `onChunks` callback** to `CodeRenderable` — this runs after `treeSitterToTextChunks` produces chunks, allowing consumers to post-process `TextChunk[]` (e.g., add `link` properties). It follows the same pattern as the existing `onHighlight` hook from #568.

3. **Exported `detectLinks()` helper** from `@opentui/core` — consumers pass this as `onChunks` to opt-in to link detection. It detects `markup.link.url` and `string.special.url` scopes from the highlight data and enriches the corresponding chunks with `link: { url }`.

Usage in opencode:
```tsx
import { detectLinks } from "@opentui/core"

<code filetype="markdown" onChunks={detectLinks} ... />
```

Tested in iTerm2 — OSC 8 hyperlinks now work correctly, including URLs that wrap across multiple terminal lines (thanks to the `id` parameter fix in renderer.zig + the link detection via this hook).

kommander:
Cool, I'll re-check shortly.

mugnimaestra:
https://github.com/user-attachments/assets/f8deffc0-2c45-4aba-b186-39200f9551cc

I couldn't make the UI highlighting works, it seems the issues is quite complex so I refrain to do any further, but at least when cmd+click url that gets separated by newline now it still captures the full URL as expected

PR #737, opened 2026-02-24
feat(tree-sitter): detect URL scopes and set link property on text chunks

mugnimaestra:
## Summary

When tree-sitter highlights markdown content via the `<code filetype="markdown">` rendering path, URL-bearing text chunks never receive a `link` property, so OSC 8 hyperlink sequences are never emitted. This causes terminals to fall back on visual URL auto-detection, which breaks when URLs wrap across lines.

## Changes

**`packages/core/src/lib/tree-sitter-styled-text.ts`**

Added URL detection to `treeSitterToTextChunks()`:

1. **Pre-scan phase** (lines 51-68): Builds a `Map<number, string>` mapping highlight indices to URLs. Scans for `markup.link.url` and `string.special.url` scopes, extracts the URL text, and also maps associated `markup.link.label` highlights (found by walking backwards past `markup.link` bracket highlights) to the same URL.

2. **Chunk creation** (lines 204-227): After merging styles, checks if any active highlight has a URL mapping. If so, sets `link: { url }` on the chunk.

This enables:
- URL text (`https://example.com`) → clickable OSC 8 hyperlink
- Link labels (`Click here` in `[Click here](url)`) → also clickable, linking to the same URL
- URLs in code comments tagged with `string.special.url` → also clickable

**`packages/core/src/lib/tree-sitter-styled-text.test.ts`**

Added 5 new tests in a `"Link detection"` describe block:
- `markup.link.url` chunks get `link` property ✅
- `markup.link.label` chunks get associated URL ✅
- `string.special.url` chunks get `link` property ✅
- Non-URL chunks don't get `link` ✅
- Integration test with real tree-sitter parser ✅

## Test Results

All 49 tests pass (44 existing + 5 new).

## Related

- Fixes the default rendering path issue described in anomalyco/opentui#735
- Complements #736 (which fixes OSC 8 id parameter for the experimental markdown path)
- Related opencode issue: sst/opencode#(pending)

### DIFF
--- a/packages/core/src/lib/tree-sitter-styled-text.test.ts
+++ b/packages/core/src/lib/tree-sitter-styled-text.test.ts
@@ -1250,4 +1250,124 @@ Normal paragraph with [link](https://example.com).`
       expect(chunk.attributes).toBe(expectedAttributes)
     })
   })
+
+  describe("Link detection", () => {
+    test("should set link property on markup.link.url chunks", () => {
+      const content = "[Click here](https://example.com)"
+      const mockHighlights: SimpleHighlight[] = [
+        [0, 1, "markup.link"], // [
+        [1, 11, "markup.link.label"], // Click here
+        [11, 13, "markup.link"], // ](
+        [13, 32, "markup.link.url"], // https://example.com
+        [32, 33, "markup.link"], // )
+      ]
+
+      const style = SyntaxStyle.fromStyles({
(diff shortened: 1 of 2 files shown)


mugnimaestra:
Combining this with #736 — both fixes are needed together to solve the same issue (URLs breaking at line wrap).

PR #739, opened 2026-02-24
fix crash on mouse hover of a destroyed renderable

### LINKED ISSUE #738: Mouse hoverring over renderable that gets removed can cause a crash
I have this basic Link component:
```typescript
import { useTheme } from '../../theme/index.ts';
import { ICONS } from '../../utils/icons.ts';
import { openInBrowser } from '../../utils/shell.ts';

type Props = {
	text: string;
	url: string;
};

export function Link(props: Props) {
	const theme = useTheme();
	return (
		<text
			onMouseUp={() => openInBrowser(props.url)}
			onMouseDown={function () {
				this.fg = theme.text;
			}}
			onMouseOver={function () {
				this.fg = theme.accent;
			}}
			onMouseOut={function () {
				this.fg = theme.text;
			}}
			fg={theme.text}
			wrapMode='none'
		>
			<span>{props.text}</span>
			<span style={{ fg: theme.accent }}>{ICONS.urlHint}</span>
		</text>
	);
}
```

It seemingly works fine, except when I use the Link component within a conditional Show component. If the condition flips and the Link component is no longer rendered *while the mouse is still hovering over it*, we crash with:
```
TextBuffer is destroyed
```

justfortheloveof:
Fixes #738 

Tested in my TUI, see the line with `// HACK:` where I added:
```typescript
if (this.isDestroyed) return;
```
Without this guard, we crash in the scenario described in #738 . With the guard, we don't.

```typescript
import { useTheme } from '../../theme/index.ts';
import { ICONS } from '../../utils/icons.ts';
import { openInBrowser } from '../../utils/shell.ts';

type Props = {
	text: string;
	url: string;
};

export function Link(props: Props) {
	const theme = useTheme();
	return (
		<text
			onMouseUp={() => openInBrowser(props.url)}
			onMouseDown={function () {
				this.fg = theme.text;
			}}
			onMouseOver={function () {
				this.fg = theme.accent;
			}}
			onMouseOut={function () {
				if (this.isDestroyed) return; // HACK:
				this.fg = theme.text;
			}}
			fg={theme.text}
			wrapMode='none'
		>
			<span>{props.text}</span>
			<span style={{ fg: theme.accent }}>{ICONS.urlHint}</span>
		</text>
	);
}
```

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -1311,7 +1311,11 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     }
 
     if (!sameElement && (mouseEvent.type === "drag" || mouseEvent.type === "move")) {
-      if (this.lastOverRenderable && this.lastOverRenderable !== this.capturedRenderable) {
+      if (
+        this.lastOverRenderable &&
+        this.lastOverRenderable !== this.capturedRenderable &&
+        !this.lastOverRenderable.isDestroyed
+      ) {
         const event = new MouseEvent(this.lastOverRenderable, { ...mouseEvent, type: "out" })
         this.lastOverRenderable.processMouseEvent(event)
       }
--- a/packages/core/src/tests/renderer.mouse.test.ts
+++ b/packages/core/src/tests/renderer.mouse.test.ts
@@ -1162,6 +1162,42 @@ describe("renderer handleMouseData", () => {
       renderer.destroy()
     }
   })
+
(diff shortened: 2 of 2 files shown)


kommander:
Could you generate a regression test that covers that high level, in `src/tests/renderer.mouse.test.ts` for example?

justfortheloveof:
Sure! I will add this, hopefully before the weekend.

simonklee (approved):

Issue #563, opened 2026-01-21
bug: <diff mode={"split"} > scrolling is not synced


safareli:
used the example of diff component [packages/react/examples/diff.tsx ](https://github.com/anomalyco/opentui/blob/9ad8525c4ef1878029105e18126998051ab98c56/packages/react/examples/diff.tsx) and when diff is and as a result mouse scroll is kicked in, i can scroll added lines and removed lines independently. one would expect the scroll to move full diff.

https://github.com/user-attachments/assets/46fb2139-d18e-4059-934f-190522e063f9

kommander:
That was intentional. Synced scroll could be added though.

viralcodex:
Hi @kommander, if not implemented, I can take it up, also we can have both behaviours, synced and separate as a setting flag?

PR #742, opened 2026-02-25
feat(DiffRenderable): synchronize scroll between split view panes in diff.ts

viralcodex:
Fixes: #563
Hi @kommander and @simonklee,

Added this synced scroll in code diff as mentioned in this issue...

I was thinking to keep both separate and synced scroll which can be toggled through hotkeys depending on user preferences.

Whatever you guys say 😄 

https://github.com/user-attachments/assets/bdc27805-d5e4-4f1c-b1f4-d31b94da4601


simonklee:
Thanks for taking a stab at this. If i understand the impl right you changed the behavior to always be "synced"?

viralcodex:
@simonklee , yes I have changed it to always sync, um, like everywhere else we see. If you want I can have a hotkey based toggle between both separate/synced behaviours too.

simonklee:
> @simonklee , yes I have changed it to always sync, um, like everywhere else we see. If you want I can have a hotkey based toggle between both separate/synced behaviours too.

I understand, but this is effectively just changing the behavior of a design decision made by @kommander. See his comment in the original issue:

> That was intentional. Synced scroll could be added though.

So, to land this PR, that should be addressed.

viralcodex:
Yes, I read it, but for starters I did the change.
@kommander, what would you suggest? A toggle between both states, or anything else. Will make the change accordingly.

simonklee:
Yeh, something like that. An option on the DiffRenderable that defaults to the current behavior.

viralcodex:
> Yeh, something like that. An option on the DiffRenderable that defaults to the current behaviour.

Hey @simonklee , made the changes, added the option in the API as syncScroll, false by default to keep the current behaviour. can be set to true to make it synced.

PR #749, opened 2026-02-27
fix(core): prevent input starvation from high-frequency mouse scroll events

### LINKED ISSUE #748: High-frequency scroll events cause input starvation and dropped keyboard data
## Bug

When using a mouse with a free-spinning scroll wheel (e.g. Logitech MX Master 3 with ratchet unlocked), scrolling in any opentui-based app can freeze keyboard input entirely. On Ghostty, raw SGR escape sequences like `\e[<65;93;34M` also appear as visible text.

## Root cause

Two issues in `packages/core/src/renderer.ts`:

### 1. `stdinListener` drops non-mouse data from mixed chunks

```ts
private stdinListener = ((data: Buffer) => {
        return  // ← entire buffer consumed for mouse events; remainder is lost
    }
    this._stdinBuffer.process(data)
})
```

When `parseAllMouseEvents` encounters non-mouse data (keyboard input, focus-in/out sequences, etc.) after mouse events in the same stdin chunk, it stops parsing. But `handleMouseData` returns `true` and `stdinListener` returns early — the non-mouse remainder is **silently dropped**.

This means keyboard input, focus-in (`\e[I`), and other control sequences that arrive in the same OS read as mouse data are permanently lost.

### 2. No scroll event coalescing

```ts
if (mouseEvent.type === "scroll") {
    const maybeRenderableId = this.hitTest(mouseEvent.x, mouseEvent.y)
    // ... immediately dispatch to renderable
}
```

Every scroll event is dispatched individually with no throttling or batching. Free-spinning wheels generate hundreds of events per second. Each event triggers hit testing → MouseEvent construction → `processMouseEvent` → Solid reactivity → layout. This saturates the Node.js event loop, starving keyboard event processing.

## Reproduction

1. Use a Logitech MX Master 3 (or any mouse with a free-spinning scroll wheel)
2. Unlock the scroll wheel ratchet
3. Open any opentui-based app (e.g. OpenCode)
4. Scroll quickly or free-spin the wheel
5. Try typing — keyboard input is delayed or entirely lost
6. On Ghostty: raw SGR sequences appear as visible text in the terminal

## Environment

- opentui v0.1.84 (also confirmed on v0.1.73 bundled with OpenCode)
- macOS, Ghostty and Terminal.app
- Logitech MX Master 3 with free-spin wheel

## Fix

PR forthcoming with:
- `parseAllMouseEvents` returns consumed byte count; `stdinListener` forwards remainder to `StdinBuffer`
- Consecutive same-direction scroll events within a chunk are coalesced into a single event with accumulated `delta`

androolloyd:
## Summary

Fixes #748 — high-frequency scroll wheels (Logitech MX Master 3 free-spin) cause keyboard input loss and event loop starvation.

### 1. Forward remaining non-mouse data from mixed stdin chunks

`stdinListener` previously returned early after `handleMouseData` found any mouse events, silently dropping non-mouse data (keyboard input, focus-in/out sequences) that arrived in the same OS read. Now `handleMouseData` returns the consumed byte count and `stdinListener` forwards the remainder to `StdinBuffer`.

### 2. Coalesce consecutive scroll events within a chunk

Added `coalesceScrollEvents()` that merges adjacent same-direction scroll events at the same coordinates into a single event with accumulated `delta`. This reduces the number of hit tests, MouseEvent constructions, and renderable dispatches from O(n) to O(1) per direction per chunk.

### 3. Return consumed offset from `parseAllMouseEvents`

Changed `parseAllMouseEvents` return type from `RawMouseEvent[]` to `{ events: RawMouseEvent[], consumed: number }` so callers know exactly how many bytes were consumed and can handle the remainder appropriately.

## Testing

- Updated all existing `parseAllMouseEvents` tests for new return type
- Added 5 new tests for consumed byte tracking (trailing focus sequences, keyboard data, mixed buffers)
- Updated `integration.test.ts` for new return type
- All 196 tests pass (mouse parser + stdin buffer + integration)

### DIFF
--- a/packages/core/src/lib/parse.mouse.test.ts
+++ b/packages/core/src/lib/parse.mouse.test.ts
@@ -429,14 +429,14 @@ describe("MouseParser parseAllMouseEvents (multi-event chunks)", () => {
   })
 
   test("single event returns array of one", () => {
-    const events = parser.parseAllMouseEvents(encodeSGR(0, 10, 5, true))
+    const { events } = parser.parseAllMouseEvents(encodeSGR(0, 10, 5, true))
     expect(events).toHaveLength(1)
     expect(events[0]).toMatchObject({ type: "down", button: 0, x: 10, y: 5 })
   })
 
   test("two SGR events concatenated are both parsed", () => {
     const buf = Buffer.concat([encodeSGR(32, 69, 49, true), encodeSGR(32, 68, 49, true)])
-    const events = parser.parseAllMouseEvents(buf)
+    const { events } = parser.parseAllMouseEvents(buf)
     expect(events).toHaveLength(2)
     expect(events[0]).toMatchObject({ type: "move", x: 69, y: 49 })
     expect(events[1]).toMatchObject({ type: "move", x: 68, y: 49 })
@@ -449,7 +449,7 @@ describe("MouseParser parseAllMouseEvents (multi-event chunks)", () => {
       encodeSGR(32, 68, 48, true),
       encodeSGR(32, 67, 48, true),
     ])
-    const events = parser.parseAllMouseEvents(buf)
+    const { events } = parser.parseAllMouseEvents(buf)
     expect(events).toHaveLength(4)
     expect(events[0]).toMatchObject({ x: 69, y: 49 })
     expect(events[1]).toMatchObject({ x: 68, y: 49 })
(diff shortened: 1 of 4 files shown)


simonklee:
Thanks for the attempt at solving this issue. We're going to solve this a slightly different way that unifies the parser's and this pr builds on the current setup. I'm closing this, but keeping your tests for the later fix.

PR #751, opened 2026-02-27
feat(core): expose no_zwj as opt-in WidthMethod

Flare576:
## Problem

ZWJ emoji sequences (👩‍🚀 👨‍👩‍👧 🏳️‍🌈) cause layout corruption in terminals that don't render them as joined glyphs — most notably tmux, but also many older terminals. Text after a ZWJ sequence shifts across the screen, wraps incorrectly, and input box spacing breaks.

The `no_zwj` width calculation mode already exists in the Zig layer, is fully implemented, and has comprehensive test coverage (`utf8_no_zwj_test.zig`). The environment variable `OPENTUI_FORCE_NOZWJ` is already registered and wired through `terminal.zig`. However, `no_zwj` is missing from the TypeScript `WidthMethod` union type and from the numeric mapping at the FFI boundary, making it completely unreachable from JavaScript consumers.

## Solution

Wire `no_zwj` through the full TypeScript→FFI chain without changing any defaults:

- **`packages/core/src/types.ts`**: Add `"no_zwj"` to the `WidthMethod` union
- **`packages/core/src/zig.ts`**: Map `"no_zwj"` → `2` in the four `widthMethodCode` conversions (previously only `0`/`1` existed)
- **`packages/core/src/zig/lib.zig`**: Map `u8` value `2` → `.no_zwj` in the four FFI entry points; map `.no_zwj` → `2` in `getTerminalCapabilities` serialization
- **`packages/core/src/renderer.ts`**: Update the `widthMethod` getter to return `"no_zwj"` when capabilities report `2` (e.g. via `OPENTUI_FORCE_NOZWJ`)

## Behavior

| Mode | ZWJ emoji (👩‍🚀) | Skin tones (👋🏿) | Flags (🇺🇸) | Accents (é) |
|---|---|---|---|---|
| `unicode` (default, unchanged) | 1 glyph, 2 cols | 1 glyph, 2 cols | 1 glyph, 2 cols | 1 glyph, 1 col |
| `no_zwj` (opt-in) | 2 glyphs, 4 cols | 1 glyph, 2 cols | 1 glyph, 2 cols | 1 glyph, 1 col |

Only ZWJ-joined sequences are affected. Skin tone modifiers, regional indicator flags, combining accents, and CJK characters are all unaffected.

## No Breaking Changes

- Default is still `"unicode"` everywhere
- `"wcwidth"` behavior is unchanged  
- `OPENTUI_FORCE_NOZWJ` env var now correctly round-trips through `getTerminalCapabilities` → `widthMethod` getter
- Existing consumers passing `"wcwidth"` or `"unicode"` see identical behavior

## Related

- Related to #609 (emoji wrap corruption at line boundaries)
- Companion PR for OpenCode to expose this as a user config option: https://github.com/Flare576/opencode/tree/feat/no-zwj-width-method

### DIFF
--- a/packages/core/src/zig.ts
+++ b/packages/core/src/zig.ts
@@ -99,6 +99,7 @@ registerEnvVar({
 const CURSOR_STYLE_TO_ID = { block: 0, line: 1, underline: 2, default: 3 } as const
 const CURSOR_ID_TO_STYLE = ["block", "line", "underline", "default"] as const
 const MOUSE_STYLE_TO_ID = { default: 0, pointer: 1, text: 2, crosshair: 3, move: 4, "not-allowed": 5 } as const
+const WIDTH_METHOD_TO_CODE = { wcwidth: 0, unicode: 1, no_zwj: 2 } as const
 
 // Global singleton state for FFI tracing to prevent duplicate exit handlers
 let globalTraceSymbols: Record<string, number[]> | null = null
@@ -2422,7 +2423,7 @@ class FFIRenderLib implements RenderLib {
       console.error(new Error(`Invalid dimensions for OptimizedBuffer: ${width}x${height}`).stack)
     }
 
-    const widthMethodCode = widthMethod === "wcwidth" ? 0 : widthMethod === "no_zwj" ? 2 : 1
+    const widthMethodCode = WIDTH_METHOD_TO_CODE[widthMethod]
     const idToUse = id || "unnamed buffer"
     const idBytes = this.encoder.encode(idToUse)
     const bufferPtr = this.opentui.symbols.createOptimizedBuffer(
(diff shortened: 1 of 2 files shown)


packages/core/src/zig/lib.zig
```diff
@@ -275,7 +275,7 @@ export fn createOptimizedBuffer(width: u32, height: u32, respectAlpha: bool, wid
 
     const pool = gp.initGlobalPool(globalArena);
     const link_pool = link.initGlobalLinkPool(globalArena);
-    const wMethod: utf8.WidthMethod = if (widthMethod == 0) .wcwidth else .unicode;
+    const wMethod: utf8.WidthMethod = if (widthMethod == 0) .wcwidth else if (widthMethod == 2) .no_zwj else .unicode;
```
kommander: Seeing the repetition so clearly I think these should be using `@enumFromInt` and `@enumToInt` utils.

packages/core/src/zig.ts
```diff
@@ -2400,7 +2400,7 @@ class FFIRenderLib implements RenderLib {
       console.error(new Error(`Invalid dimensions for OptimizedBuffer: ${width}x${height}`).stack)
     }
 
-    const widthMethodCode = widthMethod === "wcwidth" ? 0 : 1
+    const widthMethodCode = widthMethod === "wcwidth" ? 0 : widthMethod === "no_zwj" ? 2 : 1
```
kommander: Similar here, seeing the repetition, it would make sense to encapsulate that conversion.

PR #755, opened 2026-02-28
Support 256-color fallback for terminals without truecolor

ryan-z:
## Problem

macOS Terminal.app does not support truecolor (24-bit RGB) by default, limiting colors to the standard 16 or 256 color palette. This makes it hard to see the full range of colors in opentui.

## Solution

This PR adds a 256-color fallback:
- Detects if the terminal supports truecolor via `$COLORTERM` environment variable
- When truecolor is not available, converts RGB colors to the nearest 256-color palette index
- Uses the 6x6x6 color cube (216 colors) and grayscale ramp (24 shades)

## Use case

Running opentui in macOS Terminal.app and seeing more than 16 colors.

### DIFF
--- a/packages/core/src/zig/ansi.zig
+++ b/packages/core/src/zig/ansi.zig
@@ -32,6 +32,15 @@ pub const ANSI = struct {
         writer.print("\x1b[48;2;{d};{d};{d}m", .{ r, g, b }) catch return AnsiError.WriteFailed;
     }
 
+    // 256-color mode output functions (for terminals without truecolor support)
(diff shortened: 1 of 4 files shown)


thrgreenwald:
Also running into this. My OpenTUI app looks terrible on default terminal.app on macOS. Would love for this to get merged!

simonklee:
Pull request #844 already added 256-color fallback for terminals that do not
support truecolor. The current implementation also detects terminal color
support and preserves palette colors.

PR #757, opened 2026-03-01
fix: forward unconsumed data after mouse sequence parsing

amosbird:
## Summary

- `parseAllMouseEvents` now returns a `consumed` byte offset alongside parsed events, telling the caller exactly how many bytes were successfully parsed
- `handleMouseData` in the renderer forwards any unconsumed trailing data to `_stdinBuffer.process()` instead of silently dropping it

## Problem

When a stdin chunk contains complete mouse escape sequences followed by a partial one (e.g. `\x1b[<0;31;16M\x1b[<0;`), `parseAllMouseEvents` parses the complete event but the partial trailing bytes are silently dropped. When the next stdin chunk arrives with the remainder (`31;16M`), it can't be parsed as a mouse event (no `\x1b[` prefix), so it falls through to keyboard input handling — causing characters like `31M;` to appear in the input box.

This is most commonly triggered by rapid mouse movement during terminal focus switches.

## Changes

**`parse.mouse.ts`**: Changed `parseAllMouseEvents` return type from `RawMouseEvent[]` to `{ events: RawMouseEvent[]; consumed: number }`.

**`renderer.ts`**: `handleMouseData` now checks if `consumed < data.length` and forwards the unconsumed remainder to `_stdinBuffer.process()`, which correctly handles partial CSI sequences with its existing buffering/timeout logic.

**`parse.mouse.test.ts`**: Updated all call sites for new return type. Added 5 new tests covering `consumed` offset behavior (full consumption, trailing non-mouse data, partial SGR sequences, empty/non-mouse input).

**`integration.test.ts`**: Updated helper to access `.events` from the new return type.

### DIFF
--- a/packages/core/src/lib/parse.mouse.test.ts
+++ b/packages/core/src/lib/parse.mouse.test.ts
@@ -428,15 +428,18 @@ describe("MouseParser parseAllMouseEvents (multi-event chunks)", () => {
     parser = new MouseParser()
   })
 
+  // Helper: extract just the events array (most existing tests only care about events)
+  const parseAll = (data: Buffer) => parser.parseAllMouseEvents(data).events
+
   test("single event returns array of one", () => {
-    const events = parser.parseAllMouseEvents(encodeSGR(0, 10, 5, true))
+    const events = parseAll(encodeSGR(0, 10, 5, true))
     expect(events).toHaveLength(1)
     expect(events[0]).toMatchObject({ type: "down", button: 0, x: 10, y: 5 })
   })
 
   test("two SGR events concatenated are both parsed", () => {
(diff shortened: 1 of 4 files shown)


simonklee:
Thanks for the PR for this. We're going to solve this a slightly different way that unifies the parser's - stdinListener has two separate parsing paths and I want to avoid that in the first place.

amosbird:
Makes sense, unifying the parsing paths is a cleaner approach. Happy to close this — feel free to ping me if I can help with the unified parser.

simonklee:
> Makes sense, unifying the parsing paths is a cleaner approach. Happy to close this — feel free to ping me if I can help with the unified parser.

Thanks - i haven't really looked at in detail yet. We've had multiple related issues on this problem so I want to solve it generally to avoid a class of issues here.

Issue #705, opened 2026-02-18
`getPallete()` can block until timeout


akronb:
### Summary

`getPalette()` currently behaves as an all-or-timeout operation.  
In environments with partial OSC support, it can frequently resolve only after the full timeout (default 5000ms), which makes downstream apps feel frozen during startup when they wait for palette detection.

### Why this is a problem

For apps using terminal colors for initial theming, startup gets coupled to terminal response timing.  
If OSC responses are partial/incomplete, the app waits for timeout instead of getting early usable data (e.g. first 16 colors) and continuing.

simonklee:
I tried looking at it but not happy where it landed.

MaxDillon:
https://github.com/anomalyco/opencode/issues/7979#issuecomment-3981502176
> stty -echo -icanon min 0 time 5
> printf '\033]10;?\007\033]11;?\007\033]12;?\007\033]13;?\007\033]14;?\007\033]15;?\007\033]16;?\007\033]17;?\007\033]19;?\007'
> dd bs=1 count=512 2>/dev/null | /bin/cat -v                                                                                                                                                                                                                     stty sane
> 
> The root issue is in the interaction between the `getPallette` function in opentui and ghostty. Here are the different responses to this command in ghostty and wezterm:
> 
> ghostty:
> 
> ```
> ^[]10;rgb:cdcd/d6d6/f4f4^G^[]11;rgb:1e1e/1e1e/2e2e^G^[]12;rgb:f5f5/e0e0/dcdc^G%
> ```
> 
> wezterm:
> 
> ```
> ^[]10;rgb:cdcd/d6d6/f4f4^[\^[]11;rgb:1e1e/1e1e/2e2e^[\^[]12;rgb:f5f5/e0e0/dcdc^[\^[]17;rgb:5858/5b5b/7070^[\^[]19;rgb:cdcd/d6d6/f4f4^[\% 
> ```
> 
> We see that ghostty gets back 10, 11, 12 (foreground, background, cursor), but does not return the 17/19. However neither of them return 13, 14, 15, 16, all of which are expected from special colors to return cleanly:
> 
> class TerminalPallette implements TerminalPalletteDetector {
>   ...
>   private async querySpecialColors(timeoutMs = 1200): Promise<Record<number, Hex>> {
>     const out = this.stdout
>     const inp = this.stdin
>     const results: Record<number, Hex> = {
>       10: null,
>       11: null,
>       12: null,
>       13: null,
>       14: null,
[...]

PR #765, opened 2026-03-02
fix(palette): querySpecialColors resistent to term emulators with noi…

MaxDillon:
### Issue for this PR

#705 

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Opentuin getPallette function for querySpecialColors has an idleTimer meant to exit out before the explicit timeout if the terminal does not provide all of the color escape sequences 10-19. The idle timer resets if any escape sequence triggers the onData function again.

This is an issue for terminal emulators like ghostty which are very noisy with the escape sequences they send. Here we only allow the timeout to reset if a response conforming to `OSC_SPECIAL_RESPONSE` is read in the buffer. 

This is possibly overly conservative. If we were being more aggressive, we could only reset the idle timer when an entry in results gets updated from a null value. As it stands, this improves start time of getPallette in ghostty from the max timeout to almost negligable

- I tested locally with the following simple script
```js
#!/usr/bin/env bun

import { createCliRenderer, TextRenderable } from "../index"
import { setupCommonDemoKeys } from "./lib/standalone-keys"

const startTime = Date.now()

// 1. Create renderer
  exitOnCtrlC: true,
  targetFps: 60,
})

// 2. Start automatic render loop
renderer.auto()

// 3. Add trivial renderable
const text = new TextRenderable(renderer, {
  id: "placeholder",
  content: "Renderer active… (toggle stdout with `)",
  left: 0,
  top: 0,
})
renderer.root.add(text)

// 4. Hook up common demo keys (including stdout toggle)
setupCommonDemoKeys(renderer)

// 5. Request palette
const palette = await renderer.getPalette({ timeout: 5000, size: 16 })

console.log("time elapsed", Date.now() - startTime)
```
- Without fix: `[17:54:05] [LOG] 'time elapsed' 5014`
- With fix: `[17:54:28] [LOG] 'time elapsed' 220`

### Checklist

### DIFF
--- a/packages/core/src/lib/terminal-palette.ts
+++ b/packages/core/src/lib/terminal-palette.ts
@@ -230,7 +230,7 @@ export class TerminalPalette implements TerminalPaletteDetector {
 
       const onData = (chunk: string | Buffer) => {
         buffer += chunk.toString()
-        let updated = false;
+        let updated = false
 
         let m: RegExpExecArray | null
         OSC_SPECIAL_RESPONSE.lastIndex = 0
@@ -238,7 +238,7 @@ export class TerminalPalette implements TerminalPaletteDetector {
           const idx = parseInt(m[1], 10)
           if (idx in results) {
             results[idx] = toHex(m[2], m[3], m[4], m[5])
-            updated = true;
+            updated = true
           }
         }
 


kommander:
For that script without the proposed fix I get ~150ms in Ghostty on main, with all colors resolved.
Screenshot 2026-03-02 at 21 31 16

kommander:
Mhh... I just ran it again and now it's 5000ms

Screenshot 2026-03-02 at 21 35 03

kommander:
Okay I consistently get ~200ms with this fix it seems 👍

PR #770, opened 2026-03-02
core: add stdin tokenizer/router

### LINKED ISSUE #748: High-frequency scroll events cause input starvation and dropped keyboard data

simonklee:
Fixes #748 + related #705, #682, #727, #686, [OpenCode 15760](https://github.com/anomalyco/opencode/issues/15760)

We currently have a chunk-level either/or stdin split in `CliRenderer`:

- if chunk looks like mouse, route mouse and return early
- otherwise send the whole chunk to generic stdin buffering
  That design is unsafe for mixed chunks and can silently drop input (e.g. mouse + key in same read).
  Examples:
- `\x1b[<64;10;5Mx` -> trailing `x` can be dropped
- `x\x1b[<64;10;5M` -> mouse can be swallowed by non-mouse path

I want to have one canonical stdin pipeline, but i tried to add this a merge-safe way, which is why this PR is a _bit larger_ and more of a refactor than just a straight parser swap.

`main` uses a split stdin pipeline:

- raw bytes are decoded to UTF-8 too early, before the parser stack can understand the protocol
- mouse handling is pre-filtered before buffering
- `KeyHandler` reparses strings the parser stack already understands

This makes input harder to reason about and can lose details for mixed protocol chunks, X10 mouse payload bytes, delayed escape continuations, and legacy high-byte/meta cases. For instance, [issue 661](https://github.com/anomalyco/opentui/issues/661), i now realize is most likely a case of utf8 decoding happening on X10 mouse bytes

The boundary is:

- StdinParser owns byte framing and protocol recognition. It calls `parseKeypress()` and `MouseParser` internally and creates fully typed `StdinEvent` objects (key, mouse, paste, response).
- KeyHandler owns event dispatch and turns `ParsedKey` into `KeyEvent`. It is not a parser.
- Renderer owns application routing only, switching on `event.type` and forwarding to the right handler.

This removes:

- `StdinBuffer` string tokenizer and its even emitter
- `setEncoding("utf8")` on stdin (decoded strings before framing, wrong order for a binary protocol)
- `MouseParser` instance in renderer that consumed mouse bytes before they ever reached the buffer
- `parseKeypress()` call inside `KeyHandler.processInput()`
- Paste-size coupling to the parser's main byte buffer

Also:

- A Clock interface (I used a similar abstraction in my last Go project) to manually advance time in tests without relying on wall clock time. I only used this in the relevant tests, but if we like this we can expand it to other tests that use timers. Side-effect is that it makes timing tests quick and deterministic, which is a nice bonus.
- In main we're not testing the `stdin` ReadStream directly AFAIK, which is why the setEncoding issues have gone unnoticed, i think. I haven't addressed this yet. - Added many tests to the parser :)
- Its a large PR, but the lines added is deceptive. The new parser is a larger because we're operating on bytes, but other things are mostly glue + tests + comments.

In one sentence: main decodes first, parses later. In the PR we parse bytes first, decode only when semantics are known.

### DIFF
--- a/packages/core/src/lib/parse.keypress.ts
+++ b/packages/core/src/lib/parse.keypress.ts
@@ -67,6 +67,11 @@ const keyName: Record<string, string> = {
   "[c": "right",
   "[d": "left",
   "[e": "clear",
+  /* option + arrow keys (old style) */
+  f: "right",
+  b: "left",
+  p: "up",
+  n: "down",
 
   "[2$": "insert",
   "[3$": "delete",
@@ -323,7 +328,12 @@ export const parseKeypress = (s: Buffer | string = "", options: ParseKeypressOpt
     const char = parts[1]!
     const isUpperCase = /^[A-Z]$/.test(char)
 
-    if (isUpperCase) {
+    // Check if uppercase F or B map to arrow keys (old terminal style)
+    if (char === "F") {
+      key.name = "right"
+    } else if (char === "B") {
+      key.name = "left"
+    } else if (isUpperCase) {
       key.shift = true
       key.name = char
     } else {
--- a/packages/core/src/lib/parse.keypress.test.ts
+++ b/packages/core/src/lib/parse.keypress.test.ts
@@ -1763,36 +1763,61 @@ test("parseKeypress - modifyOtherKeys modified tab, space, and backspace keys", 
   expect(ctrlBackspace8.ctrl).toBe(true)
 })
 
-test("parseKeypress - meta+b/f stay as literal meta chords", () => {
-  // Terminal.app may reuse these bytes for Option+word navigation, but the parser
-  // preserves the raw Meta-b/Meta-f chords instead of rewriting them globally.
-  const metaB = parseKeypress("\x1bb")!
-  expect(metaB.name).toBe("b")
-  expect(metaB.meta).toBe(true)
-  expect(metaB.shift).toBe(false)
+test("parseKeypress - meta+arrow keys with uppercase F and B (old style)", () => {
+  // Some terminals send ESC followed by uppercase F/B for meta+arrow keys
+  // ONLY uppercase F and B map to arrow keys (not P/N which require actual shift)
+  // Lowercase f/b are just regular meta+letter combinations
 
-  const metaF = parseKeypress("\x1bf")!
+  // Meta+Right (uppercase F)
+  const metaRight = parseKeypress("\u001BF")!
+  expect(metaRight.name).toBe("right")
+  expect(metaRight.meta).toBe(true)
+  expect(metaRight.shift).toBe(false)
+  expect(metaRight.ctrl).toBe(false)
+  expect(metaRight.option).toBe(false)
+  expect(metaRight.sequence).toBe("\u001BF")
+  expect(metaRight.raw).toBe("\u001BF")
+
+  // Meta+Left (uppercase B)
+  const metaLeft = parseKeypress("\u001BB")!
+  expect(metaLeft.name).toBe("left")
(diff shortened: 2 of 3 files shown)


packages/core/src/lib/parse.keypress.ts
```diff
@@ -67,11 +67,6 @@ const keyName: Record<string, string> = {
   "[c": "right",
   "[d": "left",
   "[e": "clear",
-  /* option + arrow keys (old style) */
-  f: "right",
-  b: "left",
-  p: "up",
-  n: "down",
```
simonklee, line 74: We're doing something odd with these, but i meant to revert it to keep it in a later branch. Will fix.

packages/core/src/lib/parse.keypress.ts
```diff
@@ -303,21 +314,16 @@ export const parseKeypress = (s: Buffer | string = "", options: ParseKeypressOpt
...
+  } else if (s.length === 1 || (s.length === 2 && s.codePointAt(0)! > 0xffff)) {
+    // Single character (including emoji/surrogate pairs above BMP)
     key.name = s
   } else if ((parts = metaKeyCodeRe.exec(s))) {
     // meta+character key
     key.meta = true
     const char = parts[1]!
     const isUpperCase = /^[A-Z]$/.test(char)
 
-    // Check if uppercase F or B map to arrow keys (old terminal style)
```
simonklee, line 315: Ditto.

PR #776, opened 2026-03-04
refactor: replace `Bun.file` usages with `node:fs`

bienzaaron:
see #2. 

Replaces [`Bun.file(...)`](https://bun.com/reference/bun/file) usage with `node:fs` to start working toward node.js and deno compatibility.

I also updated AGENTS.md to recommend against Bun-specific APIs in generated code.


kommander:
Cool, I need to double check that with an opencode build.

PR #779, opened 2026-03-05
fix(core): prevent mouse escape sequences from leaking into keyboard input

Jerwand:
## Summary

Fixes the root cause of mouse escape sequences (e.g. `[555;129;29M`) appearing as typed text in input fields during normal use. This happens when stdin delivers mouse data split across chunk boundaries, which occurs regularly under high mouse activity.

- Add `parseAllMouseEventsWithOffset()` to `MouseParser` — returns consumed byte count and partial-sequence detection so callers know exactly what was parsed and what remains
- Add `isPartialMouseSequence()` to detect truncated SGR/Basic mouse sequences at chunk boundaries (e.g. `\x1b[<35;20` missing the terminating `M`)
- Add `_mousePartialBuffer` to `CliRenderer` that preserves incomplete mouse sequences between stdin chunks
- Rewrite `stdinListener` to properly route unconsumed non-mouse data to `StdinBuffer` instead of silently dropping it
- Update split-height test: mouse events above the render area are now correctly ignored instead of leaking to input handlers

## Root cause

The old `stdinListener` had an all-or-nothing design:

```typescript
// OLD (buggy)
  return  // entire chunk treated as mouse — any non-mouse remainder is lost
}
this._stdinBuffer.process(data)  // entire chunk treated as keyboard
```

When a mouse sequence was split across two stdin chunks (e.g. `\x1b[<35;20;5M\x1b[<35` followed by `;21;5M`), the partial tail was dropped and its completion in the next chunk leaked into the keyboard handler as literal text.

## Three scenarios this fixes

| Scenario | Before (bug) | After (fix) |
|----------|-------------|-------------|
| Split: `\x1b[<35;20;5M\x1b[<35` + `;21;5M` | `;21;5M` appears as typed text | Buffered and correctly parsed |
| Mixed: `\x1b[<35;20;5Mhello` | `hello` silently lost | Mouse → handler, `hello` → keyboard |
| Rapid: 10 events + partial | Last event lost | All 11 events parsed correctly |

## Test plan

- [x] All 81 existing mouse parser tests pass
- [x] 10 new tests for `parseAllMouseEventsWithOffset` covering split sequences, partial detection, mixed data, and chunk reassembly
- [x] All 35 renderer mouse tests pass (1 test updated to match corrected behavior)
- [x] All 180 keyboard/keypress tests pass (no regressions)

Fixes https://github.com/anomalyco/opencode/issues/8614
Related: https://github.com/anomalyco/opencode/issues/7316, https://github.com/anomalyco/opencode/issues/11636

### DIFF
--- a/packages/core/src/lib/parse.mouse.test.ts
+++ b/packages/core/src/lib/parse.mouse.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect, beforeEach } from "bun:test"
-import { MouseParser, type RawMouseEvent } from "./parse.mouse"
+import { MouseParser, type RawMouseEvent, type MouseParseResult } from "./parse.mouse"
 
 // Encode a basic/X10 mouse event: ESC [ M Cb Cx Cy
 // buttonByte is the logical value (before the +32 wire offset), x/y are 0-based.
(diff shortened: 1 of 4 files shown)


lortimer:
I'm facing this issue sometimes in the terminal in Webstorm. It seems intermittent because I was using it without issue earlier today. Why did this PR get closed? It's a real problem that makes using opencode impossible

Issue #775, opened 2026-03-04
WSL2: TUI frame stalls when system clock jumps backward (Date.now/setTimeout scheduling)


turazashvili:
### Summary
On WSL2, the TUI intermittently freezes for ~2–10 seconds while typing or scrolling.
This appears to happen when wall-clock time jumps backward (e.g. host/guest time sync interactions in WSL).  
Renderer scheduling currently relies on `Date.now()` + delayed `setTimeout`, which is sensitive to non-monotonic clock changes.
### Environment
- OS: Windows + WSL2
- Distro: Ubuntu (WSL2)
- Runtime: Bun
- Consumer app: opencode (uses `@opentui/core`)
- Related reports:
  - https://github.com/anomalyco/opencode/issues/5361
  - https://github.com/microsoft/WSL/issues/11790
### Observed behavior
- During normal interaction (typing/scrolling), UI occasionally stalls for several seconds.
- Stall resolves on its own; then interaction resumes.
- Reproducible in WSL2, not observed the same way on native Linux/macOS in my tests.
### Why this likely happens
Frame timing appears to use wall-clock (`Date.now`) and one-shot delayed `setTimeout`.  
If wall-clock moves backward, delay calculations can become effectively “too far in the future,” causing visible render stalls.
### Local workaround that resolves it
I patched `@opentui/core` locally and the freezes disappeared by:
1. Using monotonic clock (`performance.now()`) for frame timing
2. Replacing delayed one-shot `setTimeout(..., delay)` with a deadline loop:
   - `deadline = performance.now() + delay`
   - `setImmediate(check)` until `performance.now() >= deadline`, then render/loop
3. Clearing scheduled handles with `clearImmediate(...)`
### Proposed source-level change (renderer)
In `packages/core/src/renderer.ts`:
- Use `performance.now()` in:
  - `startRenderLoop()`
  - `requestRender()`
  - `loop()`
- Update `renderTimeout` type to `ReturnType<typeof setImmediate> | null`
- Replace delayed `setTimeout` frame scheduling with setImmediate + deadline check
- Use `clearImmediate` consistently for `renderTimeout`
### Notes
I can help validate a test/canary build on WSL2 once a patch is available.

turazashvili:
@simonklee is there anyone working on this? I saw closed PR. If this is relevant, I can push PR in the next couple days :)

PR #782, opened 2026-03-05
Use a monotonic clock for renderer scheduling

ermalos:
## Summary
- replace `Date.now()`-based renderer scheduling with a monotonic clock
- clamp negative elapsed times so backward clock jumps cannot stretch frame delays
- add a regression test that simulates a backward clock jump during `requestRender()`

## Why
On WSL, host/guest time corrections can jump the wall clock backwards. The renderer currently uses `Date.now()` to compute request/render delays, so a backward jump can turn a 16ms frame delay into a multi-second stall.

## Testing
- `npx --yes bun@1.3.10 test packages/core/src/tests/renderer.clock.test.ts packages/core/src/tests/renderer.control.test.ts packages/core/src/tests/renderer.idle.test.ts`


simonklee:
Related https://github.com/anomalyco/opentui/issues/775

simonklee:
What made you change your mind on this pr?

PR #784, opened 2026-03-06
fix(markdown): harden code block fallback and info parsing

ryanfreckleton:
Downstream: OpenCode

In OpenCode, fenced code blocks with unknown/unsupported languages are illegible on light terminal themes (for example, Solarized Light) because fallback plain-text rendering had no guaranteed base fg/bg.

## Why here

- The readability invariant belongs in `CodeRenderable`, not `MarkdownRenderable`, so all call sites (markdown and standalone) inherit safe defaults.
- This matches how Rich, Glamour, and syntect/bat treat default style ownership.

## Scope
- `CodeRenderable` now derives baseline `fg`/`bg`/`attributes` from `syntaxStyle["default"]` when those props are omitted (and reapplies on `syntaxStyle` change without overriding explicit caller values).
- `MarkdownRenderable` now normalizes fence language info as `first token + lowercase + extToFiletype alias mapping` (for example, `BASH title=...` -> `shell`, `ToMl` -> `toml`).

## Validation
- Manual: verified in the linked OpenCode worktree (`~/src/opencode-opentui-test`) that unknown-language fences remain readable on light themes.
- Automated: `bun test packages/core/src/renderables/Code.test.ts packages/core/src/renderables/__tests__/Markdown.test.ts`


asaf-genie:
+1 for the alias fix (ts -> typescript)
GPT 5.4 loves using `ts` and it doesn't render properly in opencode

PR #788, opened 2026-03-07
fix: set wrap width on resize for text buffer renderables

cjroth:
## Summary

When `wrapMode` is enabled (`"char"` or `"word"`) and a `TextBufferRenderable` is resized, the wrap width was not being updated in `onResize()`. This caused text to retain the old wrap width after a container resize.

The fix adds the same `setWrapWidth` call that already exists in:
- Initialization (`afterSetup`, line ~94)
- The `wrapMode` setter (line ~283)

Both use the identical guard pattern: `if (this._wrapMode !== "none" && width > 0)`.

## Changes

- `packages/core/src/renderables/TextBufferRenderable.ts`: Call `this.textBufferView.setWrapWidth(width)` in `onResize()` when wrap mode is active

### DIFF
--- a/packages/core/src/renderables/TextBufferRenderable.ts
+++ b/packages/core/src/renderables/TextBufferRenderable.ts
@@ -331,6 +331,9 @@ export abstract class TextBufferRenderable extends Renderable implements LineInf
   }
 
   protected onResize(width: number, height: number): void {
+    if (this._wrapMode !== "none" && width > 0) {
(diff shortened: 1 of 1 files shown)


cjroth:
Closing — this fix causes a layout feedback loop for flex-sized elements (setWrapWidth on resize triggers re-wrap → intrinsic size change → another resize → collapse). Needs a more targeted approach.

PR #791, opened 2026-03-08
fix(buffer): preserve wide characters under translucent overlays

### LINKED ISSUE #837: CJK text rendered underneath a popup/overlay gets hidden or overwritten
### Description

This issue is filed upstream based on OpenCode issue #18088:
https://github.com/anomalyco/opencode/issues/18088

When a popup/overlay is opened on top of existing content, previously rendered CJK text becomes hidden. Since the issue appears during overlay rendering and affects underlying CJK text, it seems more likely to belong to OpenTUI’s rendering/compositing layer than to application-specific logic.

### Notes

This may be related to overlay blending / wide-character grapheme handling in the buffer renderer.

I’m opening this issue first because I’d like to work on a PR.

aarcamp:
This PR improves how alpha-blended overlays interact with wide terminal grapheme spans.

When a translucent overlay tints existing content without replacing its character, wide non-emoji graphemes are now blended span-wise instead of cell-by-cell. That preserves multi-cell text such as CJK under translucent fills instead of letting the fill overwrite or inconsistently affect individual cells of the grapheme. Wide emoji-like graphemes are handled differently: under those same translucent overlays they render as a stable ASCII [] placeholder, which avoids inconsistent results from trying to tint color emoji directly.

For translucent boxes, fills now use a hybrid strategy when the visible fill is large enough to have a separable perimeter and interior. Interior cells use the span-preserving path, while perimeter cells switch to a clipped path when a box edge crosses a wide grapheme so box boundaries stay visually straight. That clipped-edge behavior is only used when the visible box perimeter actually touches a wide span; narrow-text overlays and very small/clipped visible fills keep the normal preserved path. The same logic respects offscreen and overflow:hidden clipping, including buffered ancestors. For bordered boxes, the border ring is treated as the clipped perimeter so interior space is not needlessly reduced and border cells are not double-tinted.

Buffered box rendering was also tightened up so these translucent cases behave the same whether the box renders directly or through its framebuffer. Framebuffer-local rendering, reuse, and reset behavior were adjusted so buffered rerenders do not accumulate stale tint, replay stale contents, or misapply clipping when switching between buffered and bypassed paths.

The renderer was also updated to repaint changed wide grapheme spans more reliably. It preclears replaced wide-span output before writing the new glyph, avoids emitting continuation spaces immediately after newly written wide graphemes, and avoids unnecessary cursor movement between adjacent or following wide output in fallback rendering. This prevents stale wide-glyph artifacts and ghosting after overlay-driven repaints, including placeholder-to-emoji transitions.

User-visible behavior:

- wide CJK text remains readable under translucent fills when fully inside the overlay
- wide emoji-like graphemes under translucent fills appear as []
- translucent box edges stay visually straight when crossing wide graphemes in perimeter/interior box fills
- small translucent overlays preserve normal narrow-text behavior when no wide span is involved
- clipped and buffered translucent boxes now match direct rendering more closely, including under overflow clipping
- repainting after overlays move or change no longer leaves behind wide-glyph corruption or ghosting

The changes are covered by native buffer and renderer tests plus BoxRenderable regression tests, and the wide grapheme demo/test map was updated to reflect the new cases.

Screen recording using modified wide grapheme overlay demos:

https://github.com/user-attachments/assets/031897d0-b4a0-4a42-a8e1-2910707f9ade

Closes #837

### DIFF
--- a/packages/core/docs/wide-grapheme-alpha-blending-tests.svg
+++ b/packages/core/docs/wide-grapheme-alpha-blending-tests.svg
@@ -1,0 +1,444 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1540 920" width="1540" height="920" role="img" aria-labelledby="title desc">
+  <title id="title">Wide grapheme alpha-blending test map</title>
+  <desc id="desc">Twelve panels illustrating the current wide-grapheme alpha-blending contract: wide text is preserved through interior translucent fills, wide emoji are replaced with [] placeholders, scissor clipping only updates cells inside the clip, boundary-band fills clip crossed spans with fillRectClipWideGraphemes, and renderer repaints preclear placeholder spans before restoring emoji and following cells.</desc>
+  <style>
+    .bg { fill: #f7f4ee; }
+    .panel { fill: #fffdf8; stroke: #cfc7b8; stroke-width: 1.5; }
+    .title { font: 700 28px Menlo, Consolas, monospace; fill: #1d1d1d; }
+    .subtitle { font: 400 14px Menlo, Consolas, monospace; fill: #5a5a5a; }
+    .panel-title { font: 700 17px Menlo, Consolas, monospace; fill: #1d1d1d; }
+    .panel-sub { font: 400 11px Menlo, Consolas, monospace; fill: #666; }
+    .label { font: 700 11px Menlo, Consolas, monospace; fill: #666; letter-spacing: 0.6px; }
+    .call-label { font: 700 9px Menlo, Consolas, monospace; fill: #666; letter-spacing: 0.15px; }
+    .note { font: 400 11px Menlo, Consolas, monospace; fill: #4d4d4d; }
+    .note-center { text-anchor: middle; }
+    .legend { font: 400 12px Menlo, Consolas, monospace; fill: #444; }
+    .cell { fill: #fbfbfb; stroke: #b8b8b8; stroke-width: 1; rx: 4; }
+    .wg { fill: #ddd8f8; stroke: #7569b8; stroke-width: 1; rx: 4; }
+    .emoji { fill: #fde0b8; stroke: #c67a21; stroke-width: 1; rx: 4; }
+    .tint { fill: #e5f3df; stroke: #78a06a; stroke-width: 1; rx: 4; }
+    .preserve { fill: #c7efd4; stroke: #3f8a63; stroke-width: 1.25; rx: 4; }
+    .placeholder-cell { fill: #f4d9e8; stroke: #a75e84; stroke-width: 1; rx: 4; }
+    .edge-fill { fill: #dfe9f8; stroke: #6d87b4; stroke-width: 1; rx: 4; }
+    .overlay { fill: #f9d7a8; opacity: 0.42; stroke: #d48e3a; stroke-width: 1.5; stroke-dasharray: 4 3; rx: 4; }
+    .overlay-line { fill: none; stroke: #d48e3a; stroke-width: 2; stroke-dasharray: 5 3; rx: 4; }
+    .scissor { fill: none; stroke: #d95a5a; stroke-width: 2; stroke-dasharray: 5 4; rx: 6; }
+    .wg-text { font: 700 13px Menlo, Consolas, monospace; fill: #5f4512; text-anchor: middle; }
+    .b-text { font: 700 16px Menlo, Consolas, monospace; fill: #d97b1d; }
+    .ph-text { font: 700 16px Menlo, Consolas, monospace; fill: #6c3554; text-anchor: middle; }
+    .dim { fill: #8b8b8b; }
+  </style>
+
+  <defs>
+    <g id="row-base">
+      <rect class="cell" x="64" y="0" width="28" height="28"/>
+      <rect class="cell" x="96" y="0" width="28" height="28"/>
+      <rect class="cell" x="128" y="0" width="28" height="28"/>
+      <rect class="cell" x="160" y="0" width="28" height="28"/>
+      <rect class="cell" x="192" y="0" width="28" height="28"/>
+      <rect class="cell" x="224" y="0" width="28" height="28"/>
+      <rect class="cell" x="256" y="0" width="28" height="28"/>
+      <rect class="cell" x="288" y="0" width="28" height="28"/>
+    </g>
+  </defs>
+
+  <rect class="bg" x="0" y="0" width="1540" height="920"/>
+
+  <text class="title" x="24" y="44">Wide Grapheme Alpha-Blending Test Map</text>
+  <text class="subtitle" x="24" y="68">WG = wide text. EM = wide emoji. Lavender = original wide text. Gold = original emoji. Orange = translucent overlay.</text>
+  <text class="subtitle" x="24" y="88">Pale green = overlay-filled cell. Mint = preserved wide text. Rose = [] placeholder. Blue = clipped edge fill.</text>
+
+  <rect class="wg" x="24" y="104" width="28" height="18"/>
+  <text class="legend" x="60" y="117">wide text grapheme</text>
+  <rect class="emoji" x="248" y="104" width="28" height="18"/>
+  <text class="legend" x="284" y="117">wide emoji grapheme</text>
+  <rect class="overlay" x="474" y="104" width="28" height="18"/>
+  <text class="legend" x="510" y="117">translucent overlay</text>
+  <rect class="tint" x="700" y="104" width="28" height="18"/>
+  <text class="legend" x="736" y="117">overlay-filled cell</text>
+  <rect class="preserve" x="982" y="104" width="28" height="18"/>
+  <text class="legend" x="1018" y="117">preserved wide text</text>
+  <rect class="placeholder-cell" x="1250" y="104" width="28" height="18"/>
+  <text class="legend" x="1286" y="117">[] placeholder</text>
+
+  <rect class="edge-fill" x="24" y="126" width="28" height="18"/>
+  <text class="legend" x="60" y="139">clipped edge fill</text>
+  <rect class="scissor" x="248" y="124" width="32" height="22"/>
+  <text class="legend" x="292" y="139">scissor clip</text>
+
+  <g transform="translate(24,148)">
+    <rect class="panel" x="0" y="0" width="360" height="240" rx="12"/>
+    <text class="panel-title" x="16" y="28">1. Basic preservation</text>
+    <text class="panel-sub" x="16" y="44">setCellWithAlphaBlending()</text>
+    <text class="label" x="20" y="66">BEFORE</text>
+    <g transform="translate(0,72)">
+      <use href="#row-base"/>
+      <rect class="wg" x="160" y="0" width="28" height="28"/>
+      <rect class="wg" x="192" y="0" width="28" height="28"/>
+      <text class="wg-text" x="190" y="18">WG</text>
+    </g>
+    <text class="call-label" x="20" y="126">SETCELLWITHALPHABLENDING()</text>
+    <g transform="translate(0,132)">
+      <use href="#row-base"/>
+      <rect class="wg" x="160" y="0" width="28" height="28"/>
+      <rect class="wg" x="192" y="0" width="28" height="28"/>
+      <rect class="overlay" x="64" y="0" width="252" height="28"/>
+      <text class="wg-text" x="190" y="18">WG</text>
+    </g>
+    <text class="label" x="20" y="186">AFTER</text>
+    <g transform="translate(0,192)">
+      <use href="#row-base"/>
+      <rect class="tint" x="64" y="0" width="28" height="28"/>
+      <rect class="tint" x="96" y="0" width="28" height="28"/>
+      <rect class="tint" x="128" y="0" width="28" height="28"/>
+      <rect class="preserve" x="160" y="0" width="28" height="28"/>
+      <rect class="preserve" x="192" y="0" width="28" height="28"/>
+      <rect class="tint" x="224" y="0" width="28" height="28"/>
+      <rect class="tint" x="256" y="0" width="28" height="28"/>
+      <rect class="tint" x="288" y="0" width="28" height="28"/>
+      <text class="wg-text" x="190" y="18">WG</text>
+    </g>
+  </g>
+
+  <g transform="translate(400,148)">
+    <rect class="panel" x="0" y="0" width="360" height="240" rx="12"/>
+    <text class="panel-title" x="16" y="28">2. Emoji placeholder</text>
+    <text class="panel-sub" x="16" y="44">fillRect() over a wide emoji</text>
+    <text class="label" x="20" y="66">BEFORE</text>
+    <g transform="translate(0,72)">
+      <use href="#row-base"/>
+      <rect class="emoji" x="160" y="0" width="28" height="28"/>
+      <rect class="emoji" x="192" y="0" width="28" height="28"/>
+      <text class="wg-text" x="190" y="18">EM</text>
+    </g>
+    <text class="call-label" x="20" y="126">FILLRECT()</text>
+    <g transform="translate(0,132)">
+      <use href="#row-base"/>
+      <rect class="emoji" x="160" y="0" width="28" height="28"/>
+      <rect class="emoji" x="192" y="0" width="28" height="28"/>
+      <rect class="overlay" x="64" y="0" width="252" height="28"/>
+      <text class="wg-text" x="190" y="18">EM</text>
+    </g>
+    <text class="note note-center" x="180" y="175">wide emoji become [] instead of bleeding through</text>
+    <text class="label" x="20" y="186">AFTER</text>
+    <g transform="translate(0,192)">
+      <use href="#row-base"/>
+      <rect class="tint" x="64" y="0" width="28" height="28"/>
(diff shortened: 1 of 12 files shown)


simonklee:
This correctly identifies the issue that blendCells() don't to preserve characters with display width > 1, so any translucent overlay replaced wide graphemes with spaces. Removing the destWidthIsOne guard is the right call, and the double-blend fix in setCellWithAlphaBlending() is sound: if the blended result preserves the same grapheme/continuation char, it skips set() and writes style back to only the touched cell, which avoids continuation cells getting blended twice in left-to-right loops.

The problem is what happens next. Because the fix is cell-local, partial overlap creates split styling. If 東 occupies x=2 and x=3 and an overlay only covers x=3, only the continuation cell gets tinted. One visual grapheme now has two different background colors across its cells, which isn't a stable model.

You fix this by adding span-snapping logic inside fillRect(), when it detects it's touching part of a preserved grapheme span, it manually expands the tint to the whole span. That's the bit that feels wrong, and I think the instinct is right.

The real rule should be "if you alpha-blend over part of a wide grapheme, tint the whole span", now lives only in fillRect(). But drawText() and drawFrameBuffer() still use the cell-local path in setCellWithAlphaBlending(), so they can still split a wide grapheme on partial overlap. 

I reproduced this: a translucent `drawText(" ", 3, 0, ...)` over 東 at x=2 tints only x=3, and a 1×1 respectAlpha child framebuffer placed at x=3 does the same thing. The fillRect() workaround also has a scissor-bleed issue. Once it decides to tint the whole grapheme span, it writes beyond the clipped cell range.

I prefer a general fix: Define one central rule for "alpha overlay over a preserved wide grapheme" and put it in a single helper that all alpha-writing paths use. The logic would be: inspect the destination cell; if the char isn't preserved, do the normal blend; if it's preserved but narrow, update one cell; if it's part of a wide grapheme, resolve the whole span, blend once from a canonical source cell, and write the result to the entire span. Then setCellWithAlphaBlending() becomes the source of truth and fillRect(), drawText(), and drawFrameBuffer() all get correct behavior without needing their own special cases.

aarcamp:
> I prefer a general fix: Define one central rule for "alpha overlay over a preserved wide grapheme" and put it in a single helper that all alpha-writing paths use. The logic would be: inspect the destination cell; if the char isn't preserved, do the normal blend; if it's preserved but narrow, update one cell; if it's part of a wide grapheme, resolve the whole span, blend once from a canonical source cell, and write the result to the entire span. Then setCellWithAlphaBlending() becomes the source of truth and fillRect(), drawText(), and drawFrameBuffer() all get correct behavior without needing their own special cases.

Good call, I pushed an update. What do you think of setCellWithAlphaBlending() + BlendCursor() re: generality?

I also added the following SVG visualization to `packages/core/docs/` because the wide variety of cases is way too hard to all keep straight in my head:

wide-grapheme-alpha-blending-tests

ariane-emory:
I spent about six hours last night trying to figure out what was happening when I stumbled upon this issue in my own fork of OpenCode. 

I dearly hope that this PR will come in to a safe landing!

simonklee:
I just smoke tested this branch. Is this result what you expect after your changes?


https://github.com/user-attachments/assets/dd63d004-3686-42b7-bcfd-e2998677989b

aarcamp:
> I just smoke tested this branch. Is this result what you expect after your changes?

@simonklee The commands dialog in opencode is opaque, is it not? For me it is not see-thru even for single width chars. Here's a quick test w/ red tinted translucent overlay and some [emojis](https://www.unicode.org/Public/emoji/latest/emoji-test.txt):

Screenshot 2026-03-15 at 6 44 47 AM

simonklee:
Okay - i think this is more of a design decision. But i don't love that emojis bleed through.  CJK characters blend fine in your branch, while a family emoji under a 60% opacity scrim looks fully bright while the surrounding background gets properly darkened. It's not a bug in the blending logic, just a limitation. 

One option is to detect whether the grapheme is emoji vs CJK and only preserve the ones terminals can actually tint. Another one is to have some threshold and use a heuristic approach.


https://github.com/user-attachments/assets/5e67ee37-054c-4658-8e37-8a87bc96deef

simonklee:
@kommander what do you think?

kommander:
@simonklee what you said earlier, I'd try a simple unicode range limitation for characters that can be tinted. Emojis bleeding through just looks weird. Chars that cannot be tinted could be renderer with a placeholder so it doesn't just look empty.

aarcamp:
> One option is to detect whether the grapheme is emoji vs CJK and only preserve the ones terminals can actually tint. Another one is to have some threshold and use a heuristic approach.

What about making it configurable with a policy?

```
pub const AlphaOverlayPreserveMask = u8;

pub const preserve_none: AlphaOverlayPreserveMask = 0;
pub const preserve_text: AlphaOverlayPreserveMask = 1 << 0;
pub const preserve_emoji: AlphaOverlayPreserveMask = 1 << 1;
pub const preserve_all: AlphaOverlayPreserveMask = preserve_text | preserve_emoji;

pub fn pushAlphaOverlayPreserveMask(self: *OptimizedBuffer, mask: AlphaOverlayPreserveMask) !void

export const AlphaOverlayPreserve = {
  none: 0,
  text: 1 << 0,
  emoji: 1 << 1,
  all: (1 << 0) | (1 << 1),
} as const

buffer.pushAlphaOverlayPreserveMask(AlphaOverlayPreserve.text)
buffer.fillRect(...)
buffer.popAlphaOverlayPreserveMask()
```

I tend to agree that emoji colors bleeding through can look weird so probably the default should be just 'text'. But I wouldn't assume there's no use case for preserving emojis too. Note, the intent of the push/pop pattern is to allow different compositing policies to apply to different drawing scopes within the same buffer.

> Chars that cannot be tinted could be renderer with a placeholder so it doesn't just look empty.

How about using `[]` (width 2), `[·]` (width 3), `[··]` (width 4) for the placeholders.

aarcamp:
> It's not a bug in the blending logic, just a limitation.

Yeah for this reason I'm wondering if we can merge this as-is and supply a folllow-up PR for the emoji treatment? IMO it's already a lot better since right now wide chars are just ignored entirely.

simonklee:
> > It's not a bug in the blending logic, just a limitation.
> 
> Yeah for this reason I'm wondering if we can merge this as-is and supply a folllow-up PR for the emoji treatment? IMO it's already a lot better since right now wide chars are just ignored entirely.

My point is that technically it's technically not a bug — but from a ux point of view it's a regression. Another thing, if in this pr you have distortion on that draggable rect. You can see it in the video. That doesn't happen in main.

aarcamp:
> My point is that technically it's technically not a bug — but from a ux point of view it's a regression. Another thing, if in this pr you have distortion on that draggable rect. You can see it in the video. That doesn't happen in main.

Thanks, I didn't notice the distortion issue at first.

This suggests the next step would be to implement something like the preserve mask policy I proposed above. Then we can set appropriate defaults by caller, i.e.:

- box fills / scrims / drag overlays: preserve_none
- pure text overlays: preserve_text (user can override to add preserve_emoji behavior if desired)

Does that sound reasonable? (Tbh I don't know I'd agree the emoji color bleed is a ux regression necessarily but the draggable box distortion definitely is!)

simonklee:
```
import {
  createCliRenderer,
  FrameBufferRenderable,
  RGBA,
  TextRenderable,
  BoxRenderable,
  OptimizedBuffer,
  t,
  bold,
  underline,
  fg,
  type MouseEvent,
  type KeyEvent,
} from "../index.js"
import type { CliRenderer, RenderContext } from "../index.js"
import { setupCommonDemoKeys } from "./lib/standalone-keys.js"

const GRAPHEME_LINES: string[] = [
  "東京都  北京市  서울시  大阪府  名古屋  横浜市  上海市",
  "👨‍👩‍👧‍👦  👩🏽‍💻  🏳️‍🌈  🇺🇸  🇩🇪  🇯🇵  🇮🇳  家族  絵文字  🎉🎊🎈",
  "こんにちは世界  你好世界  안녕하세요  สวัสดี  مرحبا",
  "漢字テスト  中文测试  한국어  日本語  繁體中文  简体中文",
  "🚀 Full-width: ＡＢＣＤＥＦ  Half: abcdef  ½ ⅞ ⅓",
  "混合テキスト mixed text with 漢字 and emoji 🎯",
]

const HEADER_HEIGHT = 2

let nextZIndex = 101
let draggableBoxes: DraggableBox[] = []
let scrimVisible = false
let scrim: BoxRenderable | null = null
let headerDisplay: TextRenderable | null = null

class DraggableBox extends BoxRenderable {
  private isDragging = false
  private dragOffsetX = 0
  private dragOffsetY = 0
  private alphaPercentage: number

  constructor(
    ctx: RenderContext,
    id: string,
    x: number,
    y: number,
    width: number,
    height: number,
    bg: RGBA,
    zIndex: number,
  ) {
    super(ctx, {
      id,
      width,
      height,
      zIndex,
      backgroundColor: bg,
      position: "absolute",
      left: x,
      top: y,
    })
    this.alphaPercentage = Math.round(bg.a * 100)
  }

  protected renderSelf(buffer: OptimizedBuffer): void {
    super.renderSelf(buffer)

    const alphaText = `${this.alphaPercentage}%`
    const centerX = this.x + Math.floor(this.width / 2 - alphaText.length / 2)
    const centerY = this.y + Math.floor(this.height / 2)

    buffer.drawText(alphaText, centerX, centerY, RGBA.fromInts(255, 255, 255, 220))
  }

  protected onMouseEvent(event: MouseEvent): void {
    switch (event.type) {
      case "down":
        this.isDragging = true
        this.dragOffsetX = event.x - this.x
        this.dragOffsetY = event.y - this.y
        this.zIndex = nextZIndex++
        event.stopPropagation()
        break

      case "drag-end":
        if (this.isDragging) {
          this.isDragging = false
          event.stopPropagation()
        }
        break

      case "drag":
        if (this.isDragging) {
          const newX = event.x - this.dragOffsetX
          const newY = event.y - this.dragOffsetY

          this.x = Math.max(0, Math.min(newX, this._ctx.width - this.width))
          this.y = Math.max(0, Math.min(newY, this._ctx.height - this.height))

          event.stopPropagation()
        }
        break
    }
  }
}

class GraphemeBackground extends FrameBufferRenderable {
  constructor(ctx: RenderContext, id: string, width: number, height: number) {
    super(ctx, {
      id,
      width,
      height,
      position: "absolute",
      left: 0,
      top: HEADER_HEIGHT,
      respectAlpha: false,
    })

    this.fillGraphemes(width, height)
  }

  private fillGraphemes(width: number, height: number) {
    const fgColor = RGBA.fromInts(220, 220, 220, 255)
    const bgColor = RGBA.fromInts(10, 14, 20, 255)
    this.frameBuffer.clear(bgColor)
    for (let y = 0; y < height; y++) {
      const line = GRAPHEME_LINES[y % GRAPHEME_LINES.length]
      this.frameBuffer.drawText(line, 2, y, fgColor, bgColor)
    }
  }
}

function toggleScrim(renderer: CliRenderer) {
  scrimVisible = !scrimVisible
  if (scrim) scrim.visible = scrimVisible
  updateHeader()
  renderer.requestRender()
}

function updateHeader() {
  if (!headerDisplay) return
  const dimLabel = scrimVisible ? "D: hide scrim" : "D: show scrim"
  headerDisplay.content = t`${bold(fg("#00D4AA")("Wide Grapheme Overlay"))} ${fg("#A8A8B2")(`| ${dimLabel} | Drag boxes over CJK/emoji | Ctrl+C: quit`)}`
}

export function run(renderer: CliRenderer): void {
  renderer.start()
  renderer.setBackgroundColor("#0A0E14")

  const root = new BoxRenderable(renderer, { id: "wg-overlay-root" })
  renderer.root.add(root)

  // Header row
  headerDisplay = new TextRenderable(renderer, {
    id: "wg-header",
    height: HEADER_HEIGHT,
    position: "absolute",
    left: 2,
    top: 0,
    zIndex: 200,
    selectable: false,
  })
  updateHeader()
  root.add(headerDisplay)

  // Background filled with repeating wide grapheme lines, below the header
  const bgHeight = renderer.terminalHeight - HEADER_HEIGHT
  const background = new GraphemeBackground(renderer, "wg-background", renderer.terminalWidth, bgHeight)
  root.add(background)

  // Full-screen dimming scrim (same as opencode dialog backdrop: RGBA(0,0,0,150))
  scrim = new BoxRenderable(renderer, {
    id: "wg-scrim",
    position: "absolute",
    left: 0,
    top: HEADER_HEIGHT,
    width: renderer.terminalWidth,
    height: bgHeight,
    backgroundColor: RGBA.fromInts(0, 0, 0, 150),
    zIndex: 50,
  })
  scrim.visible = false
  root.add(scrim)

  // Draggable boxes at various alpha levels
  const box1 = new DraggableBox(
    renderer,
    "wg-box-50",
    4,
    HEADER_HEIGHT + 1,
    25,
    8,
    RGBA.fromValues(64 / 255, 176 / 255, 255 / 255, 128 / 255),
    100,
  )
  root.add(box1)
  draggableBoxes.push(box1)

  const box2 = new DraggableBox(
    renderer,
    "wg-box-75",
    20,
    HEADER_HEIGHT + 5,
    25,
    8,
    RGBA.fromValues(255 / 255, 107 / 255, 129 / 255, 192 / 255),
    100,
  )
  root.add(box2)
  draggableBoxes.push(box2)

  const box3 = new DraggableBox(
    renderer,
    "wg-box-25",
    40,
    HEADER_HEIGHT + 3,
    25,
    8,
    RGBA.fromValues(139 / 255, 69 / 255, 193 / 255, 64 / 255),
    100,
  )
  root.add(box3)
  draggableBoxes.push(box3)

  const box4 = new DraggableBox(
    renderer,
    "wg-box-opaque",
    60,
    HEADER_HEIGHT + 7,
    25,
    8,
    RGBA.fromValues(30 / 255, 30 / 255, 42 / 255, 1.0),
    100,
  )
  root.add(box4)
  draggableBoxes.push(box4)

  renderer.keyInput.on("keypress", (key: KeyEvent) => {
    if (key.name === "d") {
      key.preventDefault()
      toggleScrim(renderer)
    }
  })

  renderer.on("resize", (width: number, height: number) => {
    const h = height - HEADER_HEIGHT
    background.width = width
    background.height = h
    if (scrim) {
      scrim.width = width
      scrim.height = h
    }
    renderer.requestRender()
  })
}

export function destroy(renderer: CliRenderer): void {
  renderer.clearFrameCallbacks()

  for (const box of draggableBoxes) {
    renderer.root.remove(box.id)
  }
  draggableBoxes = []
  nextZIndex = 101
  scrimVisible = false
  scrim = null
  headerDisplay = null

  renderer.root.remove("wg-overlay-root")
  renderer.setCursorPosition(0, 0, false)
}

if (import.meta.main) {
    exitOnCtrlC: true,
  })

  run(renderer)
  setupCommonDemoKeys(renderer)
  renderer.start()
}
```

aarcamp:
> My point is that technically it's technically not a bug — but from a ux point of view it's a regression. Another thing, if in this pr you have distortion on that draggable rect. You can see it in the video. That doesn't happen in main.

@simonklee Please check now. It's now a much larger change set (mostly due to added tests) but the code updates have been split 6 diffs for easier review. Note that I modified the wide-grapheme-overlay-demo.ts script a bit (to add a couple of bordered boxes for testing) and included a recording in the PR description.

> This suggests the next step would be to implement something like the preserve mask policy I proposed above.

FWIW I decided against this, so emojis never bleed through and are always replaced with `[]`.

> in this pr you have distortion on that draggable rect

This was one of the trickier parts, but it's fully working, and self-optimizes based on whether the box edge is touching a wide span or not (see updated description and screen recording).

ariane-emory:
I still got my fingers crossed that this, or some other solution, to the double-width emojis plus overlays problem can be merged. Great to see that you're still working on it!

aarcamp:
> I still got my fingers crossed that this, or some other solution, to the double-width emojis plus overlays problem can be merged. Great to see that you're still working on it!

@ariane-emory I wouldn't say I'm actively working on it, just clearing the conflicts I noticed today. Waiting for feedback so I can address any remaining issues. I may take another look at the full-unicode-demo if I find time next week.

aarcamp:
@simonklee Would you be able to provide a sense of whether this PR is more likely to be accepted or rejected? Unsurprisingly, I’ve already had to push updates 2-3 times to address conflicts, and now there’s yet another set to push through. Mostly I don’t want to waste my time maintaining the PR if it doesn’t meet the project’s needs. Thanks!

aarcamp:
@simonklee Sorry to ping again. This PR is now readjusted for your color tag changes.

It is becoming increasingly difficult to maintain 🤯 (irrelevant to any acceptance decision but does raise the likelihood that I'll lose interest in pursuing). FYI, I have been running it locally for over a month with no apparent regressions in Latin (I don't actively use CJK however).

FWIW I am also now shipping this patch with my own project. It's only at 34 stars but at least it does have some real-world use at this point.

If I can do anything to simplify the review, please let me know. It's already split into 6 logical per-commit chunks. I haven't tested them individually, so I don't know if it would make sense to ship them that way -- the separate commits are currently intended to simplify review only.

ariane-emory:
Still very eager to see this problem solved.

aarcamp:
> Still very eager to see this problem solved.

There is no community engagement at the moment, I'm maintaining this diff stack solely for the purpose of providing the fix in my own project while still being able to upgrade the OpenTUI dep. Maybe adding comments at https://github.com/anomalyco/opentui/issues/837 (I just now linked the PR to this) could help. Or if you could leave a code review and I'll take a look.

Issue #758, opened 2026-03-01
Unify stdin parsing: single parser for all escape sequence types


simonklee:
`stdinListener` in `renderer.ts` splits stdin data into two mutually exclusive paths before parsing:

```
stdin "data" event
  ├─ if handleMouseData(data) → return     [Path A: mouse]
  └─ else → _stdinBuffer.process(data)     [Path B: everything else]
```

This is an either/or dispatch at the chunk level. If a chunk starts with mouse sequences, the entire chunk is routed to Path A and any trailing non-mouse data (keyboard input, focus events, etc.) is silently dropped. If a chunk starts with non-mouse data, any mouse sequences later in the chunk go to Path B where `StdinBuffer` will split them out but they won't reach `MouseParser`.

This is the cause of #748, and other related issues when the OS joins different sequence types into the same `read()` buffer.

There are at least four independent parsers that don't coordinate:

1. `MouseParser.parseAllMouseEvents()`: parses SGR/X10 mouse from raw
   data, only reached via Path A
2. `StdinBuffer.extractCompleteSequences()`: generic escape sequence
   splitter that already understands CSI, OSC, DCS, APC, SS3, and even SGR mouse
   boundaries. Only reached via Path B.
3. `inputHandlers` chain: capability responses, focus, theme mode handlers
   that pattern-match on sequences after `StdinBuffer` emits them
4. `TerminalPalette`: registers its own `stdin.on("data")` listener
   directly on stdin, completely outside the normal flow

Related:

- https://github.com/anomalyco/opentui/pull/757
- https://github.com/anomalyco/opentui/pull/749

simonklee:
Closed by #770

PR #792, opened 2026-03-08
add regression for stdin.setEncoding

simonklee:
Just a demo for #758


simonklee:
https://github.com/anomalyco/opentui/pull/770

PR #793, opened 2026-03-08
Move 3D code to separate @opentui/three package

remorses:
Extracts `packages/core/src/3d/` into a standalone `@opentui/three` package at `packages/three/`.

- **New package** `packages/three/` with dedicated build/publish scripts and tsconfig
- **Examples remain in core on purpose** so there is a single example selector and all demos live in one place (`packages/core/src/examples/index.ts`)
- **Examples now import 3D APIs from `@opentui/three`** instead of core internal paths
- **Core exports cleaned up** by removing `src/3d.ts` and the `./3d` export
- **CI/release/publish flows updated** to build, validate, preview, and publish `@opentui/three`
- **Packaging fix** for core dist metadata so preview/release packing resolves workspace dependencies correctly


kommander:
@remorses I'll be checking this next as soon as I got the first version of plugins in opencode

PR #794, opened 2026-03-08
fix: preserve terminal's native cursor style by default

aarcamp:
The renderer's cursor style defaulted to .block, which emitted \x1b[2 q on every frame — overriding the user's terminal cursor preference (e.g. bar/line) even when no component explicitly set a cursor style. Add a .default variant to CursorStyle that emits \x1b[0 q (the ANSI "reset cursor style" sequence), and make it the initial state. Components that call setCursorPosition() without setCursorStyle() now preserve the terminal's native cursor.

https://github.com/user-attachments/assets/d7e388d7-d7f3-4e92-96ff-fe8e13da3df6

Standalone script used to produce the demo:

```
#!/usr/bin/env bun
/**
 * Minimal OpenTUI demo: cursor style default behavior.
 *
 * A custom renderable calls setCursorPosition() without setCursorStyle().
 * This shows what cursor style the renderer uses by default:
 *
 *   Before fix: block cursor (overrides terminal preference)
 *   After fix:  terminal's native cursor style (e.g. bar/line)
 *
 * Controls:
 *   Ctrl-C — quit
 *
 * Run:
 *   bun run scripts/cursor-style-demo.tsx
 */
import React from "react"
import {
  createCliRenderer,
  Renderable,
  type RenderContext,
  type RenderableOptions,
} from "@opentui/core"
import { createRoot, extend } from "@opentui/react"

// A renderable that shows a blinking cursor without setting any style.
// It only calls setCursorPosition — the cursor style comes entirely
// from the renderer's internal default.
class CursorProbe extends Renderable {
  constructor(ctx: RenderContext, options: RenderableOptions) {
    super(ctx, options)
  }

  protected renderSelf(): void {
    this.ctx.setCursorPosition(this.x + 1, this.y + 1, true)
  }
}

extend({ "cursor-probe": CursorProbe })

declare module "@opentui/react" {
  interface OpenTUIComponents {
    "cursor-probe": typeof CursorProbe
  }
}

// ---------------------------------------------------------------------------

  exitOnCtrlC: true,
  useAlternateScreen: true,
})

function App() {
  return (
    <box
      width={60}
      height={10}
      borderStyle="rounded"
      borderColor="cyan"
      title=" OpenTUI cursor default demo "
      titleColor="cyan"
      paddingLeft={1}
      paddingTop={1}
    >
      <box flexDirection="column">
        <text>
          {"This cursor's style is the renderer default.\n" +
           "No component calls setCursorStyle().\n\n" +
           "Before fix: always block\n" +
           "After fix:  terminal's native style"}
        </text>
        <cursor-probe width={1} height={1} />
      </box>
    </box>
  )
}

const root = createRoot(renderer)
root.render(<App />)
```

### DIFF
--- a/packages/web/src/content/docs/core-concepts/renderer.mdx
+++ b/packages/web/src/content/docs/core-concepts/renderer.mdx
@@ -158,9 +158,14 @@ Use these methods to control the cursor position and style:
 renderer.setCursorPosition(10, 5, true)
 
 // Cursor style
+//
+// Available styles: "default", "block", "underline", "line"
+// The default style is "default", which preserves the terminal's native cursor
+// style instead of overriding it.
 renderer.setCursorStyle({ style: "block", blinking: true }) // Blinking block
 renderer.setCursorStyle({ style: "underline", blinking: false }) // Steady underline
 renderer.setCursorStyle({ style: "line", blinking: true }) // Blinking line
+renderer.setCursorStyle({ style: "default" }) // Reset to terminal's native cursor
 
 // Cursor style with color
 renderer.setCursorStyle({


aarcamp:
> @kommander I looked at this and added a couple tests. I think it's okay. What do you think?

Thanks @simonklee abd @kommander ! 🫡

PR #801, opened 2026-03-11
paste: carry raw bytes instead of decoded text

simonklee:
PasteEvent carried a decoded, ANSI-stripped string which prevented
consumers from accessing the original bytes. Store a raw Uint8Array
instead and let each consumer (Textarea, Input) decide how to
decode and sanitize. This preserves binary paste content and enables
downstream libraries to implement their own understanding of the
clipboard data.

This is the refactor required for enabling us to implement the same thing https://github.com/anomalyco/opentui/pull/520 tried to add.

### DIFF
--- a/packages/core/src/console.test.ts
+++ b/packages/core/src/console.test.ts
@@ -1,6 +1,7 @@
 import { test, expect, describe, mock, beforeEach } from "bun:test"
 import { TerminalConsole, ConsolePosition } from "./console.js"
 import { MouseEvent } from "./renderer.js"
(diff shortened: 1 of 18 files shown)


packages/core/src/lib/KeyHandler.test.ts
```diff
@@ -9,6 +11,10 @@ function createKeyHandler(): InternalKeyHandler {
   return new InternalKeyHandler()
 }
 
+function pasteBytes(text: string): Uint8Array {
```
kommander: This should be in `testing` imo.

PR #803, opened 2026-03-12
fix: address link tracker clear decref imbalance

aarcamp:
LinkTracker tracked one pool ref per link ID, but clear() decremented once per cell reference count. For links repeated across many cells this could cause one valid decref followed by many failing decref calls during buffer clear.


aarcamp:
> Thanks for the PR. Can you add a test for this?

I thought about it but decided against. Would it even be useful? The only observable difference between the buggy and fixed code is whether decref is silently returning caught errors. All of the pool state is identical in both cases...

I can still write a test of some kind but would much appreciate your opinion on what it should check. :)

simonklee (approved):

PR #808, opened 2026-03-12
test(core): Kitty protocol coverage

simonklee:
Add Kitty protocol test suite.


packages/core/src/lib/parse.keypress-kitty.ts
```diff
@@ -151,9 +166,9 @@ const functionalKeyMap: Record<string, string> = {
   D: "left",
   H: "home",
   F: "end",
+  E: "clear",
   P: "f1",
   Q: "f2",
-  R: "f3",
```
simonklee, line 156: Yes, this is intentional.

Image

PR #811, opened 2026-03-13
feat(solid): expose terminal focus events via hooks

### LINKED ISSUE #766: Feature: Ability to detect when TUI's terminal window is out of focus
Something similar to web's `document.visibilityState` to detect when the terminal pane/window comes or goes out of focus, this could be helpful to run some side effects for which we need to know if user is actively looking at the TUI or not, for example, to send OS native notifications if user is out of focus or refetch some data when user comes back to the TUI.

Potential solution: maybe it can be detected using the FocusIn/FocusOut events emitted by most modern terminals

nybble73:
Closes #766

The renderer already emits `"focus"` and `"blur"` events when the terminal window gains/loses focus (via DECSET 1004), but there's no way to consume them from the SolidJS layer.

This adds three hooks following existing patterns:

- `onFocus(callback)` / `onBlur(callback)` — callback style, mirrors `onResize`
- `useTerminalFocus()` → `Accessor<boolean>` — signal style, mirrors `useTerminalDimensions`

Use case: opencode plugins need to know if the user is looking at the terminal to decide whether to send OS notifications. Currently done via osascript polling — this replaces it with the event-driven signal the terminal already provides.

### DIFF
--- a/packages/web/src/content/docs/bindings/solid.mdx
+++ b/packages/web/src/content/docs/bindings/solid.mdx
@@ -228,6 +228,40 @@ const App = () => {
 }
 ```
 
+### `onFocus(callback)`
+
+Run side effects when the terminal window gains focus.
+
+```tsx
+import { onFocus } from "@opentui/solid"
+
+const App = () => {
+  onFocus(() => {
+    console.log("Terminal focused")
+  })
+
+  return <text>Switch away and back to trigger focus events</text>
+}
+```
+
+### `onBlur(callback)`
+
+Run side effects when the terminal window loses focus.
+
+```tsx
+import { onBlur } from "@opentui/solid"
+
+const App = () => {
+  onBlur(() => {
+    console.log("Terminal blurred")
+  })
+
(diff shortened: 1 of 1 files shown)


packages/solid/src/elements/hooks.ts
```diff
@@ -105,6 +105,26 @@ export const usePaste = (callback: (event: PasteEvent) => void) => {
  */
 export const useKeyHandler = useKeyboard
 
+export const useTerminalFocus = () => {
+  const renderer = useRenderer()
+  const [focused, setFocused] = createSignal(true)
```
simonklee: createSignal(true) is a guess that can be wrong with no self-correcting mechanism until the next focus state change.

nybble73: Dropped `useTerminalFocus`; there does not seem to be a reliable way to query initial focus state here, so a state hook would need to be ternary. Kept `onFocus`/`onBlur` instead.

Issue #336, opened 2025-11-22
Weird artifacts due to emojis


veracioux:
I'm not sure how the behavior is best described. But it seems that the grapheme width is not being calculated correctly so the characters are not being overwritten, leaving artifacts.

https://github.com/user-attachments/assets/cbb359f3-5bb7-4e06-92c0-2ab71ea217e2

kommander:
Is that Alacritty?

veracioux:
@kommander Ah, indeed. It doesn't happen in kitty or ghostty.

Ghostty:
<img width="929" height="53" alt="Image" src="https://github.com/user-attachments/assets/79d33239-f487-40d9-8cfc-735c1512ae6a" />

Kitty:

<img width="854" height="48" alt="Image" src="https://github.com/user-attachments/assets/b52f4747-1a40-4777-9c25-d025d59bb882" />

kommander:
Yeah, Alacritty does not support ZWJ shaping. Opentui has a `no_zwj` mode, which is not properly working yet. And when tmux is in between it might shadow when Alacritty is used, so it's hard to tell.

simonklee:
I've ran our examples inside tmux with alacritty and i don't see this artifact anymore.

PR #823, opened 2026-03-17
fix(core): convert tree-sitter byte offsets to char indices for multi-byte text

graelo:
## Summary

- Fix text styling broken after emoji/CJK/multi-byte characters in `treeSitterToTextChunks()`
- Add `buildByteToCharMap()` to convert UTF-8 byte offsets from tree-sitter to JavaScript string indices

## Problem

Tree-sitter returns highlight ranges as UTF-8 byte offsets. `treeSitterToTextChunks()` uses these directly with `String.slice()`, which expects UTF-16 character indices. For ASCII text the two are identical, but emoji (e.g. ✅ = 3 bytes, 1 JS char) shifts all subsequent boundaries, leaving text after the first multi-byte character unstyled.

This causes white/invisible text on light terminal backgrounds (e.g. solarized light in tmux + Ghostty), since unstyled text inherits the terminal's default foreground.

## Fix

Convert byte offsets to character indices before use:

```typescript
const byteToChar = buildByteToCharMap(content)
const [startByte, endByte, , meta] = highlights[i]
const start = byteToChar(startByte)
const end = byteToChar(endByte)
```

## Note

`parser.worker.ts` has the same class of bug in `getNodeText()` (line 350) and injection offset arithmetic (lines 462-463), but those are separate code paths affecting injection highlighting. This PR focuses on the text styling fix.

## Related

- #336 (weird artifacts due to emojis)
- #609 (lineStarts byte offset mismatch with multi-byte text)

## Test plan

Verified locally by patching the bundled `@opentui/core` in an OpenCode build and testing with emoji-containing markdown (solarized light, tmux + Ghostty):

- [x] Render markdown containing emoji (e.g. `✅ Pre-Requisites`) — text after emoji retains syntax highlighting
- [x] ASCII-only text is unaffected
- [ ] CJK text highlighting (not tested locally)

### DIFF
--- a/packages/core/src/lib/tree-sitter-styled-text.ts
+++ b/packages/core/src/lib/tree-sitter-styled-text.ts
@@ -35,6 +35,30 @@ function shouldSuppressInInjection(group: string, meta: any): boolean {
   return group === "markup.raw.block"
 }
 
+/**
+ * Build a mapping from UTF-8 byte offsets to JavaScript string (UTF-16 code unit) indices.
+ * Tree-sitter returns byte offsets, but String.slice() needs character indices.
+ */
+function buildByteToCharMap(content: string): (byteOffset: number) => number {
+  const map = new Map<number, number>()
+  let charIdx = 0
+  let byteIdx = 0
+  for (charIdx = 0; charIdx < content.length; charIdx++) {
+    map.set(byteIdx, charIdx)
+    const code = content.codePointAt(charIdx)!
+    // Advance byte index by the UTF-8 byte length of this code point
+    if (code <= 0x7f) byteIdx += 1
+    else if (code <= 0x7ff) byteIdx += 2
+    else if (code <= 0xffff) byteIdx += 3
+    else {
+      byteIdx += 4
+      charIdx++ // Skip the second half of the surrogate pair
+    }
+  }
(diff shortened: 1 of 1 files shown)


graelo:
Using the `system` theme (for transparent background, in a light-background terminal), here's the output with the current code

Screenshot 2026-03-17 at 13 44 59

And with the above fix

Screenshot 2026-03-17 at 13 40 32

kommander:
The underlying issue for fenced code blocks using the wrong color is a different one I think, see #784. The byte offset change seems to accidentally fix that. 

Anyhow, the goal is to get rid of styled text and this method completely and instead setting highlights directly on the underlying native `TextBuffer`, which has a tree-sitter compatible string representation and we would not have to convert formats. 

Meanwhile, as `tree-sitter-wasm` is used in many JS/TS contexts, I think we might be missing a way to configure/use tree-sitter correctly in the `parser.worker.ts`, there must be an option to get the right offsets without having to do manual conversion.

graelo:
Thanks for the context! Setting the highlights directly on `TextBuffer` seems like the right solution indeed.

Happy to close my PR, no problem.

graelo:
It seems there's no path forward for this PR, I'm good with closing it.

PR #826, opened 2026-03-17
Fix markdown list hanging indent

lwlee2608:
Fix markdown list item continuation line indentation

### Problem
When a markdown list item's text wraps to the next line, the continuation lines start at column 0 instead of aligning with the text content after the list marker (e.g. 1. ). This makes wrapped list items hard to read, as the continuation text visually "falls back" to the left edge rather than staying grouped under its list item.

### Solution
Align continuation lines with the text that follows the list marker, so wrapped content stays visually connected to its list item.

### Why this matters

- Improves readability of long list items at a glance
- Makes the list structure clearer - each item reads as a cohesive block
- Better aesthetic consistency with how most markdown renderers display lists

**Before:**
image

**After:**
image

### DIFF
--- a/packages/core/src/3d/SpriteUtils.ts
+++ b/packages/core/src/3d/SpriteUtils.ts
@@ -51,8 +51,7 @@ export class SpriteUtils {
     const spriteTexture = await TextureUtils.fromFile(path)
 
     if (!spriteTexture) {
-      console.error("Failed to load sprite texture, exiting.")
-      process.exit(1)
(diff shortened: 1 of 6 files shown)


lwlee2608:
Hey @kommander @simonklee @msmps @Hona, just bumping this, mind taking a look when you get a chance?

The broken hanging indent on wrapped list items has been bugging me and this fix has been really useful. Would love to get it merged when you get a chance. Happy to make any changes if needed.

PR #827, opened 2026-03-17
chore: use enum instead of hardcoded strings

IdrisGit:
while working on #822 I saw that even though there was an Enum for render events, most events emitter were using hardcoded string values, this PR is just a cleanup to use the Enum instead.

Feel free to close this PR if it doesn't align with the project guidelines

### DIFF
--- a/packages/core/src/lib/clock.ts
+++ b/packages/core/src/lib/clock.ts
@@ -10,7 +10,11 @@ export interface Clock {
 
 export class SystemClock implements Clock {
   public now(): number {
-    return Date.now()
+    if (!globalThis.performance || typeof globalThis.performance.now !== "function") {
+      throw new Error("SystemClock requires globalThis.performance.now()")
+    }
+
+    return globalThis.performance.now()
   }
 
   public setTimeout(fn: () => void, delayMs: number): TimerHandle {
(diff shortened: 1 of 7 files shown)


kommander:
I am for using enums. Does it still allow using strings for consumers? Could we change the event types then to `CliRenderEvents[keyof CliRenderEvents]` or something so it can be either a string or the enum? Otherwise it would force all consumers to use the enum, which some users would deem not-ergonomic.

IdrisGit:
> Does it still allow using strings for consumers? 

I reviewed the code and yes it will allow consumers to use strings and enums both, however that's because the CliRendererEvent is not strictly typed (it extends to an any type), I haven't looked much but the EventEmitter has a generic type param, I think it might be better to make the Event strictly typed, which will help with TS inference and autocomplete on the consumers side.

I will update the PR.

simonklee:
> I am for using enums. Does it still allow using strings for consumers? Could we change the event types then to `CliRenderEvents[keyof CliRenderEvents]` or something so it can be either a string or the enum? Otherwise it would force all consumers to use the enum, which some users would deem not-ergonomic.

Yeh, this just replaces internal usage - it doesn't change types.

PR #828, opened 2026-03-18
Include static builds in releases

AnatoleLucet:
hi :wave: 

i updated the build scripts and release workflow to include static linked builds of the native renderer in the releases.
there's a release example on my fork: https://github.com/AnatoleLucet/opentui/releases/tag/v0.0.0-test

i did not change the name of the current dynamic archives (to avoid breaking anything if someone has automations around opentui releases), but if you'd prefer to have `opentui-native-{version}-{plateform}-{arch}-{dynamic/static}.zip` let me know


kommander:
The npm packages would only ever use the dynamic lib, so including static libs in the npm packages would bloat them for installs imo. I think static libs should only be github release artifacts available for download there.

AnatoleLucet:
true
i reverted the changes in the build script

simonklee:
Maybe i'm blind, but i don't see where we publish the gh static lib artifacts.

AnatoleLucet:
> Maybe i'm blind, but i don't see where we publish the gh static lib artifacts.

is in [this](https://github.com/AnatoleLucet/opentui/blob/2864ba66f128e6ebc4e2e20c93969a2f9b4d6aaf/.github/workflows/release.yml#L202-L222) for loop. the -static suffix is caugh by the `platform` variable (might need to rename or split into another var). then everything in release-assets/ is [added to the release](https://github.com/AnatoleLucet/opentui/blob/2864ba66f128e6ebc4e2e20c93969a2f9b4d6aaf/.github/workflows/release.yml#L247) :)

kommander:
It's the loop above https://github.com/AnatoleLucet/opentui/blob/2864ba66f128e6ebc4e2e20c93969a2f9b4d6aaf/.github/workflows/release.yml#L165-L196 but yes, it's there.

simonklee:
We are closing this pull request because the native source and release process
have changed. This version uses a removed source path, an old Zig version, and
a removed release workflow.

PR #833, opened 2026-03-19
fix(core): Markdown and selection rendering improvements

tmeire:
Fixes [opencode#16470](https://github.com/anomalyco/opencode/issues/16470 )

This PR improves markdown code blocks readability in light mode. 

The fist change in Markdown.ts forwards the style fg & bg color to the markdown renderables to make the text visible in normal rendering. Before (top) and after (bottom):

image

The second change in buffer.zig improves the rendering on selection. This change not only impacts markdown but the whole UI. When the regular background is transparent, the font defaults to black on selection, which is barely readable in many cases.

Selection Before:
select-before

Selection After:
select-after

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -456,9 +456,12 @@ export class MarkdownRenderable extends Renderable {
   }
 
   private applyMarkdownCodeRenderable(renderable: CodeRenderable, content: string, marginBottom: number): void {
+    const defaultStyle = this.getStyle("markup.raw") || this.getStyle("default")
     renderable.content = content
     renderable.filetype = "markdown"
     renderable.syntaxStyle = this._syntaxStyle
+    renderable.fg = defaultStyle?.fg
+    renderable.bg = defaultStyle?.bg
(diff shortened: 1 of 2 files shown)


simonklee:
Thanks for having a look at this. Looks like we did it concurrently. https://github.com/anomalyco/opentui/pull/832 I'm going to have have a look at your approach when i get the opportunity.

tmeire:
Your fix for the core markdown bug is a lot more extensive, that's for sure 😅
I added the selection part as well because I noticed the markdown was hardly readable when selected. Turns out a lot of UI part have this issue. Not sure if making the foreground color transparent introduces other unexpected behaviour, but from my testing, it didn't seem to.

simonklee:
> Your fix for the core markdown bug is a lot more extensive, that's for sure 😅
> I added the selection part as well because I noticed the markdown was hardly readable when selected. Turns out a lot of UI part have this issue. Not sure if making the foreground color transparent introduces other unexpected behaviour, but from my testing, it didn't seem to.

I think this is a separate issue that warrens it's own fix. I think your change break other things like transparent backgrounds and what not.

PR #835, opened 2026-03-19
feat(scrollbox): preserve scroll position across resize

benkoppe:
Hi! Thanks for your work with this repo. I use opencode in a workflow that causes it to get resized relatively often, and currently that causes large and unpredictable jumps in the scroll box, which can be jarring and very annoying.

This PR addresses the issue by snapshotting pre-resize state and restoring. The first visible child is used as an anchor, and `distanceFromBottom` is used as a fallback when anchor IDs aren't available. Here's the behavior before and after:

Before            |  After
:-------------------------:|:-------------------------:
<video src="https://github.com/user-attachments/assets/dfa066d5-2f94-45cf-97b1-0280601bfd48"> | <video src="https://github.com/user-attachments/assets/d205837e-327b-4edd-aa21-87e25f45c52c">

Regression tests are also included in a new file, `scrollbox-resize-preserve.test.ts`, which defines a `ScrollBoxResizeScenario` class ease test creation.


Arsh-S (approved):
LGTM!

benkoppe:
@kommander Wondering if I could bump this, it'd be great for my workflow 🥹

PR #839, opened 2026-03-20
fix(core): validate preserved overlay character width

### LINKED ISSUE #837: CJK text rendered underneath a popup/overlay gets hidden or overwritten

AbigailJixiangyuyu:
### Issue for this PR

Closes #837

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

- In packages/core/src/tests/absolute-positioning.snapshot.test.ts, it adds regression tests covering transparent and zero-sized absolute overlays over text like Hello 中文 world.
- In packages/core/src/zig/buffer.zig, it updates overlay preservation logic so underlying double-width characters are treated as valid preserved content, not just width-1 characters.

Tested locally with `bun test packages/core/src/tests/absolute-positioning.snapshot.test.ts --test-name-pattern "Overlay regression"`.

### Checklist


simonklee:
Thanks for attempting to fix this. I don't think this is the right approach, though.


https://github.com/user-attachments/assets/1053cf1e-1625-4523-a8c1-8fe884ade63d

AbigailJixiangyuyu:
Thanks for the feedback. I’ll revisit the issue and try to come up with a better fix.

AbigailJixiangyuyu:
When a dirty cell belongs to a wide grapheme continuation, redraw the full grapheme from its start cell instead of handling the continuation cell in isolation.

AbigailJixiangyuyu:
https://github.com/user-attachments/assets/2420e890-beb6-4934-9883-07c191d2de23

aarcamp:
Please see https://github.com/anomalyco/opentui/pull/791 for a different solution. Among other issues it solves the distortion problem around the overlay box edges which are seen in the last screen recording posted here. It's quite tricky because narrow and wide characters can share the same column of cells in the terminal, meaning to keep straight edges on the boxes, you need to reserve a 1-cell perimeter, and apply it on all 4 sides to keep things even. You also have to consider the case where the border attribute is enabled on the box, in which case you can use that as the perimeter instead. Further, you need to detect when the box edges don't touch any wide characters, so you can forego the perimeter when only narrow characters surround the box.

PR #841, opened 2026-03-20
test: failing test for long URLs broken by table column wrapping

nickveenhof:
## Problem

Long URLs in markdown table cells get split across multiple lines when the column is too narrow. The split breaks the URL: it is no longer a single clickable string, and the rendered output shows a truncated URL on one line with the remainder on the next.

Related: anomalyco/opencode#18394

## What this PR does

Adds a single failing test that reproduces the problem. No fix included.

## Test output

```
bun test src/renderables/__tests__/Markdown.test.ts --test-name-pattern "table with long URL"
```

The test expects the full URL to appear intact in the rendered table. Instead, the renderer splits it:

```
│Exam│https://github.com/charmbracelet/glamour/issues/     │
│ple │149#issuecomment-1234567890 (https://github.com/     │
│    │charmbracelet/glamour/issues/149#issuecomment-       │
│    │1234567890)                                          │
```

The 74-character URL is split at the column boundary. The "Name" column value "Example" is also split into "Exam" / "ple" across rows.

## Expected behavior

Long URLs should either:
1. Remain intact as a single OSC 8 hyperlink span across the line break
2. Be rendered as a footnote below the table
3. Truncate with an ellipsis while keeping the full URL in the link metadata


simonklee:
This test requires a 74-character URL to fit on one line in a 60-column
display. The URL already keeps its complete hyperlink target across wrapped
lines. Pull request #1259 fixed the separate short-column wrapping issue. This
test does not check hyperlink data and would fail when hyperlinks work
correctly.

PR #844, opened 2026-03-20
core: preserve terminal color intent

simonklee:
Flattening indexed and default colors into RGBA snapshots makes palette-relative roles lose meaning before the renderer sees them. Carry a color tag alongside RGBA through the TS, FFI, buffer, and native renderer, and publish the active terminal palette so the renderer can emit 39/49, keep 38;5 slots stable, and remap RGB fallback after palette changes instead of reusing stale indices


jahooma:
Awesome work, can't wait to see this merged. opentui is great!

PR #855, opened 2026-03-22
Switch tsconfig to NodeNext module resolution

simonklee:
Prevent regressions where extensionless imports break consumers using NodeNext
or Node's ESM loader.

tsconfigs override back to "bundler" because bun-ffi-structs and bun-webgpu
don't export NodeNext-compatible types yet.

Change this once @kommander fixes upstream. I think we can merge this pr now and update the last things once you've updated upstream.

Close: #843


inakineitor:
In case it's useful this should fix the upstream for bun-ffi-structs (https://github.com/anomalyco/bun-ffi-structs/pull/3) and bun-webgpu (https://github.com/kommander/bun-webgpu/pull/5). Happy to make changes or work more on those or these pull request if helpful!

PR #856, opened 2026-03-22
fmt changes

simonklee:
swap prettier with oxfmt/oxlint, update CI, and fix the formatting/lint issues needed for the new checks to pass.


simonklee:
> I'd check that treesitter still behaves fine in opencode with the intializer change, but other than that looks fine.

The problem with the async promise is that if it's rejected you will not catch it (async is an implicit promise, right). There is still an issue that i didn't touch that if it fails and you want to retry the failed promise is stored in `initializePromise` and never cleared.

PR #857, opened 2026-03-22
fix: renderer not destroyed on exit

### LINKED ISSUE #853: Random Artifacts after closing an opentui app
I am building a TUI app using opentui. After the app is closed, there seems to be random characters being input into the terminal. 

Image

Pressing `ctrl+c` or sending the command to reset the shell does nothing. The only thing that works is to kill the terminal. 

When the terminal tab closes, the terminal app does give me a warning that there is an ongoing process in the tab. Maybe there is something opentui failed to cleanup?

I see issues in the opencode github mentioning similar issues:
https://github.com/anomalyco/opencode/issues/5936
https://github.com/anomalyco/opencode/issues/6136
which I think could be related.

AnatoleLucet:
fixes #853

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -618,7 +618,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
 
     process.on("uncaughtException", this.handleError)
     process.on("unhandledRejection", this.handleError)
-    process.on("beforeExit", this.exitHandler)
+    process.on("exit", this.exitHandler)
 
     const kittyConfig = config.useKittyKeyboard ?? {}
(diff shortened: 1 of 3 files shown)


simonklee:
thanks for digging into this. You are right that beforeExit does not fire for process.exit(), so this reproducer is valid. However, this is intentional: opentui does not own explicit process shutdown, and applications are expected to call `renderer.destroy()` before exiting.

AnatoleLucet:
oh my bad. i thought you might wanted destroy to be implicit in ts since there already was a beforeExit listenner
i'll close this pr. feel free to reopen if we can get anything of it :)

simonklee:
No worries. We'll improve the docs around this.

Issue #805, opened 2026-03-12
Transparent backgrounds are composited against black instead of the terminal background


akronb:
Semi-transparent backgrounds rendered over transparent areas seem to be composited against black rather than the terminal’s actual background color. I first noticed this in `opencode` (https://github.com/anomalyco/opencode/issues/13363), but it seems to be related to `opentui`.

I made [a minimal repro](https://github.com/akronb/opentui-transparent-compositing-repro) that renders the same semi-transparent red overlay in two cases: over a solid white base and over a transparent background.

As you can see from the screenshot of my repro, when the background is transparent, it becomes black under the overlay for some reason.

<img width="1512" height="887" alt="Image" src="https://github.com/user-attachments/assets/267eb217-9538-4712-a6f4-57bb659bfd04" />

#### Environment
- `@opentui/core ^0.1.87`
- `bun 1.3.10`
- `macOS 15.7.4`
- `Ghostty 1.3.0`

kommander:
Check the core examples, the index has a transparent background, it's not an opentui issue. It used to work in opencode, someone must have messed that up again.

akronb:
In the transparency example, only the draggable blocks are transparent, the background behind them is solid.

https://github.com/anomalyco/opentui/blob/956f15c1332f426dc0ea5e1b40e5630b57722296/packages/core/src/examples/transparency-demo.ts#L93

If you change the color on this line to transparent, you’ll get this

<img width="764" height="387" alt="Image" src="https://github.com/user-attachments/assets/0f430fe3-df4b-4f12-b93b-c18bb7afaf25" />

kommander:
Yes, layering transparency onto the transparent terminal background is not possible. Only the lowest level of the background can be transparent like the terminal does. There is no sane cross terminal way to support layered transparency to the lowest level, not even in modern terminals.

kommander:
Oh I think I know what you mean now. These are two different things.

akronb:
Yeah, I meant that if the `opentui` app background is transparent, there is a problem with alpha blending in terminals that use light colors as background color, because it matches black as the base.

I know that it is possible to get the terminal background color with OSC11, and it's possible to add this as a default base for alpha blending if the background is transparent. But is that a good decision performance-wise? What to do with terminals that don't support OSC11, or users that use background gradients, images, etc.?

I just want to figure out if this is possible and even worth it.

kommander:
I think I had that fixed in some playground branch, let me try to dig it up.

[...]

PR #865, opened 2026-03-23
perf(native): skip transparent draws

AnatoleLucet:
hi :wave: 

while playing with the zig renderer i noticed transparent boxes were pretty slow to draw (mainly because there's no check to skip invisible draws, and the renderer tries to blend the colors anyway).
i figured that might be a slight performance hit for apps using a lot of transparent boxes for layout/positioning purposes.

so i added a simple check to skip invisible draws (opacity of 0, or rgba with alpha of 0), and added a few benchmarks to make it easier to debug and see the improvement.

<details>
<summary>without transparency check (current)</summary>
image
</details>

<details>
<summary>with transparency check (new)</summary>
image

> notice how `bg alpha 0` (bg alpha 0 + borders alpha 1) is a bit slower than, `fully opaque` (bg alpha 1 + borders alpha 1) because even though we dont try to fill+blend  the whole box anymore when the background is transparent, blending is still required for borders background, making it a bit slower than having a fully opaque background (can vary depending on the box size). but that's already way faster than trying to blend everything

</details>

let me know what you think!
i might try and improve color blending after if that's fine


AnatoleLucet:
also just noticed this fixes a bug were the following would render as `     w` because `above` would override `bellow` even with opacity 0, which is probably a bit counter intuitive.
```ts
const rdr = await createCliRenderer();

const t1 = new TextRenderable(rdr, { content: "bellow" });
rdr.root.add(t1);

const t2 = new TextRenderable(rdr, {
  content: "above",
  position: "absolute",
  opacity: 0,
});
rdr.root.add(t2);
```
but the behavior might not be fully correct now because invisible text is not possible anymore (it just wont be rendered at all instead of being invisible), so it can't be copied for instance.

what the preference here? should i remove the transparency conditions on text draws and only keep it for `fillRect()`, or we go with opacity 0 = not rendered?

simonklee:
I would play around with this example. It's sort of hard to know from looking at code what is changing and trade-offs at play.
 https://github.com/anomalyco/opentui/blob/main/packages/core/src/examples/wide-grapheme-overlay-demo.ts

AnatoleLucet:
the wide grapheme example seem to work the same on my branche and on main (left is main, right is my branch):
image

apart from the perf gain, the changes in `buffer.zig` from this pr changes a few things (in its current state, happy to make it behave differently if needed) :

with the following example:
```ts
const rdr = await createCliRenderer();

const container = new BoxRenderable(rdr, {
  backgroundColor: "blue",
  flexDirection: "row",
});
rdr.root.add(container);

container.add(new TextRenderable(rdr, { content: "invisible text ->" }));
container.add(new TextRenderable(rdr, { content: "hello world", opacity: 0 }));
container.add(new TextRenderable(rdr, { content: "<- invisible text" }));
```

<details><summary>video (left is main, right is this pr)</summary>
<video src="https://github.com/user-attachments/assets/a32645bd-c6c6-48f0-9498-3d6b9710f46f"></video>
</details>

on main:
- `hello world` is rendered, but with a fg and bg matching the container's background color to make it appear as invisible (or black if no container has a background. maybe because of #805)
- the invisible text can be selected and copied with term selection (but cannot with opentui's selection)
- it overrides any piece of text that was rendered bellow the invisible text (shown on the screenshot bellow)

on this pr:
- `hello world` is not rendered at all.
- it cannot be copied with term selection
- but it also does not override underlying text (screenshot bellow)

<details><summary>text override (left is main, right is this pr)</summary>

```ts
const rdr = await createCliRenderer();

const container = new BoxRenderable(rdr, { backgroundColor: "blue", });
rdr.root.add(container);

// rendering text bellow
container.add(new TextRenderable(rdr, { content: "lorem ipsum dolor sit amet consectetur adipiscing elit", }));

// display text above the previous text
const abs = new BoxRenderable(rdr, {
  height: 1,
  width: "100%",
  position: "absolute",
  flexDirection: "row",
});
container.add(abs);

abs.add(new TextRenderable(rdr, { content: "invisible text ->" }));
abs.add(new TextRenderable(rdr, { content: "hello world", opacity: 0 })); // hello world is invisible
abs.add(new TextRenderable(rdr, { content: "<- invisible text" }));
```

image

main overrides the lorem ipsum bellow, even in the section where the text has an opacity of 0.

<details><summary>web behavior just for reference</summary> 
image
</details>

this is also the case for borders:
```ts
const rdr = await createCliRenderer();

const container = new BoxRenderable(rdr, { backgroundColor: "blue" });
rdr.root.add(container);

// rendering text bellow
container.add(new TextRenderable(rdr, { content: "lorem ipsum dolor sit amet consectetur adipiscing elit" }));

// invisible box with borders above
const abs = new BoxRenderable(rdr, {
  height: 4,
  width: 10,
  position: "absolute",
  border: true,
  opacity: 0,
});
container.add(abs);
```

image

on main (left) the border is black when there's no underlying background, and blue when there is (overriding the text).
while on this pr (right), the border is not visible at all, and does not override the text.

</details>

hope this makes more sens :)

if the behavior in this pr seems correct to you i'll be happy to add tests for it!

PR #866, opened 2026-03-23
fix(markdown): hide URL suffix for concealed links

brendandebeasi:
## Summary

- Remove redundant ` (url)` suffix from concealed markdown links — the link text is already an OSC 8 hyperlink (Cmd/Ctrl+click), so displaying the raw URL is unnecessary
- Update two existing tests to match the new concealed link output

## Context

When `conceal` is enabled, markdown links `[text](url)` render as `text (https://...)`. Since the link text already carries the `linkHref` for OSC 8 terminal hyperlinks, the parenthesized URL is redundant visual noise. This change makes concealed links render as just `text` — clickable via Cmd/Ctrl+click in supporting terminals.

The non-concealed (`conceal=false`) path is unchanged and still shows the full `[text](url)` syntax.

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -386,9 +386,6 @@ export class MarkdownRenderable extends Renderable {
           for (const child of token.tokens) {
             this.renderInlineTokenWithStyle(child as MarkedToken, chunks, "markup.link.label", linkHref)
           }
-          chunks.push(this.createChunk(" (", "markup.link", linkHref))
-          chunks.push(this.createChunk(token.href, "markup.link.url", linkHref))
-          chunks.push(this.createChunk(")", "markup.link", linkHref))
         } else {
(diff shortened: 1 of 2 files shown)


AnatoleLucet:
maybe this could be an option instead of the default?

for instance a new `concealUrl: boolean` options similar to [`concealCode: boolean`](https://github.com/brendandebeasi/opentui/blob/ed602f494f3d434c621b01db6f3aab2418ece76f/packages/core/src/renderables/Markdown.ts#L72):

- `concealUrl: true` -> `<label>`
- `concealUrl: false` -> `<label> (<url>)`

would still allow links to work on terminals that don't support OSC8 like today when `false`, and it wouldn't be a breaking change either :)

simonklee:
This change does not hide URLs in Markdown paragraphs, and it breaks Markdown
table tests. It also removes visible URLs on terminals that do not support OSC
8 links.

PR #870, opened 2026-03-24
perf(native): faster color blending

AnatoleLucet:
> ~~follow up to #865  **(only the last commit is relevant)**~~

this pr makes a small change to the color blending function (roughly 4x faster now) which improves prefs for draw calls with translucent colors and for postprocessing draws (`drawSuperSampleBuffer`, `drawGrayscaleBuffer`, etc).

i replaced the use of `std.math.pow` with a faster [approximate pow](https://github.com/anomalyco/opentui/commit/65637b796055534f498d7c51c9e3808b0b0e56ca#diff-600539073a348cb8d31e35f59420f85cab51b5f319cebd773f373226b2b5330fR107-R112).
the approximate pow is a not 100% accurate so the result does not fully match `std.math.pow` but for color blending the change is (very) hardly noticeable.

<details><summary>benchmark (left with <code>std.math.pow</code>, right with approximate pow)</summary>
image
(running on my laptop so take it with a grain of salt)
</details>

<details><summary>color comparison (left with <code>std.math.pow</code>, right with approximate pow)</summary>
<video src="https://github.com/user-attachments/assets/3e47a67f-03a7-41cf-a2e8-0f60e98c2325"></video>
</details>


simonklee:
This would replace #498, since you make the same change. in general i would try to limit this to one-idea per pr. You both add a bunch of early outs + replace the pow in this one.

simonklee:
I've triggered a run for your branch at https://opentui-bench.fly.dev/ ~30 min and it should show up there.

AnatoleLucet:
> This would replace #498, since you make the same change. in general i would try to limit this to one-idea per pr. You both add a bunch of early outs + replace the pow in this one.

yeah sorry i hope the fact it's a follow up of #865 doesn't make it harder to review. i just preferred to open a separate pr since color blending felt unrelated. i'll add the opacity benchmark in #865 instead so both pr can be merged separately. but we won't have the benchmarks on this branch.

i'll remove the condition that's already covered by #498 :) did you stumble on anything that stopped you from merging that pr, or was it just forgotten?

AnatoleLucet:
i added a separate benchmark here for color blending / translucency, and moved the box opacity benchmark to #865. hope it's clearer now, sorry about that

PR #877, opened 2026-03-26
fix(native): terminal not updated on new capabilities

AnatoleLucet:
When rendering right after creating the renderer, features that depend on special capabilities (e.g. hyperlinks) won't work right away because the capability response has not been processes/received yet.
It results in a terminal output were some features won't work until their cells are updated.

For instance:
```ts
const rdr = await createCliRenderer();

const l = new TextRenderable(rdr, {
  content: t`a link: ${link("https://google.com")(underline("Click Me!"))}`,
});
rdr.root.add(l);
```

If we render a simple link right after creating the renderer (on a terminal that supports OSC8), the link won't work until these cells are re-rendered by the renderer, because the initial render did not send the OSC8.

To fix it, i added a forced render after processing new capabilities to make sure every feature works correctly with the latest capabilities.

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -393,6 +393,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
   private minTargetFrameTime: number = 1000 / this.maxFps
   private immediateRerenderRequested: boolean = false
   private updateScheduled: boolean = false
+  private forceNextRender: boolean = false
 
   private liveRequestCounter: number = 0
   private _controlState: RendererControlState = RendererControlState.IDLE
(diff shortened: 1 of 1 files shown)


AnatoleLucet:
made a little update to the pr because i realized my previous fix in the zig renderer could produce half-baked renders if stdin is processed mid-draw.

<details><summary>demo of the issue with my hold fix</summary>
<video src="https://github.com/user-attachments/assets/681144d3-f929-4ab3-88db-58e3f5dc4dd0"></video>
</details>

this is now fixed. i moved the "re-render when receiving capabilities" logic to typescript so it can be properly scheduled with the rest.

simonklee:
Pull request #844 already added a forced repaint after terminal capability
responses. The main branch also includes a regression test for this behavior,
so this change is no longer necessary.

PR #878, opened 2026-03-26
fix(native): xtversion in tmux

AnatoleLucet:
currently `hyperlinks` need the terminal name to know if they are supported or not.
and the terminal name is either found using xtversion or `TERM_PROGRAM=`.

when running under tmux, the xtversion simply returns `tmux`, it does not return the real terminal's xtversion.
and `TERM_PROGRAM=` will more often than not be set to `*-256color`, which does not give use the terminal name either.

this will result in `hyperlinks` being detected as unsupported even though the user is running in an OSC8 compliant terminal.

> currently `hyperlinks` are also predicted as being supported if `COLORTERM` is set to `truecolor` or `24bit`, which can be false in a few environment, especially over ssh. i'll open a separate pr for that if that's fine

to fix this, i added an extra DSC wrapped `xtversion` query to make sure we get the proper terminal name under tmux.
i also cleaned the logic a bit around `is_tmux` because it was inconsistent.

<details><summary>[main] left kitty w/ tmux, right kitty w/o tmux</summary>
<video src="https://github.com/user-attachments/assets/c30f0132-29db-46ed-93f3-ab87a5310ccd"></video>
hyperlinks are wrongly detected as unsupported under tmux.
</details>

<details><summary>[this pr] left kitty w/ tmux, right kitty w/o tmux</summary>
<video src="https://github.com/user-attachments/assets/6b05de14-752e-427e-ba91-4c313657ce98"></video>
hyperlinks are correctly detected as supported under tmux.
</details>

### DIFF
--- a/packages/core/src/zig/ansi.zig
+++ b/packages/core/src/zig/ansi.zig
@@ -126,6 +126,7 @@ pub const ANSI = struct {
         }
     }
 
+    pub const xtversionTmux = wrapForTmux(xtversion);
     pub const kittyGraphicsQueryTmux = wrapForTmux(kittyGraphicsQuery);
     pub const capabilityQueriesTmux = wrapForTmux(capabilityQueriesBase) ++ csiUQuery;
     pub const sixelGeometryQuery = "\x1b[?2;1;0S";
--- a/packages/core/src/zig/terminal.zig
+++ b/packages/core/src/zig/terminal.zig
@@ -104,6 +104,7 @@ host_env_map: ?std.process.EnvMap = null,
 in_tmux: bool = false,
 skip_graphics_query: bool = false,
 skip_explicit_width_query: bool = false,
+xtversion_query_pending: bool = false,
 graphics_query_pending: bool = false,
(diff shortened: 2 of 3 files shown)


simonklee:
This breaks the tmux-via-xtversion flow the PR is trying to fix.

After the first unwrapped xtversion response, term_info is "tmux", but self.in_tmux has not been updated yet when sendPendingQueries() runs. With the new self.in_tmux checks, the wrapped retry xtversion / capability / graphics queries are skipped in that case. The old self.in_tmux or self.isXtversionTmux() logic handled this.

The test was also changed to set term.in_tmux = true up front, so it no longer covers the real transition.

AnatoleLucet:
ah! indeed that's my mistake.
i reverted that part. it should me good now!

simonklee:
The current main branch places the native code in `packages/native` and uses a different multiplexer model. This pull request no longer applies to the current code. The tmux hyperlink issue still needs a fix that keeps both multiplexer and terminal identities.

PR #888, opened 2026-03-28
fix(core): markdown link labels not clickable

AnatoleLucet:
fixes #869 

markdown link labels weren't clickable because treesitter reports `concealed` highlights instead of `markup.link` for symbols around the url and label (`[...](...)`). the `detectLinks()` fn wasn't taking that into account.


simonklee:
Can you make this test use the real highlight format tree-sitter returns here? Right now it uses synthetic concealed ranges, but markdown conceal output is conceal with `meta.conceal`.

AnatoleLucet:
i think its just `conceal` without `meta.` in this case: https://github.com/anomalyco/opentui/blob/main/packages/core/src/lib/tree-sitter/assets/markdown_inline/highlights.scm#L33-L41

but i'll update the tests to use a real treesitter output so we're sure it works and wont break again :)

AnatoleLucet:
there's a separate issue on `main`.
one of the test expect [`string.special.url`](https://github.com/anomalyco/opentui/blob/main/packages/core/src/lib/detect-links.test.ts#L29-L40) to be a thing for urls in code comments, but it doesn't really exist. the only instance of this capture i could find is for [require imports](https://github.com/anomalyco/opentui/blob/main/packages/core/src/lib/tree-sitter/assets/typescript/highlights.scm#L401-L402) but that's unrelated.

since this probably never worked, should i remove the test?
or i could add a special case for `comment` to detect links with a regex or something, but we're [probably not allowed](https://regexlicensing.org/) (and for good reasons).

i pushed the failing test but the rest is good. let me know how you want to proceed :)

simonklee:
No, i think the fix was good.

AnatoleLucet:
> there's a separate issue on `main`.
> one of the test expect [`string.special.url`](https://github.com/anomalyco/opentui/blob/main/packages/core/src/lib/detect-links.test.ts#L29-L40) to be a thing for urls in code comments, but it doesn't really exist.

relevant log in ci: https://github.com/anomalyco/opentui/actions/runs/23841810400/job/69500558649?pr=888#step:10:2403

PR #898, opened 2026-03-31
fix(markdown): prevent hr tokens from causing frontmatter misparse

shaofengluo-ctrl:
## Summary

- Fix markdown rendering bug where content after `table + ---` (thematic break) renders as monospace/unstyled text instead of formatted markdown

## Problem

`buildRenderableTokens()` groups consecutive non-table/code/blockquote tokens into a single `CodeRenderable` block. When a table is followed by `---` (hr), the grouped block starts with `---`, which Tree-sitter's markdown parser interprets as YAML frontmatter (`minus_metadata`). Everything between two `---` delimiters is then rendered as metadata instead of markdown.

**Reproduction**: Any markdown with this pattern:
```
| col1 | col2 |
|------|------|
| a    | b    |

---

## Heading

1. **Bold item**: description
   - nested list
```

The heading may render correctly, but the ordered list and nested content render as green monospace (metadata style).

## Fix

Add `"hr"` to `shouldRenderSeparately()` so thematic breaks are rendered as independent blocks, preventing them from starting a markdown group that Tree-sitter misparses as frontmatter.

```diff
  private shouldRenderSeparately(token: MarkedToken): boolean {
-   return token.type === "code" || token.type === "table" || token.type === "blockquote"
+   return token.type === "code" || token.type === "table" || token.type === "blockquote" || token.type === "hr"
  }
```

## Impact

- `hr` tokens now render as separate `CodeRenderable` blocks (Tree-sitter correctly parses standalone `---` as `thematic_break`)
- Subsequent content starts with headings/paragraphs instead of `---`, avoiding frontmatter misinterpretation
- `getInterBlockMargin` returns 1 for `hr` (since it's now in `shouldRenderSeparately`), adding natural spacing after horizontal rules

## Discovered in

opencode TUI (opencode-ai v1.2.25) — PRD review output with tables followed by `---` separators and ordered lists.


kommander:
Could you add a test for this to make sure it behaves as expected and doesn't regress?

shaofengluo-ctrl:
Added tests as requested:

- `hr after table does not cause subsequent content to render as frontmatter` — the core bug scenario (table + hr + paragraph)
- `hr between paragraphs renders content on both sides correctly` — bold/italic preserved across hr
- `multiple hrs do not break subsequent markdown formatting` — chained table + hr + bold

Two existing inline snapshots updated: hr now renders separately so there's one less blank line before it — expected side effect of the fix.

All 97 tests passing.

simonklee:
PR #1043 already added the same horizontal-rule fix to main. Horizontal rules
now render as separate structural blocks. The tests in this PR also pass
without its code change.

PR #907, opened 2026-04-01
fix(renderer): handle SIGTSTP to prevent mouse garbling on external suspend

agutmanstein-scale:
### Issue for this PR

Closes #906

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

When a process receives SIGTSTP from an external source (job control, terminal multiplexer, code-server), the process is suspended by the kernel without any terminal cleanup. Mouse tracking and raw mode remain active, causing mouse events to echo as garbled text in the shell.

This PR adds SIGTSTP/SIGCONT signal handlers to the renderer that call `suspend()`/`resume()` — the same cleanup that the manual suspend keybind uses.

**Implementation:**
- `sigtstpHandler`: calls `suspend()` to disable mouse/raw mode, removes itself, re-raises SIGTSTP for default suspend behavior
- `sigcontHandler`: re-registers the SIGTSTP handler, calls `resume()` to restore terminal state
- `addSigtstpListeners()`/`removeSigtstpListeners()`: symmetric lifecycle management matching `addExitListeners()`/`removeExitListeners()`
- Registered in constructor, removed in `suspend()`, `cleanupBeforeDestroy()`; re-added in `resume()`

- Code review: the handlers delegate to the existing `suspend()`/`resume()` which are already well-tested
- The SIGTSTP handler correctly removes itself before re-raising (avoids infinite loop)
- Lifecycle is symmetric with exit listeners

### Reproduction without fix

```bash
# Terminal 1: run any opentui app with mouse tracking
opencode

# Terminal 2: send SIGTSTP
kill -TSTP $(pgrep -f opencode)

# Terminal 1: move mouse → garbled escape sequences
```

### Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -491,6 +491,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
   private exitSignals: NodeJS.Signals[]
   private _exitListenersAdded: boolean = false
   private _sigtstpListenersAdded: boolean = false
+  private _suspendedBySigtstp: boolean = false
   private _isDestroyed: boolean = false
   private _destroyPending: boolean = false
   private _destroyFinalized: boolean = false
@@ -844,6 +845,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
   }
 
   private sigtstpHandler = (): void => {
+    this._suspendedBySigtstp = true
     this.suspend()
     // Remove handler to allow default SIGTSTP behavior (suspend process)
     process.removeListener("SIGTSTP", this.sigtstpHandler)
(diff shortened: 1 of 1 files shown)


kommander:
Opencode is currently handling suspend/resume manually. This needs to be aligned.

agutmanstein-scale:
There are scenarios, like a user alt-tabbing, that can cause the TUI to receive a SIGTSTP. This is designed to address external scenarios. Resume cannot be triggered in the same terminal window because opentui is still ingesting mouse input @kommander

PR #914, opened 2026-04-03
fix: pressTab({ shift: true }) sends correct ANSI back-tab sequence

LeTamanoir:
## Summary

- Fix `pressTab({ shift: true })` in standard ANSI mode to emit `\x1b[Z` (back-tab / CBT) instead of silently dropping the shift modifier and sending raw `\t`
- Add round-trip test verifying the emitted sequence parses back as `{ name: "tab", shift: true }`
- Add regression test confirming unmodified `pressTab()` still sends `\t`

## Test plan

- [x] All 115 existing mock-keys tests pass
- [x] New round-trip test: `pressTab({ shift: true })` output is parsed by `parseKeypress` as shift+tab
- [x] Regression test: `pressTab()` without modifiers still sends `\t`

### DIFF
--- a/packages/core/src/testing/mock-keys.test.ts
+++ b/packages/core/src/testing/mock-keys.test.ts
@@ -381,6 +381,28 @@ describe("mock-keys", () => {
 
     mockKeys.pressTab({ shift: true })
 
+    expect(mockRenderer.getEmittedData()).toBe("\x1b[Z")
+  })
+
+  test("pressTab with shift modifier parses as shift+tab", async () => {
+    const { parseKeypress } = await import("../lib/parse.keypress")
+    const mockRenderer = new MockRenderer()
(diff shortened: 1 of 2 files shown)


simonklee:
you're just changing testing code paths in this PR

LeTamanoir:
Thanks for the review! Just wanted to add some context. I was using `createMockKeys` from `opentui/testing` in my project to test tab navigation, and `pressTab({ shift: true })` was sending a raw `\t` instead of `\x1b[Z`, so my tests for Shift+Tab (back-tab) behavior were silently passing with the wrong input. Since `createMockKeys` is exported as part of the public API, I thought it would be worth fixing upstream rather than working around it on my side. Happy to hear your thoughts if there's a better approach!

simonklee (approved):

PR #918, opened 2026-04-05
fix: use new URL() instead of import with { type: file } for cross-runtime compat

lu-zero:
The Bun-specific `import ... with { type: "file" }` syntax is not supported by Deno or Node. Replace with `new URL(path, import.meta.url)` + `fileURLToPath()` which works identically across all runtimes.

Update the generator (assets/update.ts) to produce cross-runtime compatible code so future regeneration preserves this change.


kommander:
I see the issue. This weird generation was to support Bun standalone executables bundling the existing default parsers and queries I think. Need to check if that still works this way for opencode.

simonklee (approved):
Tested with opencode and for me the highlights still work.

PR #928, opened 2026-04-06
fix: full repaint on resume to fix transparency artifacts in Kitty

### LINKED ISSUE #922: Transparency artifacts after `Ctrl+Z` suspend/resume in terminals with background opacity
## Description: 

When suspending OpenCode with `Ctrl+Z` and resuming with `fg`, the TUI displays visual artifacts, specifically transparency bleeding through cell backgrounds in terminals like kitty that have background transparency enabled.

## Steps to Reproduce
 
1. Set `background_opacity` below `1.0` in `kitty.conf`
2. Run any OpenTUI app in Kitty
3. Press `Ctrl+Z` to suspend
4. Run `fg` to resume

## Expected Behavior
 
TUI renders identically after resuming with no visual changes.

## Environment
 
- Terminal: Kitty with `background_opacity 0.85`
- OS: Arch linux
- Shell: zsh

### note:

resizing fixes it.

acegikmoo:
Fixes #922

When Kitty resumes after `SIGCONT`, it restores the alternate screen buffer with transparency compositing applied. The diff render then skips cells that haven't changed, so the transparency persists since nothing in the UI actually changed during suspension.

The fix is to call `lib.render` with `force=true` right after `currentRenderBuffer.clear()` in `resume()`. This bypasses the diff and repaints every cell immediately, overwriting whatever Kitty had in its restored buffer.

This is also why resizing fixes it, `buffer.resize()` clears both buffers which makes the diff see everything as dirty and forces a full repaint as a side effect.

```ts
this.currentRenderBuffer.clear(this.backgroundColor)
// Normal diff skips unchanged cells leaving transparency visible. force=true repaints all.
this.lib.render(this.rendererPtr, true)
```

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -2111,6 +2111,10 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     }
 
     this.currentRenderBuffer.clear(this.backgroundColor)
+
+    // Normal diff skips unchanged cells leaving transparency visible. force=true repaints all.
+    this.lib.render(this.rendererPtr, true)
+
     this._controlState = this._previousControlState
 
     if (
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -2110,8 +2110,6 @@ export class CliRenderer extends EventEmitter implements RenderContext {
       this.enableMouse()
     }
 
-    this.currentRenderBuffer.clear(this.backgroundColor)
-
     // Normal diff skips unchanged cells leaving transparency visible. force=true repaints all.
     this.lib.render(this.rendererPtr, true)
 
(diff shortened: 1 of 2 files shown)


kommander:
`this.currentRenderBuffer.clear(this.backgroundColor)` was supposed to do that, but I guess behaviour might have changed recently when `this.backgroundColor` has `alpha=0` it does not reset the cells. When we do a force re-render it will draw all cells and the `currentRenderBuffer.clear` becomes futile, so we would not need to do both. 

I wonder if that is easily coverable with a test maybe, so it doesn't regress again 🤔

acegikmoo:
> `this.currentRenderBuffer.clear(this.backgroundColor)` was supposed to do that, but I guess behaviour might have changed recently when `this.backgroundColor` has `alpha=0` it does not reset the cells. When we do a force re-render it will draw all cells and the `currentRenderBuffer.clear` becomes futile, so we would not need to do both.
> 
> I wonder if that is easily coverable with a test maybe, so it doesn't regress again 🤔

removed the `clear()` call since the force render covers it. Also added a test that spies on `lib.render` and asserts `force=true` is passed on resume, so this should be covered against regression ig

Issue #929, opened 2026-04-06
Feature: Hierarchy focus


garthdw:
Currently when creating a layout that is

```
root
	- Box (focusable)
		- Input
	- Box (focusable)
		- Text
		- Input
```

When the input becomes focused, the box loses any indication that it is part of the focus chain.

I have done some exploration into an opt in ancestor focusing implementation. Where we focus each parent that is focusable when focusing on an input. I made it opt in, because it would be a breaking UI change.

https://github.com/user-attachments/assets/b78cc880-1f0a-4ff1-a8eb-e62f025e166b

PR #930, opened 2026-04-06
feat(core): allow focusing of ancestors of renderables

garthdw:
Only focusable ancestors are focused.

Opt in behaviour at the renderer level and needing to mark direct ancestors as focusable.

https://github.com/user-attachments/assets/61e8028c-4b55-4487-a726-937727041b84

For #929


garthdw:
Dammit, nothing worse than that quick code change before a PR to break tests and then make you end up with more brutal code. :/

kommander:
Interesting. I think that makes sense. I don't know how opentui users currently use it.  The renderer knows only about exactly one focused renderable. It could however be an additional tree attribute that is propagated? Wonder how the browser and apps there handle that.

If you want a box around a Textarea to react to the Textarea focus, it can subscribe to that state.

I consider only Renderables that handle key events to be focusable. So style/ui only elements would not set focusable=true on themselves.

garthdw:
So looks like html has a `:focus-within` or at least firefox does.

https://github.com/user-attachments/assets/806aed3c-fe3c-4ca0-b802-01ae2c0383ee

I'll explore some alternatives to how to represent this behaviour.

garthdw:
I have implemented it a different way.

I'm toying between doing it as just `focusable` on parent elements or if we should add a separate property, I just can't think of a good name for it.

packages/core/src/Renderable.ts
```diff
@@ -407,12 +408,25 @@ export abstract class Renderable extends BaseRenderable {
...
+    this.propagateFocusChange(true)
     this.emit(RenderableEvents.FOCUSED)
   }
 
+  private propagateFocusChange(hasFocus: boolean): void {
+    let parent = this.parent
+    while (parent) {
+      if (parent._hasFocusedDescendant !== hasFocus) {
+        parent._hasFocusedDescendant = hasFocus
+        parent.requestRender()
```
kommander: It should be enough to call this once after the while loop, as multiple calls will just be ignored anyway when a render was already requested.

packages/core/src/Renderable.ts
```diff
@@ -407,12 +408,25 @@ export abstract class Renderable extends BaseRenderable {
...
+      }
+      parent = parent.parent
+    }
+  }
+
   public blur(): void {
     if (!this._focused || !this._focusable) return
 
+    this._ctx.blurRenderable(this)
```
kommander, line 431: I realised earlier this week that `renderer.focusedRenderable` basically never gets reset. It's also not reset when the renderable is destroyed. So in the renderable destroy method it should probably be blurred as well. Would be cool to get this new blur behaviour for renderer level focus/blur test covered.

packages/core/src/Renderable.ts
garthdw: I was also cautious about the multiple calls. The benefit we get is the mark as dirty, but if we don't need that and can't foresee any other issues, I'm happy to move it out of the loop and call it on itself.

packages/core/src/Renderable.ts
garthdw: So "luckily" this change will help with the destroy. It calls `blur` when destroying.

Let me add test coverage though.

PR #932, opened 2026-04-07
feat(core): support ANSI 16 colors for borderColor and backgroundColor

### LINKED ISSUE #852: Support ANSI 16 color codes for borderColor and backgroundColor
## Problem

Currently, `borderColor` and `backgroundColor` on `BoxRenderable` are rendered using true color escape sequences (`\e[38;2;r;g;bm`). This means the colors are absolute RGB values that don't adapt to the user's terminal color scheme.

ANSI 16 color codes (`\e[31m`, `\e[41m`, etc.) are mapped by the terminal emulator to its configured palette, so an app can automatically adapt to the user's theme — especially important over SSH where the client terminal defines the colors.

## Current behavior

- `fg()` styled text supports ANSI named colors via `red()`, `blue()`, `cyan()` etc.
- `borderColor` and `backgroundColor` accept `ColorInput` but always render as true color, even when a named color like `"red"` is passed — `parseColor` converts it to a fixed RGB value (`rgba(1,0,0)`)

## Expected behavior

When a named ANSI color (e.g. `"red"`, `"blue"`, `"brightCyan"`) is used as `borderColor` or `backgroundColor`, it should emit the corresponding ANSI 16 escape code instead of a true color sequence, so the terminal can map it to its palette.

## Use case

Building a TUI app that respects the user's terminal theme out of the box, without requiring manual color configuration. This is standard behavior in tools like lazygit and other popular terminal apps.

seffhunnn:
## Summary

Preserves ANSI 16 named colors for `borderColor` and `backgroundColor` instead of always converting them into truecolor RGB escape sequences.

This allows terminal themes and SSH client palettes to map colors correctly for named colors such as `"red"`, `"blue"`, and `"brightCyan"`.

## Changes

* tagged ANSI named colors with palette metadata in `RGBA`
* added ANSI 16 foreground/background helpers
* updated renderer to prefer ANSI palette escape sequences when available
* added end-to-end Box rendering tests

## Testing

* ✅ `bun test src/lib/RGBA.test.ts` → 131 passing
* ⚠️ `bun test src/renderables/Box.test.ts` requires native binary package (`@opentui/core-win32-x64`) unavailable locally, expected to run in CI

Closes #852

### DIFF
--- a/packages/core/src/ansi.ts
+++ b/packages/core/src/ansi.ts
@@ -10,6 +10,8 @@ export const ANSI = {
   moveCursorAndClear: (row: number, col: number) => `\x1b[${row};${col}H\x1b[J`,
 
   setRgbBackground: (r: number, g: number, b: number) => `\x1b[48;2;${r};${g};${b}m`,
+  setAnsi16Background: (index: number) => index < 8 ? `\x1b[${40 + index}m` : `\x1b[${100 + (index - 8)}m`,
+  setAnsi16Foreground: (index: number) => index < 8 ? `\x1b[${30 + index}m` : `\x1b[${90 + (index - 8)}m`,
   resetBackground: "\x1b[49m",
 
   // Bracketed paste mode
--- a/packages/core/src/lib/RGBA.ts
+++ b/packages/core/src/lib/RGBA.ts
@@ -1,5 +1,9 @@
 export class RGBA {
   buffer: Float32Array
+  /** ANSI 16-color palette index (0-15) when this color was created from a
+   *  standard named color (e.g. "red", "brightcyan"). Undefined for all other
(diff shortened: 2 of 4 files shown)


seffhunnn:
Added a TypeScript-only fix that preserves ANSI 16 palette metadata through the Box rendering pipeline, with end-to-end tests for named color behavior.

Local TS tests are passing; Box integration verification is expected to complete in CI because native binaries are unavailable locally.

simonklee:
Thank you for the contribution. We are closing this pull request because the
current color system stores palette information in the native RGBA format. This
change does not send its ANSI color index to the native renderer, so box
borders and backgrounds still use RGB colors. Issue #852 remains open for a
solution that uses the current color system.

PR #934, opened 2026-04-08
workaround windows terminal bug when sending XTVERSION before alt screen

### LINKED ISSUE #933: Windows Terminal stalls when XTVERSION is sent before alt-screen
## Summary
On a real Windows Terminal-attached session, OpenTUI startup can poison the session so ordinary JS timers and resize handling stall for seconds. The trigger appears to be sending `XTVERSION` (`ESC[>0q`) before switching to alt-screen.

Reordering startup so `queryTerminalSend()` runs after `setupTerminalWithoutDetection(useAlternateScreen)` made both the minimal harness and the real `opencode` TUI resize smoothly in local testing.

## Symptoms
- resize (`SIGWINCH`) appears delayed by 5-30s
- plain `setTimeout(..., 100)` can be delayed by the same amount
- `Ctrl+C` can also feel delayed while in the stalled state
- once resize/render actually runs, the work is fast

## Environment
- Windows Terminal / real attached host session
- Windows 11 / ConPTY path
- reproduces with both Bun and Node
- does **not** reproduce in a synthetic PTY driven by `bun-pty`

## Smallest standalone repro
This can be reproduced outside OpenTUI with a tiny script that only sends `XTVERSION` before alt-screen and logs watchdog gaps:

```js
process.stdout.write("\x1b[>0q\x1b[?1049h\x1b[?25l")
setInterval(() => {
  const now = performance.now()
  const gap = now - last
  last = now
  if (gap > 200) console.log({ gap })
}, 50)
setTimeout(() => process.exit(0), 4000)
```

Observed result in the real terminal:
- `\x1b[>0q` **before** alt-screen: watchdog lag entries appear
- `\x1b[>0q` **after** alt-screen: smooth
- no `XTVERSION`: smooth

I also verified the same ordering effect under both:
- `node`
- `bun`

So this does **not** look Bun-specific.

## OpenTUI-specific evidence
Current startup order in `packages/core/src/zig/renderer.zig`:

```zig
self.terminal.queryTerminalSend(writer)
writer.flush()
self.setupTerminalWithoutDetection(useAlternateScreen)
```

That matches the bad ordering.

Relevant local experiments against the real bugged `textarea` harness:
- `OPENTUI_MINIMAL_SETUP=1`: smooth
- `OPENTUI_SKIP_QUERY_SEND=1`: smooth
- `OPENTUI_SKIP_PENDING_QUERIES=1`: still laggy
- `OPENTUI_SKIP_ENABLE_FEATURES=1`: still laggy
- `OPENTUI_SKIP_XTVERSION=1`: smooth
- `OPENTUI_SKIP_QUERY_WRAPPER=1`: still laggy

This points specifically at the **initial query send**, and more specifically at `XTVERSION`.

## Likely fix
Move `queryTerminalSend(writer)` to run **after** `setupTerminalWithoutDetection(useAlternateScreen)`.

That reorder fixed the issue in my local Windows Terminal testing for:
- a minimal OpenTUI `TextareaRenderable` harness
- the real `opentui` example selector
- the real `opencode` TUI (after ensuring the patched DLL was actually loaded)

## Notes
The problem did **not** reproduce when I replayed the same query bundles in a synthetic PTY or when the same query bundle was sent after alt-screen, which is why the ordering looks important rather than the query content alone.

Hona:
## Summary
- move startup terminal queries to run after alternate-screen setup
- avoid the Windows Terminal stall caused by sending `XTVERSION` before entering alt-screen
- keep the existing startup detection logic intact and only change ordering

## Testing
- bun run build:native
- verified locally against the OpenTUI textarea harness, example selector, and OpenCode after loading the patched native build

Closes #933

### DIFF
--- a/packages/core/src/zig/renderer.zig
+++ b/packages/core/src/zig/renderer.zig
@@ -309,6 +309,8 @@ pub const CliRenderer = struct {
         self.useAlternateScreen = useAlternateScreen;
         self.terminalSetup = true;
 
+        self.setupTerminalWithoutDetection(useAlternateScreen);
+
         var stdoutWriter = std.fs.File.stdout().writer(&self.stdoutBuffer);
         const writer = &stdoutWriter.interface;
 
@@ -316,8 +318,6 @@ pub const CliRenderer = struct {
             logger.warn("Failed to query terminal capabilities", .{});
         };
         writer.flush() catch {};
-
-        self.setupTerminalWithoutDetection(useAlternateScreen);
     }
 
     fn setupTerminalWithoutDetection(self: *CliRenderer, useAlternateScreen: bool) void {


Hona:
upstream WT has kindly fixed this quickly... won't need this :)

https://github.com/microsoft/terminal/pull/20095

PR #936, opened 2026-04-08
feat(core): add input type password with character masking

LeTamanoir:
## Summary
  - Add `type` prop (`"text" | "password"`) and `passwordChar` prop to `<input>`
  - When `type="password"`, all rendered characters are replaced with a mask character (default `●`) while the edit buffer retains the real value
  - Masking is applied at the Zig rendering layer via a new `mask_codepoint` field on `EditorView`, so cursor, selection, and navigation all work unchanged
  - Placeholder text is not masked


selfmadexyz (approved):
+1

simonklee:
1. I would have expected something like this to be implemented completely in the Input renderable and not native layers.
2. Because of 1) you've introduced misc unicode bugs.

Closing since it's not the right approach.

LeTamanoir:
okay will try in the input renderable then 👍

simonklee:
> okay will try in the input renderable then 👍

Awesome!

Issue #821, opened 2026-03-16
Roadmap


kommander:
Build the features OpenCode needs, stabilize the core, then refactor and optimize.

## Now

_v0.1, v0.2, v0.3 — Exploring the problem space._

- [DONE] CLI render mode (split mode), with interactive TUI at the bottom and immutable scrollback log stream at the top.https://github.com/anomalyco/opentui/commit/bf8f195c84408011ab96ed69bd17d825431ae9f1
- [DONE] Rework color transport for RGBA+A and indexed metadata. Cache nearest-index translations so both TS and native paths skip repeated conversion. https://github.com/anomalyco/opentui/commit/95d36f3553ea9eb2b3a5f4d9f13ecd261e51602d
- [DONE] Split the 3d/threejs core part into a separate package https://github.com/anomalyco/opentui/commit/a10e71c9bfe2fd228fbe24d813184c1fc26c9850
- Find what features OpenCode and terminal.shop need
- Replace `setStyledText` with a native interface for incremental text updates (depends on MarkTree)
- Cross-platform clipboard with OSC52 and platform-specific fallbacks
- Make OpenTUI run on other runtimes like Node and Deno (debunification)
- Native streaming path for SSH connections
- Add virtual text for textarea autocompletion. TypeScript-only first, native via MarkTree later
- Hunk navigation for code review in OpenCode. Inline editing likely depends on MarkTree
- Timing-based scroll handling with inertia and speed, telling trackpad from mouse. Partial version exists
- Load external modules at runtime (in progress)
- OpenCode tests use OpenTUI test helpers. Fill gaps in OpenTUI's testing tools
- Kitty Graphics & Sixel support

## Next

_v0.x — Refactor the hell out of it._

- Neovim-style sticky marks (MarkTree/extmarks) for highlights, virtual text, and conceal. Marks move with edits so treesitter skips full re-highlighting
- Move markdown parsing and rendering to native code, compatible with MarkTree
- TextBufferView caches all virtual lines, even invisible ones. Constrain to the viewport
- Close the gap between OpenTUI's grapheme/width model, Unicode, and how terminals render complex text
- Kitty keyboard protocol follow-ups for repeat event shape, modified Tab encoding, KP_BEGIN naming, and other spec gaps

## Later

_v1.0 — Everything moves native._

- The renderable tree, layout engine (Yoga), and rendering move to native code. Native callbacks for `calculateLayout` let you update renderables directly without a separate update pass

## Ongoing

_Work that spans every phase._

[...]

PR #939, opened 2026-04-10
Node.js support

justjake:
Partially addresses #2 

This PR brings Node.js compatibility. Bun-only imports and APIs have been replaced by compatibility interface modules, which pick a runtime-appropriate implementation. Most codepaths when running in `bun` are the same as before the PR.

### Node.js differenes

- `bun:ffi` -> `@opentui/core/compat/ffi`: I re-implemented relevant `bun:ffi` APIs on top of [koffi](https://koffi.dev). There's more copying than in bun, and I'm unsure about correctness, but I think it's a worthwhile start. The upcoming version of koffi offer capabilities much closer to bun, but I'm not sure when the author will release it.
- `bun:test` -> `@opentui/core/compat/test`: I know the AGENTS.md says no vitest, but vitest is the most popular Node.js test runner, and its api is very similar to bun:test. We use vite alias to replace bun:test with the compat module based on vitest. Source code still says `"bun:test"` because "bun:test" only works when your test file directly imports it.
    -  to avoid snapshots from different test runners thrashing, we create independent snapshots for nodejs. Perhaps this is avoidable somehow? I didn't investigate.
- `globalThis.Worker ` -> `@opentui/core/compat/Worker`. Node.js doesn't have a global worker, and its worker_threads worker isn't web-compatible without this shim. Warning: there's a bit of weird node module loader shenanigans here to work around node not understanding `import "./foo.js"` means `import "./foo.ts"`, but it's pretty benign for opentui's use.
- `bun-ffi-structs`: oh boy. I ended up having codex vendor this because module-loader tricks won't work once we're bundled, and it wasn't immediately obvious how to do as a build step. Bun runtime still uses upstream, nodejs gets this hack.
- `Bun.spawnSync` -> `@opentui/core/compat/testHelpers`. This is only used in test, so I put in own file to avoid bundling it.
- `Bun.*` -> `opentui/core/compat/runtime`: Re-implemented Bun.sleep, Bun.writeFile, and installed the typical npm packages for Bun.stringWidth, Bun.stripANSI
- `import ... with { ... }` -> `new URL(..., import.meta.url)`. These were pretty mechanically easy to switch to the more universal option.

### Testing

Test scripts:

- `test:bun`: the old `bun run test` | `bun run test:js`. Unit tests w/ bun:test.
- `test:nodejs`: `npx vitest run`. Maybe I could use `bunx` but I want to run under node!
- `test:dist`: New type of integration test that tests that the `npm pack`'d artifact actually works in a few very simple example projects. Node is not as forgiving as Bun, so this caught a few goofs that don't show up in vitest or bun test. For example, for some reason Bun@1.2 would emit invalid bundles with duplicate exports, detected by this test. Another case was needing to import react-reconciler/something.js w/ the extension. ¯\_(ツ)_/¯ node.

### Thanks

This is a cool project, and thank you in advance for your attention to this matter. 

Let me know if you'd prefer I split up the changes. I personally prefer to review 1 big pr rather than 10 small prs without context, but I know its not the common opinion. Lots of those lines of code are duplicate snapshots!


packages/core/src/lib/selection.ts
```diff
@@ -1,4 +1,5 @@
-import { Renderable, type ViewportBounds } from "../index.js"
```
justjake, line 1: This caused a mysterious circular dependency in node

simonklee: Yeh, ran into this as well when doing deno build of opentui

simonklee:
Thanks for taking a stab at this. A few questions, just to start;

- What is koffi and how does it compare to https://github.com/nodejs/node/pull/62072 once that lands?
- What is mise and why is it needed
- Why Node 22 and not target latest version
- We [plan](https://github.com/anomalyco/opentui/issues/821) on moving threejs/3d to its own package so any compat/complexity added because of that should probably be avoided.

justjake:
> Thanks for taking a stab at this. A few questions, just to start;
> 
> * What is koffi and how does it compare to [lib,src,test,doc: add node:ffi module nodejs/node#62072](https://github.com/nodejs/node/pull/62072) once that lands?

koffi is a NAPI native module that implements ffi, similar to bun:ffi or this node:ffi pr. Like node:ffi, it passes ArrayBuffer backed values as pointers to native code. 

They're all quite similar, but bun:ffi is notable for exposing `ptr(typedArray) -> number`, which neither the current version of koffi or node:ffi expose. However, the next version of koffi has an equivalent `koffi.address(typedArray) -> bigint`, which is missing from node:ffi and would make compat/nodejs/ffi.ts more efficient and truly zero-copy.

We could make both current version of koffi & any ffi layer based on node:ffi zero-copy by extending the `Pointer` type to be `type Pointer = number & { __pointer__: null } | TypedArray | ArrayBufferLike` and making ptr an identity function under nodejs (`export const ptr = (value) => value`), that may not be too much work, I didn't investigate to what degree opentui relies on `ptr` returning number values.

Fun fact, koffi supports and provides premade builds for many architectures / platforms. I only care about the usual linux_x64 / darwin_arm64 / win32_x64 but I think it's cute:

```
        case 'android_arm64': { native = require('@koromix/koffi-android-arm64'); } break;
        case 'android_x64': { native = require('@koromix/koffi-android-x64'); } break;
        case 'darwin_arm64': { native = require('@koromix/koffi-darwin-arm64'); } break;
        case 'darwin_x64': { native = require('@koromix/koffi-darwin-x64'); } break;
        case 'freebsd_arm64': { native = require('@koromix/koffi-freebsd-arm64'); } break;
        case 'freebsd_ia32': { native = require('@koromix/koffi-freebsd-ia32'); } break;
        case 'freebsd_x64': { native = require('@koromix/koffi-freebsd-x64'); } break;
        case 'linux_armhf': { native = require('@koromix/koffi-linux-arm'); } break;
        case 'linux_arm64': { native = require('@koromix/koffi-linux-arm64'); } break;
        case 'linux_ia32': { native = require('@koromix/koffi-linux-ia32'); } break;
        case 'linux_loong64': { native = require('@koromix/koffi-linux-loong64'); } break;
        case 'linux_riscv64d': { native = require('@koromix/koffi-linux-riscv64'); } break;
        case 'linux_x64': { native = require('@koromix/koffi-linux-x64'); } break;
        case 'openbsd_ia32': { native = require('@koromix/koffi-openbsd-ia32'); } break;
        case 'openbsd_x64': { native = require('@koromix/koffi-openbsd-x64'); } break;
        case 'win32_arm64': { native = require('@koromix/koffi-win32-arm64'); } break;
        case 'win32_ia32': { native = require('@koromix/koffi-win32-ia32'); } break;
        case 'win32_x64': { native = require('@koromix/koffi-win32-x64'); } break;
```

> * What is mise and why is it needed

mise is a version manager like `asdf` or `nvm`. it's not needed in ci, just a convenience for other engineers who may want to set up zig + node + bun with a single command (`mise install`) in the repo. I can remove `mise.toml` fine with no affect on the rest of the pr.

> * Why Node 22 and not target latest version

In general think it's good to target the oldest LTS version when building new things unless it's a significant disadvantage to implementation quality. What's the benefit from only targeting the latest version?

Probably more importantly, I want to use opentui in a 25 million line codebase that's currently running on Node.js 22 (https://www.notion.com).

> * We [plan](https://github.com/anomalyco/opentui/issues/821) on moving threejs/3d to its own package so any compat/complexity added because of that should probably be avoided.

No incidental complexity from 3d stuff. Maybe I changed some imports in there to compat/ffi, same as other files. The 3D examples don't run under Node due to deeper dependency on Bun stuff it seems.

justjake:
Implemented `bun:ffi` `ptr` and `toArrayBuffer` as a Node.js native module [unsafe-pointer](https://github.com/justjake/node-unsafe-pointer), [npm](https://www.npmjs.com/package/unsafe-pointer). Includes prebuilds for linux-${arm64 | x64}-${glibc | musl}, windows-${arm64 | x64}, darwin-${arm64 | x64}. All builds done in Github Actions using `zig cc` cross compilation, tested on the target platform, then published to NPM with attestation w/ OIDC connect auth. It's about [500 lines of simple c](https://github.com/justjake/node-unsafe-pointer/blob/main/src/binding.c).

With this, compat/nodejs/ffi.ts is now zero-copy and less complex.

justjake:
We can port from `koffi` and `node-unsafe-pointer` to `node:ffi` once it's stable and upstream merges this PR:  https://github.com/nodejs/node/pull/62762

Is there anything I can do to make this PR easier to review or merge?

simonklee:
> Is there anything I can do to make this PR easier to review or merge?

We have a call with the Node.js/Platformatic people that are working on the node ffi things next week. I'll probably have a look at this around that time. For now its mostly a priorities kind of thing.

MBearo:
> > Is there anything I can do to make this PR easier to review or merge?
> 
> We have a call with the Node.js/Platformatic people that are working on the node ffi things next week. I'll probably have a look at this around that time. For now its mostly a priorities kind of thing.

What is the current progress on this?

simonklee:
> > > Is there anything I can do to make this PR easier to review or merge?
> > 
> > 
> > We have a call with the Node.js/Platformatic people that are working on the node ffi things next week. I'll probably have a look at this around that time. For now its mostly a priorities kind of thing.
> 
> What is the current progress on this?

https://github.com/anomalyco/opentui/pull/1055

justjake:
@MBearo I published this branch to NPM :)

https://www.npmjs.com/package/@jitl/opentui-react
https://www.npmjs.com/package/@jitl/opentui-core

Demos:

https://github.com/justjake/opentui-node-demo

MBearo:
> @MBearo I published this branch to NPM :)
> 
> https://www.npmjs.com/package/@jitl/opentui-react https://www.npmjs.com/package/@jitl/opentui-core
> 
> Demos:
> 
> https://github.com/justjake/opentui-node-demo

Thank you for your efforts!

justjake:
Closing as upstream has merged Node 26 support.

@kommander or @simonklee, are you interested in a more reasonably scoped PR to add FFI support for older node? We've moved several Notion internal tools from Ink to my node22 compat fork, works great!

remorses:
I think making opentui work in older Node.js versions would be great. It's the only way to ship TUIs on npm that can work for most users without bundling Node.js 26

PR #941, opened 2026-04-11
core: compact keypress debug output and add IME repro

simonklee:
Rework the keypress debug demo around parsed events instead of large per-event cards. Keep the main feed dense, move raw input into session detail and export output, and show the latest event with useful fields such as source, code, and Kitty baseCode.

Add test that reproduces Ctrl+C when Kitty reports an alternate base layout for a non-Latin IME. For https://github.com/anomalyco/opencode/issues/21163

### DIFF
--- a/packages/core/src/console.test.ts
+++ b/packages/core/src/console.test.ts
@@ -315,6 +315,23 @@ describe("TerminalConsole", () => {
       expect(onCopy).toHaveBeenCalledWith("Hello")
     })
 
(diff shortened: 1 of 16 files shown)


kommander (dismissed):

simonklee:
I'll add a followup in opencode that adds a few of these:

```
if (keybind.match("ctrl+c", evt)) {
```

instead of just matching on evt.name which may be a different codepoint

PR #944, opened 2026-04-12
performance

kommander:
- Added _screenX/_screenY layout-time absolute coordinate caching and used it in hot render paths (Renderable, Box, TextBufferRenderable, EditBufferRenderable).
- Skipped drawBox entirely for boxes with no visible border and no visible fill.
- Optimized native transparent fill paths in buffer.zig:
  - fillRect uses cheaper paths when no grapheme/link tracking is involved
  - fully transparent fills preserve existing visible text while normalizing empty cells
- Optimized native transparent border drawing in drawBox for width-1 opaque border glyphs over transparent backgrounds.
- Optimized native text-buffer rendering for opaque width-1 text over transparent backgrounds, bypassing generic alpha blending when the result is equivalent.
- Avoided visible-child filtering work for renderables that do not actually filter children by adding _hasVisibleChildFilter(); only viewport-culling ScrollBox content now pays for _getVisibleChildren() and membership checks.
- Switched primary-axis child sorting and ScrollBox viewport culling to absolute screenX/screenY, including updating objects-in-viewport to operate on screen-space coordinates instead of recursive x/y walks.
- Added/expanded correctness tests 
- Added new headless benchmarks
  - box-draw-benchmark.ts
  - render-traversal-benchmark.ts
  - text-buffer-render-benchmark.ts


packages/core/src/zig/buffer.zig
```diff
@@ -852,6 +852,38 @@ pub const OptimizedBuffer = struct {
...
 
+    inline fn trySetTransparentTextCellFast(
+        self: *OptimizedBuffer,
+        index: u32,
+        char: u32,
+        fg: RGBA,
+        attributes: u32,
+    ) bool {
+        // drawTextBuffer spends a lot of time in generic alpha blending when the
+        // common case is really "opaque glyph over transparent bg". In that case
```
simonklee, line 875: https://github.com/anomalyco/opentui/pull/870/changes

Maybe bring this in as well.

simonklee:
https://github.com/anomalyco/opentui/pull/865/changes also did some adjacent changes.

PR #948, opened 2026-04-14
fix: Possible fix for ignoring "sticky scroll" when manually scrolled up

jabr:
See https://github.com/anomalyco/opencode/pull/19540

I wasn't able to recreate this issue with tests, but I _suspect_ this is the issue.

### DIFF
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -752,7 +752,7 @@ export class ScrollBoxRenderable extends BoxRenderable {
 
         if (this._stickyStart && !this._hasManualScroll) {
           this.applyStickyStart(this._stickyStart)
-        } else {
+        } else if (!this._hasManualScroll) {
           if (this._stickyScrollTop) {
             this.scrollTop = 0
           } else if (this._stickyScrollBottom && newMaxScrollTop > 0) {


jabr:
https://github.com/anomalyco/opencode/pull/19540#issuecomment-4272792421

I've not been able to test this patch (I'm unable to build this library successfully so far), but it appears to actually have nothing to do with the "sticky scroll" setting. I completely disabled that setting to the opentui scrollbox in a custom opencode build, and it's doing the same intermittent scroll jumps on output for me still.

From what I can tell, it seems to be specific kinds of output (e.g. code blocks, tables) but not all (e.g. thinking or normal text output) that cause the scroll position to move.

And it's not entirely consistent on how it effects the scroll. Sometimes a small movement, sometimes all the way to the bottom. That might be related to where I am in the scroll window. Closer to the bottom seems more likely to jump be all the rest of the way, but further up towards the top tends to be smaller jumps. But hard to say for certain on that bit.

simonklee:
I don’t think this is the right fix. The change only affects the sticky-scroll path in ScrollBox.recalculateBarProps(), so this looks speculative rather than validated.

jabr:
Yeah, it doesn't seem to be sticky-scroll setting related. As best as I can tell, it's resizing of certain child elements of the scrollbox that cause the scroll position to shift.

PR #951, opened 2026-04-15
fix: emit OSC 11/111 to sync terminal background color with renderer

mynameistito:
When `setBackgroundColor()` is called, opentui updates the render buffer but the terminal's own background color stays at whatever the profile default is. On most terminals this is invisible. On Windows Terminal (maximized or snapped), there's a pixel gutter on the right and bottom edges where the character grid doesn't divide evenly into the window. Those pixels get painted with the terminal's default, not the app's theme. It's a visible seam.

This PR fixes it by emitting OSC 11/111 in the right places inside the renderer, so apps don't have to do it themselves.

## Changes

**`zig/ansi.zig`**: `setTerminalBgColorOutput()`: emits OSC 11 (`\e]11;rgb:RR/GG/BB\e\`)

**`zig/renderer.zig`**: `setBackgroundColor()` now emits OSC 11 after updating the buffer. Skips when alpha is 0 (transparent background, let the terminal keep its own default) and when `testing` is true.

**`zig/terminal.zig`**: `resetState()` now emits OSC 111 after resetting the title. This fires on both `destroy()` and `suspend()` through `performShutdownSequence()`.

**`renderer.ts`**: Exposes `resetTerminalBgColor()` as a public method for consumers that need explicit control before SIGTSTP or similar.

## Compatibility

OSC 11/111 is supported by Windows Terminal, iTerm2, kitty, Alacritty, foot, and others. Terminals that don't support it ignore it silently. No behavior change on unsupported terminals.

## Testing

Mini repro with unit tests: https://github.com/mynameistito/opentui-osc11-repro

Verified manually on Windows 11 (OS Build 26200) with Windows Terminal `v1.24.10621.0` and Terminal Preview `v1.25.622.0`. Gutter matches theme background, resets on exit.

https://github.com/user-attachments/assets/19bc3fb1-e45c-407a-b5af-5465bb9293da

## Fixes
Fixes: https://github.com/anomalyco/opentui/issues/950
Fixes: https://github.com/anomalyco/opencode/issues/18055
Fixes: https://github.com/anomalyco/opencode/issues/19383
Fixes: https://github.com/microsoft/terminal/issues/19860
Fixes: https://github.com/microsoft/terminal/issues/19989

PR #952, opened 2026-04-16
fix(build): pass --cache-dir on Windows to avoid cross-drive zig panic

mynameistito:
When opentui's build script runs `zig build` on Windows and the project directory and zig's global cache live on different drives (project on `F:`, zig on `C:`), zig 0.15.x panics. `std.fs.path.relative` returns an absolute path across drives, which trips an assertion in `Build/Step/Run.zig:662` inside the `uucode` dependency's build-tables step.

The fix: pass `--cache-dir` pointing to `%USERPROFILE%\.zig-cache\opentui-core` when building on Windows. This keeps all path operations on one drive and sidesteps the panic. No manual steps required from users.

Split from #951 per https://github.com/anomalyco/opentui/pull/951#issuecomment-4258584684

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,10 @@ NOTE: You must have [Zig](https://ziglang.org/learn/getting-started/) installed 
 bun install @opentui/core
 ```
 
+### Windows — cross-drive build
+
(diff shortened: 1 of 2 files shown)


simonklee:
We already have a  fix for this in ci: https://github.com/anomalyco/opentui/commit/82d7fa4ed5b47d093bb2a809876c6de79a00820f i'll consider if we should consolidate.

mynameistito:
sweet as ty @simonklee :)

simonklee:
I dont think this is no longer required.

PR #970, opened 2026-04-24
fix(core): keep renderable ids valid

aarcamp:
Ensure renderables always retain a non-empty string id. Invalid constructor ids now fall back to generated ids, and invalid id assignments are ignored so parent child maps remain consistent.

This prevents children from becoming orphaned when removal relies on child.id after an invalid id value is assigned. In practice, this can happen when a reused element in a dialog or mode-switching subtree drops its id before being unmounted, causing parent.remove(child.id) to receive an invalid id and leave the child attached.

Basically, this change protects the basic rule that every renderable must always have a usable id, because parent maps and unmount/removal logic depend on it.

In case it helps, here's the simple change in my app that surfaced this bug:

https://github.com/honeymux/honeymux/commit/5825514966ffab01a57c7e331674fcdbf8cae645#diff-6df7e2acca06e3499038f86a1fbd0712da07bb7b8f4baca4f5beadc973f440d5

The addition of `id` here broke unmount of the component until I fixed this problem on the OpenTUI side.

### DIFF
--- a/packages/core/src/Renderable.ts
+++ b/packages/core/src/Renderable.ts
@@ -36,6 +36,10 @@ import {
 
 const BrandedRenderable: unique symbol = Symbol.for("@opentui/core/Renderable")
 
+function isValidRenderableId(value: unknown): value is string {
+  return typeof value === "string" && value.length > 0
+}
+
 export enum LayoutEvents {
   LAYOUT_CHANGED = "layout-changed",
   ADDED = "added",
@@ -146,7 +150,7 @@ export abstract class BaseRenderable extends EventEmitter {
   constructor(options: BaseRenderableOptions) {
     super()
     this.num = BaseRenderable.renderableNumber++
-    this._id = options.id ?? `renderable-${this.num}`
+    this._id = isValidRenderableId(options.id) ? options.id : `renderable-${this.num}`
   }
 
   public abstract add(obj: BaseRenderable | unknown, index?: number): number
@@ -163,6 +167,10 @@ export abstract class BaseRenderable extends EventEmitter {
   }
 
   public set id(value: string) {
+    if (!isValidRenderableId(value)) {
+      return
+    }
+
     this._id = value
   }
 
@@ -308,6 +316,10 @@ export abstract class Renderable extends BaseRenderable {
   }
 
   public override set id(value: string) {
+    if (!isValidRenderableId(value)) {
+      return
+    }
+
     if (this.parent) {
       this.parent.renderableMapById.delete(this.id)
       this.parent.renderableMapById.set(value, this)
--- a/packages/core/src/tests/renderable.test.ts
+++ b/packages/core/src/tests/renderable.test.ts
@@ -105,6 +105,22 @@ describe("BaseRenderable", () => {
     expect(renderable.id).toBe("custom-id")
   })
 
+  test("falls back to generated id for invalid constructor id", () => {
+    for (const id of ["", null, 42]) {
+      const renderable = new TestBaseRenderable({ id } as any)
+      expect(renderable.id).toMatch(/^renderable-\d+$/)
+    }
+  })
+
+  test("ignores invalid id assignments", () => {
+    const renderable = new TestBaseRenderable({ id: "stable-id" })
+
+    for (const id of ["", null, undefined, 42]) {
+      renderable.id = id as any
+      expect(renderable.id).toBe("stable-id")
+    }
+  })
+
   test("has unique numbers", () => {
     const r1 = new TestBaseRenderable({})
     const r2 = new TestBaseRenderable({})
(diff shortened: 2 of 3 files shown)


aarcamp:
Here is a PoC script (copy to and run from `packages/react/scripts/opentui-pr-970-poc.tsx`):

```
aaron@bow:~/src/opentui$ cat packages/react/scripts/opentui-pr-970-poc.tsx
#!/usr/bin/env bun

// Repro for anomalyco/opentui#970.
// Run from the OpenTUI repo root:
// bun run packages/react/scripts/opentui-pr-970-poc.tsx

import React, { act, useState } from "react"
import { testRender } from "../src/test-utils.js"

const PROBE_TEXT = "PR970_STALE_CHILD"
const TARGET_ID = "pr-970-target"

let child: any
let setStep: any

function App() {
  const [step, setStepState] = useState(0)
  setStep = setStepState

  return (
    <box>
      {step === 2 ? null : (
        <box
          id={step === 0 ? TARGET_ID : undefined}
          ref={(node: any) => {
            if (node) child = node
          }}
        >
          <text>{PROBE_TEXT}</text>
        </box>
      )}
    </box>
  )
}

const setup = await testRender(<App />, { height: 4, width: 40 })

function frameContainsProbe() {
  return setup.captureCharFrame().includes(PROBE_TEXT)
}

try {
  await setup.renderOnce()

  const parent = child.parent
  console.log("1. Initial render")
  console.log(`   child.id = ${String(child.id)}`)
  console.log(`   parent children = ${parent.getChildrenCount()}`)
  console.log(`   probe visible = ${frameContainsProbe()}`)

  await act(async () => setStep(1))
  await setup.renderOnce()

  console.log("\n2. Same child re-rendered with id={undefined}")
  console.log(`   child.id = ${String(child.id)}`)
  console.log(`   parent children = ${parent.getChildrenCount()}`)
  console.log(`   probe visible = ${frameContainsProbe()}`)

  await act(async () => setStep(2))
  await setup.renderOnce()

  const stillMounted = child.parent === parent
  const childCount = parent.getChildrenCount()
  const stillVisible = frameContainsProbe()

  console.log("\n3. Child unmounted")
  console.log(`   child.parent is still parent = ${stillMounted}`)
  console.log(`   parent children = ${childCount}`)
  console.log(`   probe visible = ${stillVisible}`)

  if (stillMounted || childCount !== 0 || stillVisible) {
    console.log("\nBUG: child survived unmount after its id became invalid.")
    process.exitCode = 1
  } else {
    console.log("\nOK: child was removed cleanly.")
  }
} finally {
  await act(async () => setup.renderer.destroy())
}
aaron@bow:~/src/opentui$
```

Output before fix:

```
aaron@bow:~/src/opentui$ git switch main
Switched to branch 'main'
Your branch is up to date with 'upstream/main'.
aaron@bow:~/src/opentui$ rm -rf node_modules/ ~/.bun/install/cache
aaron@bow:~/src/opentui$ bun install
bun install v1.3.13 (bf2e2cec)

+ oxfmt@0.41.0
+ oxlint@1.56.0

1018 packages installed [1.53s]
aaron@bow:~/src/opentui$ bun run packages/react/scripts/opentui-pr-970-poc.tsx
1. Initial render
   child.id = pr-970-target
   parent children = 1
   probe visible = true

2. Same child re-rendered with id={undefined}
   child.id = undefined
   parent children = 1
   probe visible = true

3. Child unmounted
   child.parent is still parent = true
   parent children = 1
   probe visible = true

BUG: child survived unmount after its id became invalid.     <-- HERE
aaron@bow:~/src/opentui$
```

Output after fix:

```
aaron@bow:~/src/opentui$ git switch fix-renderable-id-invariant
M	bun.lock
Switched to branch 'fix-renderable-id-invariant'
Your branch and 'origin/fix-renderable-id-invariant' have diverged,
and have 32 and 1 different commits each, respectively.
  (use "git pull" if you want to integrate the remote branch with yours)
aaron@bow:~/src/opentui$ bun run packages/react/scripts/opentui-pr-970-poc.tsx
1. Initial render
   child.id = pr-970-target
   parent children = 1
   probe visible = true

2. Same child re-rendered with id={undefined}
   child.id = pr-970-target
   parent children = 1
   probe visible = true

3. Child unmounted
   child.parent is still parent = false
   parent children = 0
   probe visible = false

OK: child was removed cleanly.
aaron@bow:~/src/opentui$
```

Let me know if there's anything else I can do to further demonstrate the bug.

simonklee:
@kommander: What do you think about

- Give every renderable an immutable internal id/key used for parent maps and removal.
- Treat the user id prop as an optional alias.

I don't like that these id's are used for internal and external things.

kommander:
@simonklee having two kinds of ids would be confusing imo. If there is an internal id that is never exposed anywhere, okay. But I don't like having both exposed.

aarcamp:
@simonklee What's the way forward here?

simonklee:
Thank you for the contribution. Pull request #1224 fixed this problem by
removing children by object identity instead of ID. The proposed changes and
tests depend on code that no longer exists so I'm closing this.

PR #1015, opened 2026-05-04
fix: Add GraphemePoolError and clamp for oversized cluster

jettptacek:
## Summary

I can across this issue with my ssh chat server project which is using a decently modified fork of opentui. 
`GraphemePool.alloc` passes `bytes` straight through to `ClassPool.allocInternal`, which panics if `bytes.len > slot_capacity`. Also`CLASS_SIZES` tops out at 128 B, so any cluster ≥ 129 B was a hard crash.

I've instead clamped the incoming bytes to the largest slot on a UTF-8 code-pointboundary, and converted the remaining `@panic` in `ClassPool.allocInternal`into a recoverable `GraphemePoolError.GraphemeTooLarge` so callers can handle misuse instead of dying.

## Testing

I add a new grapheme test "GraphemePool - oversized zalgo cluster truncates instead of panicking"

#### P.S.
As always I understand if you want to go a different direction than a GraphemePoolError and the clamping


kommander:
Cool, this has been on my list for a very long time. I would drop the `std.log.warn` as we should not write to stdout/stderr directly to avoid rendering disruption when the renderer writes directly to stdout itself.

simonklee:
> Cool, this has been on my list for a very long time. I would drop the `std.log.warn` as we should not write to stdout/stderr directly to avoid rendering disruption when the renderer writes directly to stdout itself.

I think this PR is sort of implemented in the wrong layer and has some other issues. I was messing around with it locally the other day and will commit something if i have time tomorrow.

jettptacek:
I just saw the panic and put all the work in the GraphemePool alloc. Sorry I should have done proper due diligence. I agree the grapheme pool is this wrong layer for "clamping"

PR #1017, opened 2026-05-04
fix: cursor position restoration for undo/redo in EditBuffer

### LINKED ISSUE #980: EditorView.undo() / redo() does not sync cursor position with restored buffer state
## Description

`EditorView.undo()` and `EditorView.redo()` in `@opentui/core` restore the
underlying `editBuffer` text to a previous snapshot but never move the cursor
to a position consistent with the restored content. The cursor stays wherever
it was at the moment the user pressed undo/redo, even though that offset may
no longer be meaningful (e.g. it can land past the end of the buffer or in
the middle of where text used to be).

This makes undo/redo feel broken from a user's perspective in any host that
uses `EditorView` for text input — most notably the prompt input in
`opencode`.

## Steps to reproduce

1. Use any host application embedding `@opentui/core`'s `EditorView`
   (verified in `opencode` 1.14.25 with `@opentui/core` 0.1.103).
2. Type a few words into the input. Cursor is at the end of the line.
3. Move the cursor back a few characters (e.g. `Left` arrow several times).
4. Type more text. Cursor is now somewhere in the middle.
5. Press the undo keybind.

**Expected**: cursor moves to the position it had at the snapshot being
restored (or at minimum, to a defined position consistent with the restored
text — typically where the next edit would happen).

**Actual**: cursor stays at the offset where step 4 left it, which no longer
corresponds to any meaningful location in the restored text.

The reverse symptom appears for redo.

## Root cause

`EditorView.undo()` (and `redo()`):

```js
undo() {
    this._ctx.clearSelection();
    this.editBuffer.undo();
    this.requestRender();
    return true;
}
```

Compared to other mutating actions in the same file — e.g. `moveWordBackward`,
`deleteWordBackward` — which explicitly call
`this.editBuffer.setCursorByOffset(...)` after mutating the buffer, the
undo/redo paths skip that step entirely.

Looking one layer deeper, `EditBuffer.undo()` calls into native
(`this.lib.editBufferUndo(this.bufferPtr, maxSize)`) and returns the
decoded metadata bytes:

```js
undo() {
    this.guard();
    const maxSize = 256;
    const metaBytes = this.lib.editBufferUndo(this.bufferPtr, maxSize);
    if (!metaBytes) return null;
    return this.lib.decoder.decode(metaBytes);
}
```

The native side already produces metadata that includes the cursor position
recorded at the snapshot, but the JS layer discards it (or at least never
applies it back to the buffer's cursor state).

## Suggested fix

Either:

1. Have `EditBuffer.undo()` / `redo()` parse the metadata returned from
   the native call and apply the snapshot's cursor offset back to the
   buffer (preferred — preserves intuitive "undo restores the prior cursor
   too" behavior).
2. Or, at minimum, have `EditorView.undo()` / `redo()` call
   `setCursorByOffset` to clamp the cursor to a valid offset (e.g. end of
   buffer) so it never sits in nonsense positions, even if the precise
   pre-undo position is not preserved.

Option 1 matches what most editors do (Vim, VSCode, browsers all restore
cursor on undo) and avoids surprising users.

The `extmarks-history` wrapper around undo currently restores extmarks
metadata via `restoreSnapshot(snapshot)`; it could plausibly restore the
cursor offset too if the snapshot were extended to include it.

## Environment

- `@opentui/core` 0.1.103
- Reproduced via `opencode` 1.14.25+5a5a2e5 on Linux (NixOS, kernel
  6.19.11), zsh, but the bug is in `@opentui/core` and is host-independent.

## Why this matters

Several `opencode` issues describe symptoms that ultimately trace back to
this:

- anomalyco/opencode#15083 (closed without resolution, original reporter
  noted the undo issue is distinct from the cursor-after-paste duplicates
  it was bucketed with)
- anomalyco/opencode#17602 (closed by stale-bot, describes "overly
  aggressive command+z undo" — same root cause area)

Fixing it in `@opentui/core` resolves all of these in one place.

viralcodex:
Fixes: #980

Hi @simonklee @kommander,

This fixes the issue of cursor not moving according to the undo/redo texts. 

Implementation includes adding cursor position metadata in the rope nodes encoded as format `cursor:row:col:desired_col`

So we decode on undo and redo and then move the cursor based on the encoded rope metadata.

Let me know if any changes are needed or I missed something.

Thanks

https://github.com/user-attachments/assets/4771d97d-51aa-481e-aa5c-7ea149f89bd2


simonklee:
gpt found these. i have my head focused on some other things so i didn't think deeply about it but its worth to check out:

- `packages/core/src/zig/buffer.zig:1473`: when `drawTextBuffer` skips a wide grapheme that overflows the row, it only `break`s the current chunk loop and does not advance `currentX` or `column_in_line`. If the same visual line has another `VirtualChunk` after that grapheme, the next chunk can render at the stale edge column even though it should be off-screen. Example: a line built from `"aaa界"` plus an appended `"Z"` in a width-4 buffer would skip `界` at x=3, then draw `Z` at x=3. The overflow path should advance to/past the buffer edge or stop processing remaining chunks for the line.
- `packages/core/src/zig/grapheme.zig:540`: `GraphemeTracker.add` now ignores stale/invalid IDs, but callers like `OptimizedBuffer.setInternal` write `cell.char` before calling `add`. If this path is hit, the buffer can contain an invalid packed grapheme with no tracker entry, which breaks the invariant used by fast paths gated on `grapheme_tracker.hasAny()`. If stale IDs are expected to be recoverable, the write path needs a way to clear/skip the cell rather than only removing the ID from `used_ids`.

viralcodex:
> gpt found these. i have my head focused on some other things so i didn't think deeply about it but its worth to check out:
> 
> * `packages/core/src/zig/buffer.zig:1473`: when `drawTextBuffer` skips a wide grapheme that overflows the row, it only `break`s the current chunk loop and does not advance `currentX` or `column_in_line`. If the same visual line has another `VirtualChunk` after that grapheme, the next chunk can render at the stale edge column even though it should be off-screen. Example: a line built from `"aaa界"` plus an appended `"Z"` in a width-4 buffer would skip `界` at x=3, then draw `Z` at x=3. The overflow path should advance to/past the buffer edge or stop processing remaining chunks for the line.
> * `packages/core/src/zig/grapheme.zig:540`: `GraphemeTracker.add` now ignores stale/invalid IDs, but callers like `OptimizedBuffer.setInternal` write `cell.char` before calling `add`. If this path is hit, the buffer can contain an invalid packed grapheme with no tracker entry, which breaks the invariant used by fast paths gated on `grapheme_tracker.hasAny()`. If stale IDs are expected to be recoverable, the write path needs a way to clear/skip the cell rather than only removing the ID from `used_ids`.

Hi, isn't these for grapheme panic issue?

simonklee:
lol, yeh, sorry. as i said, head in different place :p

viralcodex:
> lol, yeh, sorry. as i said, head in different place :p

understandable, will fix and commit on the PR #1006

simonklee:
I didn't find anything that stood out in this pr, but need to do more testing locally.

simonklee:
undo/redo is sort of quirky.

viralcodex:
> I didn't find anything that stood out in this pr, but need to do more testing locally.

yep can have edge cases with CJK characters too. I can run some more tests if you want :)

kommander:
It should not have CJK edge cases, because the current cursor position should be CJK compliant already when it is stored and as the Rope history is a snapshot of the current state and would be restored exactly as it is the related cursor info at that point in time should be fully compatible. I think this is the correct and intended usage of Rope history meta data.

viralcodex:
Hmm, then i guess this implementation to store the cursor position in the rope metadata will work

> It should not have CJK edge cases, because the current cursor position should be CJK compliant already when it is stored and as the Rope history is a snapshot of the current state and would be restored exactly as it is the related cursor info at that point in time should be fully compatible. I think this is the correct and intended usage of Rope history meta data.

kommander:
@viralcodex are you on our opencode discord server? DM me.

viralcodex:
@kommander i am not actually on discord, have to create an account.

viralcodex:
@kommander, dm'd you on discord (avrl.dev)

PR #1019, opened 2026-05-05
feat(core): add Go syntax highlighting

anaskhan96:
## Summary
- Add Go as a default tree-sitter parser, including vendored WASM and highlight query assets.
- Support `go` and `golang` markdown code fence injections.
- Cover direct Go highlighting, alias preloading, and markdown Go fences in tree-sitter client tests.

## Test
- `bun test src/lib/tree-sitter/client.test.ts`

## Reference Images

(checked in a local build of [ghui](https://github.com/kitlangton/ghui))

image
image


kommander:
Hey, thanks for your interest. It is very simple for applications to support additional built-in parsers, like OpenCode does. The built-ins in opentui are limited on purpose and mainly for testing and the demos actually. I would prefer applications use the pattern that OpenCode uses to include parsers they want to provide out of the box. If that is not clear enough in the docs I'd be happy to make it clearer on how it works.

anaskhan96:
@kommander thanks for letting me know, the decision on taking this approach makes sense! closing this PR.

PR #1021, opened 2026-05-05
feat: qr code component

neriousy:
Hi there,

Here's a PR for adding a `QRCode` component for opentui

ISO COMPLAINT!! 
https://github.com/user-attachments/assets/ba7a1cd7-359a-46d1-bac3-b2d3dc70082f


simonklee:
This looks great but i would love to find a way in which this could be an external library and not something we need to support in OpenTUI initially. QR codes are cool and have many uses, but i'd much rather point people at your package than us taking on maintain this (we already have too much) 😉

neriousy:
Can you checkout the contrbutor's channel on discord? I did talk w/ @kommander there

simonklee:
> Can you checkout the contrbutor's channel on discord? I did talk w/ @kommander there

I see.

Adictya:
> but i'd much rather point people at your package 

+1 fwiw

kommander:
It's cool that this is fully spec driven. After consideration I have one concern: Opentui currently uses barrel index files, we might eventually move away from that, but for now it is like that. This means much of the code has to be parsed and interpreted at startup, influencing startup time. I would like to have a QR code component out of the box. So either this gets its own entrypoint, like the testing facilities have for example, so it is isolated and can be used as addon or we introduce the `@opentui/extras` package here in the monorepo and move it there. I'd actually prefer the extras package, where it can have its own entrypoint as well, so applications can opt in.

PR #1022, opened 2026-05-05
markdown: preserve spacing before code blocks

simonklee:
The token coalescing logic dropped blank lines preceding code
blocks, causing text and code to render without visual separation.
Preserve the gap in both the default and custom renderNode paths.


kommander:
The gap after a code block is also missing. Does it add this as well? Can't it be as simple as a marginTop+marginBottom for code blocks?

kommander:
Fixed in #1043

PR #1023, opened 2026-05-05
feat: latex rendering

neriousy:
Adding basic latex rendering 

https://github.com/user-attachments/assets/18b0c9a2-b075-46b1-95de-27f700660685


onyedikachi23:
I'm currently using OpenCode TUI in learning Linear Algebra and I really need this.

PR #1026, opened 2026-05-06
core/zig: added support as a Zig package

jettptacek:
## Support consumption as a Zig package

For all those that don't want to do any typescript 🫠
Lets downstream Zig projects pull `@opentui/core`'s Zig surface directly, without going through Bun, TypeScript, or the FFI layer.

I went down this route because multiple renders easily took all my CPU usage and when profiling, babel calls were the main issue. So why do I have to use typescript?

### Before

```zig
// downstream build.zig
const opentui = b.dependency("opentui", .{ .target = target, .optimize = optimize });
const mod = opentui.module("opentui");  // → null
```

`zig fetch --save` returned an empty package (`.paths` listed only
`build.zig` and `build.zig.zon`), and `addModule` was never called, so
there was nothing for downstreams to import.

### After

```zig
// downstream build.zig
const opentui = b.dependency("opentui", .{ .target = target, .optimize = optimize });
exe.root_module.addImport("opentui", opentui.module("opentui"));
```

```zig
// downstream code
const opentui = @import("opentui");
var pool = opentui.GraphemePool.init(allocator);
defer pool.deinit();
const buf = try opentui.OptimizedBuffer.init(allocator, 80, 24, .{ .pool = &pool });
```

### What changed

- **`opentui.zig`** (new): pure-Zig public surface. Re-exports `OptimizedBuffer`, `RGBA`, `CliRenderer`, `GraphemePool`, `EditBuffer`,`EditorView`, plus `buffer` / `renderer` / `grapheme` / `text_buffer` /`terminal` / `utf8` namespaces. `lib.zig` is still the entry point for TS/FFI consumers.
- **`build.zig.zon`**: expand `.paths` to ship every source file the public module reaches. I did Bump `.version` to `0.1.12`.
- **`build.zig`**: register `b.addModule("opentui", ...)`. Switch the hand-rolled `-Dtarget` string option to `b.standardTargetOptions(.{})`. Collapse `buildSingleTarget` and `buildNativeTarget` into a single `buildResolvedTarget`.
- **`tests/public-api_test.zig`** (new): compile-time pin of the consumer surface so renaming a re-exported type fails `zig build test`.

### Target options

The build.zig diff is bigger than expected because when a consumer does `b.dependency("opentui", .{ .target = target, .optimize = optimize })`, Zig serializes that one `target` parameter into **four** `-D` flags passed to the dependency's `build.zig`:

-Dtarget=...  -Dcpu=...  -Ddynamic_linker=...  -Dofmt=...

The old `b.option([]const u8, "target", ...)` only registered the first. The other three caused `error: invalid option: -Dcpu`. The standard fix is `b.standardTargetOptions(.{})`, which registers all four — but it conflicts with the existing string `target` option, so the old one had to be replaced. That cascaded into rewriting the string-typed `buildSingleTarget` / `buildNativeTarget` helpers in terms of `ResolvedTarget` (now `buildResolvedTarget`).

Maintainer-facing CLI is unchanged: `-Dtarget=x86_64-linux-gnu` and `-Dall` work the same as before.

### FYI!!!

The old `buildSingleTarget` did exact-string match on the full triple, so `-Dtarget=x86_64-linux-musl` fell through to the "custom target" path and produced `lib/x86_64-linux-musl/`. The new `buildResolvedTarget` matches on arch + OS only, so musl now lands in `lib/x86_64-linux/` alongside the gnu build. Output dir naming was already arch+OS-only (the gnu entry's `output_name = "x86_64-linux"`), so this is consistent with the existing naming convention.

I totally understand if these target changes are not wanted


jettptacek:
Ignore the Claude investigation I was doing on my phone.... I have a bunch native layout rendering working in my side project

jettptacek:
https://github.com/anomalyco/opentui/issues/821
Based this roadmap ^, I believe the `opentui.zig` public interface I'm introducing is not the proper path forward.  Once the renderable tree, layout engine and rendering moves to native code then maybe opentui as a zig package could be implemented

PR #1045, opened 2026-05-11
refactor(core): replace Bun import attributes with URL constructor

lu-zero:
## Summary
- Replace `import ... with { type: "file" }` with `new URL(path, import.meta.url)` + `fileURLToPath` in tree-sitter asset resolution and generated native package index.ts
- Minor cleanups in KeyHandler, EditBufferRenderable, and Textarea

The `with { type: "file" }` import attribute is a Bun-specific extension not part of the TC39 Import Attributes spec (which only defines `type: "json"`). Using `new URL()` + `fileURLToPath` produces the same resolved file paths and works across all runtimes.

## Test plan
- [x] `bun test` passes (58 tests across RGBA, border, clipboard, tree-sitter client)
- [x] `bun run fmt` and `bun run lint` pass
- [x] Native build (`bun run build --native`) produces correct artifacts
- [x] Generated `@opentui/core-{platform}-{arch}/index.ts` uses the new format

### DIFF
--- a/packages/core/scripts/build.ts
+++ b/packages/core/scripts/build.ts
@@ -136,9 +136,9 @@ if (buildNative) {
       continue
     }
 
-    const indexTsContent = `const module = await import("./${libraryFileName}", { with: { type: "file" } })
-const path = module.default
-export default path;
+    const indexTsContent = `import { fileURLToPath } from "node:url"
+const module = new URL("./${libraryFileName}", import.meta.url)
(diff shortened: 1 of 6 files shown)


simonklee:
Why did you open another?

lu-zero:
By mistake :) Sorry for the noise.

simonklee:
No worries.

PR #1046, opened 2026-05-11
Node.js runtime support for native core

simonklee:
Adapt the FFI and worker layers so the native Zig core runs on Node.js


simonklee:

PR #1049, opened 2026-05-11
Render markdown lists structurally

kitlangton:
## Summary

- Implements structured list rendering in `MarkdownRenderable` for top-level/document layout.
- Aligns nested fenced code under the nested bullet content column.
- Preserves source blank-line spacing around fenced code inside list items.
- Keeps default/coalesced list rendering unchanged.

## Before

Nested fenced code inside lists was rendered through source-text indentation and concealment, so it could appear over-indented:

```text
2. Second ordered item before a nested list:
   - Nested bullet before fenced code:

            const nested = true
```

## After

Lists are rendered structurally in top-level/document mode, with marker and content columns:

```text
2. Second ordered item before a nested list:
   - Nested bullet before fenced code:

     const nested = true

3. Third ordered item after the nested fence.
```

## Verification

Before rebasing onto latest `main`:

```text
```

```text
120 pass
0 fail
76 snapshots
```

After rebasing onto latest `main`, local JS tests are blocked by a stale native dylib/cache mismatch (`triggerNotification` missing) and local `bun run build:native:dev` currently fails with Zig/linker symbols on this machine. CI should build the native library from source.

Also checked the `Markdown Demo` manually with `tuistory` before the rebase.


kommander:
Streaming list rendering flickers, probably because of list renderable recreations. It must not destroy and recreate for streaming content. Need to look into that.

https://github.com/user-attachments/assets/99469581-7d2e-4bc6-b0c9-e5c3b2038e8f

PR #1051, opened 2026-05-11
Improve markdown document rendering

kitlangton:
## Summary

- Exposes markdown document rendering options through public setters so Solid/React props can update `MarkdownRenderable` after construction.
- Improves MarkdownRenderable document output for blockquotes, horizontal rules, top-level spacing, and table padding controls.
- Adds `tree-sitter-diff` to the default parser set so fenced `diff` blocks use the normal `CodeRenderable` highlighting path.
- Expands the Markdown demo and adds snapshot coverage for assistant-style markdown layout and structural block spacing.

## Before

Top-level document output could collapse or over-expand block boundaries, and some rendered structures were still source-like:

```text
Start with prose before a code block.
export function parse(input: string) {
  return input.trim().split(/\s+/)
}
Then continue with prose after the code block.

> Quoted note after the list.

---
```

Markdown tables were already parsed and rendered, but table cell padding was all-or-nothing. There was no way to add horizontal breathing room without also adding vertical padding:

```text
┌─┬─┐
│A│B│
├─┼─┤
│1│2│
└─┴─┘
```

## After

Top-level document rendering now has explicit, tested spacing around structural markdown blocks:

```text
Start with prose before a code block.

export function parse(input: string) {
  return input.trim().split(/\s+/)
}

Then continue with prose after the code block.

│ Quoted note after the list.

────────────────────────────────────────────────────────────
```

Tables can request horizontal-only padding:

```text
┌───┬───┐
│ A │ B │
├───┼───┤
│ 1 │ 2 │
└───┴───┘
```

Fenced diff blocks now use the default tree-sitter highlighting path:

```diff
- const renderer = oldMarkdown
+ const renderer = experimentalMarkdown
```

## Verification

```text
```

```text
118 pass
0 fail
74 snapshots
```

```text
bunx oxlint packages/core/src/lib/tree-sitter/parsers-config.ts packages/core/src/lib/tree-sitter/default-parsers.ts packages/core/src/lib/tree-sitter/assets/update.ts packages/core/src/renderables/Markdown.ts packages/core/src/renderables/__tests__/Markdown.test.ts packages/examples/src/markdown-demo.ts
```

Also manually checked the updated `Markdown Demo` with `tuistory`.

## Notes

- Markdown tables already rendered before this PR; this only adds horizontal/vertical padding controls and updates the demo styling.
- Nested list rendering is still not fully structural; that should be handled by a future dedicated list renderer rather than source-string indentation fixes.
- Ghostty links require Shift+Cmd-click in mouse-capturing fullscreen TUIs; OpenTUI is emitting hyperlink metadata, but the terminal must bypass mouse capture.

Restacked from #1043 onto `main` after parent merge.

### [Stack](https://github.com/kitlangton/stack)

1. #1043
2. **#1051** 👈 current
3. #1049


kitlangton:
Closing duplicate stack parent. #1043 already merged the markdown document rendering changes.

PR #1052, opened 2026-05-11
examples: add Node.js 26 runtime support

simonklee:
Allow examples to run under Node 26 with --experimental-ffi in addition to Bun. Replace some Bun-specific APIs with standard Node equivalents, lazy-load demos that still require Bun (WebGPU, sprites, shaders), and import Rapier via its ESM subpath.


simonklee:

PR #1054, opened 2026-05-12
build: target Node.js for portable dist output

simonklee:
Add conditional exports (bun/node/default) so each runtime resolves the correct variant, and add a packed dist smoke test that verifies imports work in both Node.js and Bun.


simonklee:

PR #1055, opened 2026-05-12
Debun tests

simonklee:
Tests keep real bun:test imports;
a module hook rewrites them only in the emitted Node lane.


simonklee:

PR #1060, opened 2026-05-13
feat(deno): Deno runtime support

lu-zero:
## Summary

Adds Deno runtime support to opentui, building on top of the `debun-urls` branch (PR #1027). With this, opentui core runs under Deno with native FFI, tree-sitter highlighting, and test infrastructure.

### What's included

- **Root `deno.json`** with sloppy-imports and `deno task test:core` / `test:solid`
- **Core `deno.json`** import map: `bun:ffi` → `bun-compat@0.4.2`, `bun:test` → `@std/testing` shim, `@opentui/core` → src, node: builtins
- **`bun-test.ts` shim** delegating to `@std/testing/bdd`, `@std/expect`, `@std/expect/fn` with `spyOn`/`mockImplementation` compatibility and auto-installing `bun-compat` for `globalThis.Bun`
- **Solid `deno.json`** with `npm:solid-js` specifiers and `jsx-runtime` mapping
- **`jsx-runtime.ts`** for Deno-compatible JSX transformation
- **Module workers** (`{ type: "module" }`) for Deno compat
- **WASM fallback** from `with { type: "wasm" }` to `import.meta.resolve` + URL
- **FFI `bun:ffi` fallback** when `node:ffi` is unavailable (bun-compat)
- **Timer/worker cleanup** for Deno's strict leak detection
- **`isBunRuntime` check** in `resolveBundledFilePath` to distinguish real Bun from bun-compat shim
- **Platform shim** for native binary path resolution using `import.meta.url`

### Test results

**Deno**: 225 passed (1349 steps), 3 failed
- 2 failures are Deno leak detection strictness (signal handlers, FFI resources)
- 1 failure uses `Bun.serve` (Bun-only test)

**Bun**: 58/58 pass (no regressions)

### How to test

```bash
# Run core lib tests under Deno
deno test --allow-all --no-check --config packages/core/deno.json packages/core/src/lib/

# Run a demo under Deno
deno run --allow-all --config packages/core/deno.json packages/examples/src/console-demo.ts
```

### Dependencies

- Based on `debun-urls` (PR #1027) — should be merged first
- Requires `bun-compat@0.4.2+` (published to jsr.io)

### Known limitations

- `--no-check` needed (type errors from Deno stdlib mismatches)
- 3 test failures due to Deno's stricter resource leak detection
- Solid tests need further reconciler/renderer lifecycle work under Deno


simonklee:
Thanks for this ­— but you're a bit too early. We need to go through refactoring to support `node:ffi` first, and only then we can consider another platform. Adding multiple platforms at once will just cause too much chaos.

lu-zero:
I'll rebase on top of the node:ffi support once it is ready then :)

simonklee:
> I'll rebase on top of the node:ffi support once it is ready then :)

I may cherry pick a couple of the the things you have here, tbh. I think the `resolveBundledFilePath` may be better than what i have and i haven't gotten around to the solid.js part yet. So i think in the end you'll need way less for Deno support.

simonklee:
Thank you for the contribution. This draft uses an old runtime design. The
current main branch already includes the Node FFI, worker, asset, and Solid JSX
infrastructure. This branch removes native pointer checks and fails an FFI
test. Deno support is not yet available and not currently planned.

lu-zero:
I'll check if what is present this weekend, but probably it should work out of box already.

PR #1067, opened 2026-05-14
notifications: remove OSC 99 from startup probes

simonklee:
Sending OSC 99 queries during generic startup can crash unpatched
foot due to a bug in its query reply path. Restrict OSC 99 to
terminals identified by name (kitty, foot) via XT_VERSION
heuristics rather than broadcasting the probe to all terminals.

foot moves from OSC 777 to OSC 99 in the identity-based detection,
matching its documented protocol support.

Issue https://github.com/anomalyco/opencode/issues/27418


yangyingchao:
thanks

PR #1073, opened 2026-05-16
solid: add JSX runtime and Node test lanes

simonklee:
Add jsx/jsxs/jsxDEV/Fragment runtime implementations, split the build into Node and Bun entry points, and gate CI on Bun tests, Node source tests, and a packed-consumer smoke test that verifies the tarball installs and runs under Node.


simonklee:

PR #1076, opened 2026-05-17
Improve CodeRenderable streaming performance

kitlangton:
## Summary
- Coalesce in-flight `CodeRenderable` highlights during streaming so rapid appends no longer enqueue one Tree-sitter worker parse per chunk.
- Add cached restyling for style-only changes (`conceal`, `syntaxStyle`, `drawUnstyledText`, `baseHighlight`, `onChunks`) so they re-chunk existing highlights instead of reparsing.
- Add a small `highlightOnce` LRU and streaming display cache, plus provisional lexical coloring for the active streaming tail to avoid color flicker and the final plain-text flash when streaming ends.
- Add a benchmark harness and recorded benchmark artifacts for the before/after numbers.

## Performance
Benchmark command:

```bash
cd packages/core
bun src/benchmark/code-benchmark.ts --runs=9 --warmup=1 --json=/tmp/code-latest.json
```

Baseline was `main @ b2c885d9`; latest run was this branch after the streaming color/no-flash/simplify fixes.

| Metric | Baseline | Latest | Change |
| --- | ---: | ---: | ---: |
| `b1_single_shot_ms` | 15.82 ms | 12.36 ms | -22% |
| `b1_single_shot_msgs` | 1 | 0 | cache hit |
| `b1_single_shot_bytes` | 6,270 B | 0 B | cache hit |
| `b2_streaming_total_ms` | 472.52 ms | 61.24 ms | -87% |
| `b2_streaming_settle_ms` | 456.28 ms | 27.07 ms | -94% |
| `b2_streaming_msgs` | 200 | 1 | -99.5% |
| `b2_streaming_bytes` | 2.06 MB | 6.09 KB | -99.7% |
| `b3_conceal_toggle_msgs` | 1 | 0 | no worker parse |
| `b3_syntaxstyle_swap_msgs` | 1 | 0 | no worker parse |
| `b4_content_reset_msgs` | 2 | 0 | cache hit |
| `b4_conceal_pingpong_msgs` | 2 | 0 | no worker parse |

## Scope
- Biggest win is streaming fenced-code / AI-output scenarios: fewer worker round-trips, far less posted data, and much less settle work after streaming stops.
- Non-streaming also benefits from the LRU when the same content/filetype is highlighted again.
- Style-only updates benefit outside streaming too because they now reuse cached highlights and only re-run chunk generation.
- This does not implement stateful incremental Tree-sitter edits; that is deferred until a larger-file benchmark shows it is needed.

## UX Notes
- Active streaming text uses provisional lexical coloring for stable-looking colors while syntax is incomplete; it is not time-debounced.
- Completed lines and final output still use Tree-sitter highlights.
- Ending streaming now reuses cached highlights or preserves the current styled frame while the final highlight is pending, avoiding the final all-white flash.

## Simplify Follow-up
- Switched the benchmark CLI to the existing `commander` pattern.
- Made the one-shot Tree-sitter cache safer: compact hash keys, defensive highlight copies, and invalidation on parser/data/cache changes.
- Centralized `CodeRenderable` cache cleanup and guarded async highlight application against style-state changes while callbacks are pending.

## Tests
- `bun test src/renderables/Code.test.ts`
- `bun src/benchmark/code-benchmark.ts --runs=9 --warmup=1 --json=/var/folders/dd/5fz89drs5p9_r0fk7rwqqnbr0000gn/T/opencode/code-latest-simplified.json`
- `bunx oxfmt --check src/renderables/Code.ts src/renderables/Code.test.ts src/lib/tree-sitter/client.ts src/benchmark/code-benchmark.ts`
- `bun run lint:ci`


kitlangton:
Closing this rather than continuing the current implementation. The investigation found useful opportunities around avoiding reparses for style-only changes and reducing obsolete streaming highlight work, but this branch grew too broad and has unresolved correctness risks (including the hash-keyed highlight cache and stalled in-flight highlight behavior).

I will revisit the useful pieces as smaller, separately verified changes on current `main`.

PR #1077, opened 2026-05-17
Fix(solid/reconciler): incorrect insert order in Show with fragments

Adictya:
### Bug: incorrect insert order in \<Show\/\> with fragments
```tsx
<box id="top" />
<Show when={visible()}>
  <>
    <box id="a" />
    <box id="b" />
    <box id="c" />
  </>
</Show>
<box id="bottom" />
```
Toggling the Show content off and back on could lead to the wrong child order:
["top", "bottom", "a", "b", "c"]
instead of:
["top", "a", "b", "c", "bottom"]

### Cause
getNextChild in the reconciler was resolving the abstract SlotNode instead of the concrete layout slot node.
That missed the sibling order check, so reinserted children were appended in the wrong place.
### Fix
Resolve slot nodes to their concrete layout slot node before checking sibling order.

### DIFF
--- a/packages/solid/src/reconciler.ts
+++ b/packages/solid/src/reconciler.ts
@@ -376,8 +376,15 @@ export const {
       log("No parent found for node:", logId(node))
       return undefined
     }
-    const siblings = getNodeChildren(parent)
 
+    if (node instanceof SlotRenderable) {
+      const layoutSlotNode = node.getSlotChildForRemoval(parent)
+      if (layoutSlotNode) {
+        node = layoutSlotNode
+      }
+    }
+
+    const siblings = getNodeChildren(parent)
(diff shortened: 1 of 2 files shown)


Adictya:
The slotNode implementation without rigid boundaries of universal renderer only operating on physical nodes rather than logical is creating a lot of edge cases. Ill see if i can refactor this to be more consistent and make the abstraction non leaky

simonklee:
> The slotNode implementation without rigid boundaries of universal renderer only operating on physical nodes rather than logical is creating a lot of edge cases. Ill see if i can refactor this to be more consistent and make the abstraction non leaky

That would be awesome

simonklee (approved):

PR #1088, opened 2026-05-18
fix(core): guard recalculateBarProps sticky scroll with _hasManualScroll

### LINKED ISSUE #1087: [BUG] recalculateBarProps else branch ignores _hasManualScroll, sticky scroll fights user during streaming
## Symptom

When `stickyScroll=true` and `stickyStart="bottom"`, scrolling up during content growth (streaming) causes the viewport to fight the users scroll position. The fix in PR #722 (opencode issue #709) addressed the `updateStickyState` path but left two remaining issues in `recalculateBarProps`.

## Root Cause 1: Missing `_hasManualScroll` guards in `recalculateBarProps` else branch

In `recalculateBarProps`, the `_stickyStart && !_hasManualScroll` branch is correctly guarded. But the `else` branch still checks `_stickyScrollBottom` and forces `scrollTop = newMaxScrollTop` without checking `_hasManualScroll`:

```js
if (this._stickyStart && !this._hasManualScroll) {
  this.applyStickyStart(this._stickyStart);
} else {
  if (this._stickyScrollTop) {
    this.scrollTop = 0;
  } else if (this._stickyScrollBottom && newMaxScrollTop > 0) {
    this.scrollTop = newMaxScrollTop;
  }
```

`_stickyScrollBottom` is set to true in `updateStickyState` whenever `scrollTop >= maxScrollTop`. During streaming, content grows, the scrollbar onChange fires, and if the position is near the bottom, the flag flips back to true. The next `recalculateBarProps` cycle forces the scroll to bottom again.

## Root Cause 2: `_hasManualScroll` never resets when user scrolls back to sticky edge

Once `_hasManualScroll` is set to true by scrolling up, it stays true even when the user scrolls back to the bottom. The `updateStickyState` reset is blocked by `_isApplyingStickyScroll` during content growth cycles. The scrollbar onChange handlers also do not reset the flag.

## Fix

1. Add `!this._hasManualScroll` guards to the else branch in `recalculateBarProps`:

```js
} else {
  if (this._stickyScrollTop && !this._hasManualScroll) {
    this.scrollTop = 0;
  } else if (this._stickyScrollBottom && !this._hasManualScroll && newMaxScrollTop > 0) {
    this.scrollTop = newMaxScrollTop;
  }
```

2. Add a re-engagement path that resets `_hasManualScroll` when the user scrolls back to the sticky edge. Use a threshold (not exact match) because streaming content growth makes exact position unreliable:

```js
} else if (this._hasManualScroll && newMaxScrollTop > 0 && this.scrollTop >= newMaxScrollTop - 1) {
  this._hasManualScroll = false;
  this.applyStickyStart(this._stickyStart);
```

## Verification

Tested on opencode 1.15.4 (opentui 0.2.11) -- issue persists. The PR #722 fix resolved the `updateStickyState` path but not these two remaining code paths. Both fixes verified working in a patched local build.

malventano:
### Issue for this PR

Fixes #1087

### Type of change

- [x] Bug fix

### What does this PR do?

Two remaining issues in ScrollBox sticky scroll during streaming after PR #722:

1. The `else` branch in `recalculateBarProps` forced `scrollTop`/`scrollLeft` even when the user had manually scrolled away, because `_hasManualScroll` guards were missing. This caused the viewport to fight the user during streaming.

2. `_hasManualScroll` was never reset when the user scrolled back to the sticky edge during streaming, preventing sticky scroll from re-engaging.

Fix:
- Add `!this._hasManualScroll` guards to all four sticky position assignments in the else branch
- Add a re-engagement path that resets `_hasManualScroll` and calls `applyStickyStart` when the user scrolls back to the bottom (using a 1-line threshold since exact position is unreliable during content growth)

Tested in a patched local build of opencode (1.15.4 with opentui 0.2.11). Verified:
- Scrolling up during streaming stays put, no fight
- Scrolling back to bottom re-engages sticky scroll
- Normal sticky behavior when not manually scrolled is unchanged

### Screenshots / recordings

N/A -- terminal behavior

### Checklist

### DIFF
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -766,16 +766,20 @@ export class ScrollBoxRenderable extends BoxRenderable {
 
         if (this._stickyStart && !this._hasManualScroll) {
           this.applyStickyStart(this._stickyStart)
+        } else if (this._hasManualScroll && newMaxScrollTop > 0 && this.scrollTop >= newMaxScrollTop - 1) {
+          // User scrolled back to bottom during streaming -- re-engage sticky
+          this._hasManualScroll = false
+          this.applyStickyStart(this._stickyStart)
         } else {
-          if (this._stickyScrollTop) {
+          if (this._stickyScrollTop && !this._hasManualScroll) {
             this.scrollTop = 0
-          } else if (this._stickyScrollBottom && newMaxScrollTop > 0) {
+          } else if (this._stickyScrollBottom && !this._hasManualScroll && newMaxScrollTop > 0) {
             this.scrollTop = newMaxScrollTop
           }
 
-          if (this._stickyScrollLeft) {
+          if (this._stickyScrollLeft && !this._hasManualScroll) {
             this.scrollLeft = 0
-          } else if (this._stickyScrollRight && newMaxScrollLeft > 0) {
+          } else if (this._stickyScrollRight && !this._hasManualScroll && newMaxScrollLeft > 0) {
             this.scrollLeft = newMaxScrollLeft
           }
         }
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -521,6 +521,23 @@ export class ScrollBoxRenderable extends BoxRenderable {
     }
   }
 
+  private isAtStickyReengagePoint(
+    stickyStart: "bottom" | "top" | "left" | "right",
+    maxScrollTop: number,
+    maxScrollLeft: number,
+  ): boolean {
+    switch (stickyStart) {
+      case "top":
+        return maxScrollTop > 0 && this.scrollTop <= 0
+      case "bottom":
+        return maxScrollTop > 0 && this.scrollTop >= maxScrollTop - 1
+      case "left":
+        return maxScrollLeft > 0 && this.scrollLeft <= 0
+      case "right":
+        return maxScrollLeft > 0 && this.scrollLeft >= maxScrollLeft - 1
+    }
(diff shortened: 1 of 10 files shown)


packages/core/src/renderables/ScrollBox.ts
```diff
@@ -766,16 +766,20 @@ export class ScrollBoxRenderable extends BoxRenderable {
 
         if (this._stickyStart && !this._hasManualScroll) {
           this.applyStickyStart(this._stickyStart)
+        } else if (this._hasManualScroll && newMaxScrollTop > 0 && this.scrollTop >= newMaxScrollTop - 1) {
+          // User scrolled back to bottom during streaming -- re-engage sticky
+          this._hasManualScroll = false
+          this.applyStickyStart(this._stickyStart)
         } else {
```
Lauri-Nomme: i noticed every child branch has `!this._hasManualScroll)` guard, maybe add it here at this level as an `else if (!this._hasManualScroll)` instead?

kommander:
A regression test reproducing the issue would be greatly appreciated.

packages/core/src/renderables/ScrollBox.ts
malventano: Agreed and updated.

malventano:
> A regression test reproducing the issue would be greatly appreciated.

Getting the test to catch this specific type of behavior within the test framework is a bit tricky. Currently iterating a few methods.

*edit* ok, should be good. Note that the second test I added will pass without this PR, because it's checking for re-engagement, but if the box keeps scrolling anyway (the bug), it fools the test. Either way re-engagement should be tested.

PR #1092, opened 2026-05-18
native: add musl libc prebuilds for Linux

simonklee:
The existing Linux prebuilds are linked against glibc and fail to load on musl-based distros such as Alpine. Ship dedicated musl binaries alongside them and detect the host libc at load time so opentui transparently picks the right one.


packages/core/src/zig.ts
```diff
@@ -62,15 +63,20 @@ import type {
 } from "./zig-structs.js"
 import { isBunfsPath } from "./lib/bunfs.js"
 
-const nativePackage = await import(`@opentui/core-${process.platform}-${process.arch}`)
+const nativePackageName = getNativePackageName()
```
kommander, line 66: I think this prevents standalone executable builds from bundling the correct native lib, because it cannot be statically analysed.

simonklee: Yeh, i was thinking about what the best way to this thing is. Wasn't sure we wanted musl either so was going to ask you before i cleaned this up.

kommander: If it helps and doesn't make it too complicated for us to provide it, why not?

PR #1093, opened 2026-05-18
feat: add minLength prop for InputRenderable

### LINKED ISSUE #1028: html-like props in input tag for solid and react (and input renderable as well)
Hi @simonklee / @kommander ,

Currently input renderable doesn't have any native prop for type = (number, text, email, etc).

Could we have them added? if yes, I can raise a PR to add this basic stuff.
Thanks

viralcodex:
Fix: #1028 

Hi @kommander / @simonklee ,

This is for adding more props to input tags to give more options to users (minLength, and input type)

Currently only text (default), number and email I have added, because for password there's already a PR to add to that and if we can wire that in (redaction is the main thing/change there)

My proposal is to have a form based input/textarea renderables natively in OpenTUI as we can build on this.

Would love to discuss this further on discord/here.

Thanks


packages/core/src/renderables/Input.test.ts
```diff
@@ -858,74 +858,6 @@ describe("InputRenderable", () => {
     expect(input.value).toBe("initialb")
   })
 
-  it("should handle Kitty keypad digits, operators, and enter", async () => {
```
kommander, line 861: why would that test be removed?

viralcodex: ahh by mistake while removing type, fixed it and added a test for minLength validation as well

Issue #11, opened 2025-08-13
Errors when building


aaronshaf:
```
ompat@0.17.3
+ @opentui/core-darwin-arm64@0.1.1
+ bun-webgpu@0.1.0
+ planck@1.4.2
+ three@0.177.0
+ @types/bun@1.2.19
+ @types/three@0.177.0
+ commander@13.1.0
+ prettier@3.6.2
+ jimp@1.6.0
+ yoga-layout@3.2.1

92 packages installed [121.00ms]

 …/opentui   main  bun run build
$ bun run build:native && bun run build:lib
$ node scripts/build.mjs --native
Building native prod binaries...
Built: @opentui/core-darwin-x64
Built: @opentui/core-darwin-arm64
Built: @opentui/core-linux-x64
Built: @opentui/core-linux-arm64
Built: @opentui/core-win32-x64
Built: @opentui/core-win32-arm64
$ node scripts/build.mjs --lib
Building library...
Bundled 39 modules in 9ms

  index.js  0.44 MB  (entry point)

Bundled 238 modules in 44ms

  3d.js                        1.39 MB  (entry point)
  supersampling-jw3fem06.wgsl  6.66 KB  (asset)

Generating TypeScript declarations...
node_modules/bun-webgpu/src/ffi.ts:8:33 - error TS2345: Argument of type 'URL' is not assignable to parameter of type 'string | URL'.
  Type 'URL' is not assignable to type 'import("url").URL'.
    Types of property 'searchParams' are incompatible.
      Type 'URLSearchParams' is missing the following properties from type 'URLSearchParams': entries, keys, values, [Symbol.iterator]

8 const __dirname = fileURLToPath(new URL(".", import.meta.url));
                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/bun-webgpu/src/GPUAdapter.ts:279:68 - error TS2345: Argument of type 'GPUDeviceDescriptor & { uncapturedErrorCallbackInfo: WGPUUncapturedErrorCallbackInfo; deviceLostCallbackInfo: { mode: "
  Types of property 'requiredFeatures' are incompatible.
    Type 'Iterable<GPUFeatureName> | undefined' is not assignable to type 'Iterable<"force-32" | "depth-clip-control" | "depth32float-stencil8" | "timestamp-query" | "texture-compression-bc" | "texture-
[...]

PR #1095, opened 2026-05-19
fix(utf8): treat U+26A0 as width 1 per UAX #11

aarcamp:
[DerivedEastAsianWidth](https://www.unicode.org/Public/16.0.0/ucd/extracted/DerivedEastAsianWidth.txt) classifies U+26A0 WARNING SIGN as N (Neutral). It appears in emoji-data.txt as Emoji=Yes but Emoji_Presentation=No, so by UTS #51 its default presentation is text (emoji presentation requires VS16, U+FE0F). The hardcoded override in eawToWidth contradicts both and causes a +1-cell drift in column accounting after bare ⚠, which misaligns any downstream fixed-column content on the same row (e.g.  tmux pane borders).

```
26A0          ; N # So       WARNING SIGN
```

The removed line also covered U+26A1, but that codepoint keeps width 2 via the earlier EAW=W check, so behavior there is unchanged.

Before (note misaligned right-side tmux pane border):

caution-before

After:

caution-after

(These are screenshots of Codex CLI running in auto-review mode.)

### DIFF
--- a/packages/core/src/zig/tests/unicode-width-map.zon
+++ b/packages/core/src/zig/tests/unicode-width-map.zon
@@ -1830,7 +1830,7 @@
     .{ .codepoint = "U+269D", .width = 1 },
     .{ .codepoint = "U+269E", .width = 1 },
     .{ .codepoint = "U+269F", .width = 1 },
-    .{ .codepoint = "U+26A0", .width = 2 },
+    .{ .codepoint = "U+26A0", .width = 1 },
     .{ .codepoint = "U+26A1", .width = 2 },
     .{ .codepoint = "U+26A2", .width = 1 },
     .{ .codepoint = "U+26A3", .width = 1 },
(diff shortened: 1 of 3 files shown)


simonklee:
Can you also check how this looks in other terminals you have available. We don't implement the Unicode standard on purpose since terminals don't implement it when it comes to determining width of grapheme clusters

simonklee:
See https://github.com/anomalyco/opentui/pull/756

aarcamp:
> Can you also check how this looks in other terminals you have available. We don't implement the Unicode standard on purpose since terminals don't implement it when it comes to determining width of grapheme clusters

Screenshots in the PR description are from latest Ghostty (macOS).

iTerm2 (same symptom):

image

Kitty (same symptom):

image

WezTerm (same symptom):

image

I don't have a screenshot but same thing with GNOME terminal (VTE) on a Fedora 44 machine.

All look correct after applying the fix.

kommander:
I think this should be fine, I've seen that before. Lets see how it goes.

PR #1097, opened 2026-05-20
fix: normalize numpad key names to standard equivalents

### LINKED ISSUE #7698: Numpad Enter should have the same behavior as the regular Enter
### Description

When I press the Numpad Enter, it doesn't submit the prompt.

### Plugins

_No response_

### OpenCode version

1.1.11

### Steps to reproduce

_No response_

### Screenshot and/or share link

_No response_

### Operating System

macOS 26.2

### Terminal

Ghostty

### LINKED ISSUE #17457: Numpad Enter key does not send messages in TUI
### Description

Pressing the numpad Enter key in the OpenCode TUI does not submit messages or confirm dialogs. Only the main keyboard Enter key works.

### Why this happens

Terminals using the Kitty keyboard protocol report numpad Enter as `kpenter` (codepoint 57414), which is a different key name from `return` (the main Enter key). The TUI currently only checks for `evt.name === "return"` in all keyboard handlers and textarea keybindings, so `kpenter` is never matched.

### Expected behavior

Both the main Enter key and the numpad Enter key should submit messages, confirm dialogs, and select options identically.

### Suggested fix

Add `kpenter` alongside `return` in all keyboard event checks and textarea keybindings throughout the TUI.

### LINKED ISSUE #27138: NumLock number keys and special keys not working in opencode CLI
### Description

When typing inside opencode CLI, NumLock number keys (0-9) and other special keys (like *, +, -, ., /) do not work. Typing produces no character or an incorrect one.

Expected Behavior:
All NumLock keys should input the correct characters, just like any other terminal application.

Actual Behavior:
Pressing NumLock number keys or special keys produces no output or wrong output in opencode's input.

Environment:
- OS: Windows 11 / 10
- Terminal: PowerShell 5.1 (or your terminal)
- opencode version: (run opencode --version)
- Model: deepseek-v4-flash-free
-
Comparison:
The same terminal works correctly with other CLI tools (e.g., gemini CLI). NumLock keys work fine there. The issue is specific to opencode.

### Plugins

_No response_

### OpenCode version

1.14.48

### Steps to reproduce

1. Run opencode in PowerShell
2. Turn on NumLock
3. Try typing number keys (0-9) or special keys like *, /, +, -
4. Observe that nothing or wrong character appears

### Screenshot and/or share link

_No response_

### Operating System

Windows 11

### Terminal

_No response_

Marenz:
Numpad keys (kp0-kp9, kpenter, kpdecimal, kpdivide, kpmultiply, kpminus, kpplus, kpequal, kpseparator) are reported with distinct names by the kitty keyboard protocol. This causes them to not match keybindings or direct name checks (e.g. `evt.name === "return"` fails for numpad enter).

Normalize numpad key names in `KeyHandler.processParsedKey()` before emitting events, so all downstream handlers see standard key names regardless of whether the key was pressed on the main keyboard or numpad.

The `printableKeypadText` map already existed in the kitty parser to set `key.sequence`, but `key.name` was left as the kp-prefixed variant. This meant text insertion worked in newer versions but keybinding matching and direct name checks still failed.

Fixes anomalyco/opencode#7698
Fixes anomalyco/opencode#17457
Fixes anomalyco/opencode#27138

### DIFF
--- a/packages/core/src/lib/KeyHandler.test.ts
+++ b/packages/core/src/lib/KeyHandler.test.ts
@@ -636,6 +636,74 @@ test("KeyHandler - internal handler error with preventDefault still respects pre
   expect(internalCalled).toBe(false)
 })
 
+test("KeyHandler - numpad keys are normalized to standard names", () => {
+  const handler = createKeyHandler()
+
+  let receivedKey: KeyEvent | undefined
+  handler.on("keypress", (key: KeyEvent) => {
+    receivedKey = key
+  })
+
+  // Simulate a kitty protocol numpad 5 key (codepoint 57404)
+  dispatchInput(handler, "\x1b[57404u", { useKittyKeyboard: true })
+
+  expect(receivedKey).toBeDefined()
+  expect(receivedKey?.name).toBe("5")
+  expect(receivedKey?.sequence).toBe("5")
+})
+
+test("KeyHandler - numpad enter is normalized to return", () => {
+  const handler = createKeyHandler()
+
+  let receivedKey: KeyEvent | undefined
+  handler.on("keypress", (key: KeyEvent) => {
+    receivedKey = key
+  })
+
+  // Simulate a kitty protocol numpad enter (codepoint 57414)
+  dispatchInput(handler, "\x1b[57414u", { useKittyKeyboard: true })
+
+  expect(receivedKey).toBeDefined()
+  expect(receivedKey?.name).toBe("return")
+})
+
+test("KeyHandler - numpad decimal is normalized to dot", () => {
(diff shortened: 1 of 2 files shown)


simonklee:
@opentui/keymap can handle this at the binding layer, and doing it in KeyHandler is a broader public behavior change which makes it impossible to build an application that cares about numpad identity. keymap resolver in opencode can solve this without changing the lower layer keyinput event.

kommander:
Yeah the keynames should not be normalized here, some apps might want to handle keypad keys explicitly and differently, while the @opentui/keymap can use aliasing if it wants to.

PR #1098, opened 2026-05-20
Chore(solid): Refactor and simplify reconciler slot nodes

Adictya:
WIP, not sure of the direction just experimenting


packages/solid/src/elements/slot.ts
```diff
@@ -1,393 +0,0 @@
...
-type LayoutNodeProvider = {
-  getLayoutNode?: () => Yoga.Node
-}
-
-type LayoutNodeConstructor = { create?: () => Yoga.Node } | undefined
-
-function getLayoutNodeConstructor(parent?: BaseRenderable): LayoutNodeConstructor {
-  const parentLayoutNode = (parent as LayoutNodeProvider | undefined)?.getLayoutNode?.()
-  return parentLayoutNode?.constructor as LayoutNodeConstructor
-}
```
Adictya, line 12: @simonklee Hey, whats the usecase for making this modular? What other constructors are there. Evaluating on how to port this over

Adictya:
overall, i’m still not fully sure which direction to go in here.

this pr tries to realign the behavior expectation to be closer to solid’s side of things, where text nodes “just work” naturally. so instead of having a dynamic factory slot node, it becomes a more homogeneous node that can be attached in both layout and text nodes.

however, to support this, text nodes now need the ability to do parent/child/anchor management, which they previously had no need to handle. so we’re effectively transferring some of that complexity into text nodes.

the goal of this effort was to reduce and eliminate future edge cases by consolidating the added complexity and reassessing the implementation. i do feel this refactor has better conceptual alignment and better complexity localization, but it also shifts the likely source of edge cases from the reconciler to text nodes.

the reconciler has at least been tested enough by now, and the implementation has already been adjusted to plug a bunch of those holes. surfacing an entirely new module as the place where this complexity lives might introduce a fresh class of regressions.

i’m also trying another more minimal refactor here: [https://github.com/Adictya/opentui/pull/3](https://github.com/Adictya/opentui/pull/3)

still torn between the two directions though. let me know if anyone has other thoughts.

PR #1103, opened 2026-05-22
fix(core): Preserve empty lines when copying selected text

benkoppe:
OpenTUI currently completely drops empty logical lines from selected text. This means highlight-copying code can remove intentional blank lines, which often breaks formatting.

This affects Opencode heavily, making copying/pasting code excessively annoying because all blank lines are completely erased.

The behavior was caused by `extractTextBetweenOffsets`. Newline output was gated on whether the line had content, therefore empty selected lines never emitted their newline. This PR changes newline emission to be based on whether the newline offset is inside the selected range, regardless of line content.

 Includes a regression test for:

 ```txt
const a = 1

const b = 2
 ```

copying back exactly as selected.

### DIFF
--- a/packages/core/src/renderables/Text.test.ts
+++ b/packages/core/src/renderables/Text.test.ts
@@ -144,10 +144,11 @@ describe("TextRenderable Selection", () => {
       const selection = text.getSelection()
       expect(selection).not.toBe(null)
       // With newline-aware offsets: Line 0 (0-5) + newline (6) + Line 1 (7-12) + newline (13) + Line 2 empty (14)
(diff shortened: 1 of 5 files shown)


benkoppe:
This issue seems to be cited in Opencode multiple times:

https://github.com/anomalyco/opencode/issues/10098
https://github.com/anomalyco/opencode/issues/16521
https://github.com/anomalyco/opencode/issues/25623

benkoppe:
Thanks so much @kommander , can't tell you how much this will improve my workflow!

PR #1108, opened 2026-05-25
core(renderer): handle ANSI text written to stdout in split-footer mode

justjake:
When `externalOutputRendering: "terminal-native"` in CliRendererConfig, commit bytes written to captured stdin with minimal transformation, instead of re-rendering them.

In this mode, we compute the layout effect of each chunk by splitting on `\n`, then using `strip-ansi` and `string-width` to get the width of the line before wrapping.

The outcome is that ANSI styling bytes written to stdout are rendered by the terminal itself without needing to implement ANSI->OpenTUI conversion.

Before (`bun run packages/examples/src/split-footer-streaming-demo.ts`):

CleanShot 2026-05-25 at 16 47 33@2x

After (`OTUI_SPLIT_STREAMING_DEMO_RENDERING=terminal-native bun run packages/examples/src/split-footer-streaming-demo.ts`):

CleanShot 2026-05-25 at 16 48 34@2x

### DIFF
--- a/packages/core/src/lib/external-output-byte-chunk.ts
+++ b/packages/core/src/lib/external-output-byte-chunk.ts
@@ -1,0 +1,66 @@
+import { stringWidth, stripANSI } from "../platform/runtime.js"
+
+export type ExternalOutputRendering = "emulated" | "terminal-native"
+
+export interface ByteChunkExternalOutputCommit {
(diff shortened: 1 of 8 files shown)


kommander:
@simonklee needs to check that when he's back

PR #1035, opened 2026-05-08
Native audio

m-pa:

PR #1114, opened 2026-05-27
Bug(linux): add musl native targets

riggy2013:
## Problem

On musl-based Linux systems (Alpine, CentOS 7.9 with musl LD_LIBRARY_PATH), opentui v0.2.9+ stalls at startup with no output. strace reveals a futex deadlock:
futex(..., FUTEX_WAIT, ... <unfinished ...>
This issue impacts opencode after 1.14.49 which integrates opentui 0.2.9.

## Root Cause

The native .so is compiled against glibc (x86_64-linux-gnu.2.17) and links libpthread. When dlopen loads this .so on a musl system, musl resolves pthread_* symbols from its own libc — but glibc and musl use different pthread_mutex_t internal layouts (size, alignment, field offsets). The Zig std.Thread.Mutex struct embedded in the audio Engine struct is laid out for glibc's pthread_mutex_t, causing futex operations to target the wrong memory address → deadlock.
This was introduced in v0.2.9 when Native audio (#1035) added the miniaudio C library + Zig audio wrapper which links libpthread.

## Fix

build.zig — Add musl compilation targets:
x86_64-linux-musl, aarch64-linux-musl
zig.ts — ABI-aware native package resolution: on Linux, try @opentui/core-linux-{arch}-musl first, fall back to glibc if unavailable.
package.json — Add musl optionalDependencies for npm publication.
build.ts — Extend build variants with abi field for musl package creation.

## Testing

Zig 0.15.2 cross-compiled from macOS → x86_64-linux-musl and aarch64-linux-musl
Verified .so loads successfully on CentOS 7.9 (kernel 3.10.x) with musl libc
Testing performed with opencode TUI at full load (all layers initialized)


packages/core/src/zig.ts
```diff
@@ -64,7 +64,18 @@ import type {
...
+    const muslPkg = `@opentui/core-linux-${process.arch}-musl`
+    try {
+      return await import(muslPkg)
+    } catch {
+      // fall through to glibc variant
+    }
+  }
+  return await import(`@opentui/core-${process.platform}-${process.arch}`)
+}
+const nativePackage = await resolveNativePackage()
```
kommander, line 121: Did you build the opencode standalone executable with a version of this? Afaik this cannot be statically analyzed by the bundler and thus the native lib is not bundled in the standalone executable so won't be available when distributed.

riggy2013: Yes, I have built the opencode musl with a patched opentui .so lib and it can work on centos 7.9 w/o issue.

I am new to this project but an opencode user. If you feel more tests needed, please tell me, or help to do the test. Basically opentui need a musl lib now.

PR #1121, opened 2026-05-28
feat: add FreeBSD x86_64 build target

mfwolffe:
Add FreeBSD x86_64 to the Zig cross-compilation targets.

- build.zig: add x86_64-freebsd to SUPPORTED_TARGETS, link pthread
- miniaudio_shim.c: enable OSS backend for FreeBSD

Built and tested on FreeBSD 15.0-RELEASE x86_64. The resulting
libopentui.so links against native libc.so.7/libthr.so.3/libm.so.5
and loads via bun:ffi dlopen().

### DIFF
--- a/.github/workflows/build-core.yml
+++ b/.github/workflows/build-core.yml
@@ -166,7 +166,7 @@ jobs:
             cd packages/core
             bun run build:lib
             bun run test:native
-            bun run test:js
+            bun run test:js:serial
 
   # Gate job for branch protection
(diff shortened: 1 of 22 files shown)


halcyon:
@kommander Any chance we could get this PR looked at? It looks like this PR is needed to get OpenCode running on FreeBSD - https://github.com/anomalyco/opencode/issues/28642. Thank you.

PR #1122, opened 2026-05-28
feat(box): add titleColor prop

MEDOYED:
## PR Description

This PR introduces improvements specifically for the `<box></box>` element and adds a new `titleColor` prop.

### What was changed
- Added support for the new `titleColor` prop.
- Updated the `<box>` component so users can independently customize:
  - the border color
  - the title color

### Why
Previously, implementing different colors for the border and title required additional wrapper code and manual styling overrides. This approach was inconvenient and looked more like a workaround than a clean API solution.

With the new `titleColor` prop, this customization is now supported directly by the component API, making the implementation cleaner and easier to use.

### Additional context
The attached screenshot demonstrates the current implementation approach that users need to write today in order to achieve this behavior manually.
image

After this PR, the same result can be achieved simply by passing the new `titleColor` prop on 33 line in root <box></box> element.


kommander:
1. Blocking: native tests do not compile. `drawBox` now expects `titleColor` at `packages/core/src/zig/buffer.zig:1863`, but test callsites still pass the old argument list at `packages/core/src/zig/tests/buffer_test.zig:1963` and `packages/core/src/zig/tests/buffer_test.zig:1997`. `bun run test:native` fails with `member function expected 14 argument(s), found 13`.
2. Blocking: native benchmarks do not compile for the same signature mismatch. Stale callsites include `packages/core/src/zig/bench/buffer-color-blending_bench.zig:104`, `packages/core/src/zig/bench/buffer-color-blending_bench.zig:160`, and multiple calls in `packages/core/src/zig/bench/buffer-draw-box_bench.zig` starting at `:84`. `zig build bench -Dbench-optimize=ReleaseFast -- --iterations=1 --filter=drawBox` fails.
3. High: `OptimizedBuffer.drawBox` is now a breaking public/direct API change and crashes when existing callers omit `titleColor`. The new required field is at `packages/core/src/buffer.ts:438-477`, and `rgbaPtr(titleColor)` crashes on `undefined` at `packages/core/src/zig.ts:131-132`. This already breaks the repo’s own TS benchmark, whose `DrawBoxOptions` lacks `titleColor` at `packages/core/src/benchmark/box-draw-benchmark.ts:66-81` and calls `buffer.drawBox(spec)` at `:308`. Reproduced with `bun src/benchmark/box-draw-benchmark.ts --scenario buffer_fill_only_panels --iterations 1 --warmup-iterations 0 --no-output`.
4. High: visible `titleColor` can be skipped entirely when border/background are transparent. `drawBox` returns early based only on `borderColor` and `backgroundColor` at `packages/core/src/zig/buffer.zig:1881`, before drawing titles at `packages/core/src/zig/buffer.zig:2043-2051`. I reproduced a box with `border: true`, transparent border/background, `title: "T"`, and red `titleColor`; the captured top row was blank and the title cell remained a space. This violates the purpose of independent title color.
5. Medium: the added test does not cover the feature behavior. `packages/core/src/renderables/Box.test.ts:207-220` only checks parsing, setter storage, and `requestRender`; it does not assert rendered cell foreground, FFI/native propagation, bottom title behavior, fallback-to-border behavior, transparent border behavior, or direct `OptimizedBuffer.drawBox` behavior. The missed failures above confirm coverage is insufficient.
6. Low: docs are not updated for a new user-facing prop. The Box docs property table omits `titleColor` at `packages/web/src/content/docs/components/box.mdx:167-178`, and `packages/react/README.md:460-494` still documents title/styling examples without it. `titleColor` has no Markdown documentation hits.

MEDOYED:
Hi! Thanks for the detailed review. I've addressed all the issues you mentioned:

- **Blocking (Tests & Benchmarks):** Added the missing `titleColor` argument to the `drawBox` calls in both native tests (`buffer_test.zig`) and benchmarks (`buffer-color-blending_bench.zig`, `buffer-draw-box_bench.zig`). They compile successfully now.
- **High (TS API):** Made `titleColor` optional in `DrawBoxOptions` and the TS API. It now falls back to `borderColor` (`options.titleColor ?? options.borderColor`) if not provided, preventing crashes for existing callers.
- **High (Zig renderer logic):** Fixed the early return bug in `buffer.zig` (`isFullyTransparent`). The renderer now correctly draws the title even if both the background and borders are transparent.
- **Medium (Test coverage):** Added a new test in `Box.test.ts` that actually checks the rendered buffer (`currentRenderBuffer`) to ensure the title foreground color is properly applied, including the transparent border edge case.
- **Low (Docs):** Added `titleColor` to the properties table in `box.mdx` and updated the code examples in the React `README.md`.
- **Bonus:** Also updated `packages/react/examples/box.tsx` to showcase the new `titleColor` prop and the transparent border behavior.

Everything builds and passes locally (`bun run build` & `bun run test:native`). Let me know if you need any further adjustments!

PR #1125, opened 2026-05-30
fix(core): don't duplicate bare URLs in concealed markdown

### LINKED ISSUE #1050: MarkdownRenderable: bare URLs duplicated as 'url (url)' in concealed mode
## Bug

When `conceal` is `true` (the default for `MarkdownRenderable`), bare URLs in markdown content render as `https://example.com (https://example.com)` — the URL text followed by the same URL in parentheses.

This happens because the `link` case in `renderInlineToken` unconditionally appends ` (href)` after the link label. For named links like `[click here](https://example.com)`, this renders correctly as `click here (https://example.com)`. But for bare/autolinked URLs, the link text _is_ the href, producing duplication.

## Location

`packages/core/src/renderables/Markdown.ts` — the `"link"` case in `renderInlineToken`:

```typescript
case "link": {
  const linkHref = { url: token.href }
  if (this._conceal) {
    for (const child of token.tokens) {
      this.renderInlineTokenWithStyle(child, chunks, "markup.link.label", linkHref)
    }
    // Always appends (href), even when link text === href
    chunks.push(this.createChunk(" (", "markup.link", linkHref))
    chunks.push(this.createChunk(token.href, "markup.link.url", linkHref))
    chunks.push(this.createChunk(")", "markup.link", linkHref))
  }
}
```

## Fix

Skip the `(href)` suffix when the link text equals the href:

```typescript
case "link": {
  const linkHref = { url: token.href }
  if (this._conceal) {
    const linkText = token.tokens.map(t => t.raw).join("")
    const isBareUrl = linkText === token.href
    for (const child of token.tokens) {
      this.renderInlineTokenWithStyle(child, chunks, "markup.link.label", linkHref)
    }
    if (!isBareUrl) {
      chunks.push(this.createChunk(" (", "markup.link", linkHref))
      chunks.push(this.createChunk(token.href, "markup.link.url", linkHref))
      chunks.push(this.createChunk(")", "markup.link", linkHref))
    }
  }
}
```

## Reproduction

Any markdown with a bare URL:

```markdown
Check https://github.com/anomalyco/opentui for details.
```

Renders as: `Check https://github.com/anomalyco/opentui (https://github.com/anomalyco/opentui) for details.`

Expected: `Check https://github.com/anomalyco/opentui for details.`

This also affects markdown tables where bare URLs are common — the duplication wastes column width and degrades readability.

## Affected version

@opentui/core 0.2.6 (and likely since 0.1.88 when `conceal` was introduced via #767)

## Related

- anomalyco/opencode#18394 — long URLs in tables split/unclickable (separate but related rendering issue)

ash543210:
## Problem

Fixes #1050.

In concealed mode (`conceal: true`, the default for `MarkdownRenderable`), the `link` case in `renderInlineToken` unconditionally appended ` (href)` after the link label. For **named** links like `[click here](https://example.com)` that's correct — it renders `click here (https://example.com)`. But for **bare/autolinked URLs** the label already *is* the href, so it rendered twice:

```
https://example.com (https://example.com)
```

This is most visible in **table cells**, where it wastes column width.

## Fix

Skip the parenthesized `(href)` suffix when the link's visible text equals the href (`token.text === token.href`, both required `string` fields on marked's `Link` token). Named links are unchanged.

## A note on the repro

The linked issue's reproduction uses a bare URL in a **paragraph**. In practice paragraphs are rendered via a syntax-highlighted `CodeRenderable` from the raw text and never reach the `link` rendering path, so they were never actually affected. The duplication only manifests in contexts that go through `renderInlineContent` (table cells, inline emphasis). The regression test added here is therefore a **table**, which is what genuinely exercises the bug.

## Tests

Added to `packages/core/src/renderables/__tests__/Markdown.test.ts`:

- `table with bare url is not duplicated` — the regression test (fails before this change, passes after).
- `bare url is not duplicated in concealed mode` — documents that paragraphs are unaffected.
- `named link still shows href in concealed mode` — guards the named-link behaviour.

The existing `table with links` test still passes, confirming named links continue to render `label (href)` with no regression.

```
150 pass, 0 fail
```

TypeScript-only change — no native build required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -457,12 +457,15 @@ export class MarkdownRenderable extends Renderable {
       case "link": {
         const linkHref = { url: token.href }
         if (this._conceal) {
+          const isBareUrl = token.text === token.href
           for (const child of token.tokens) {
             this.renderInlineTokenWithStyle(child as MarkedToken, chunks, "markup.link.label", linkHref)
           }
-          chunks.push(this.createChunk(" (", "markup.link", linkHref))
-          chunks.push(this.createChunk(token.href, "markup.link.url", linkHref))
-          chunks.push(this.createChunk(")", "markup.link", linkHref))
+          if (!isBareUrl) {
+            chunks.push(this.createChunk(" (", "markup.link", linkHref))
+            chunks.push(this.createChunk(token.href, "markup.link.url", linkHref))
+            chunks.push(this.createChunk(")", "markup.link", linkHref))
+          }
         } else {
           chunks.push(this.createChunk("[", "markup.link", linkHref))
           for (const child of token.tokens) {
--- a/packages/core/src/renderables/__tests__/Markdown.test.ts
+++ b/packages/core/src/renderables/__tests__/Markdown.test.ts
@@ -753,6 +753,39 @@ test("table with links", async () => {
   `)
 })
 
+test("bare url is not duplicated in concealed mode", async () => {
+  const markdown = "Visit https://example.com today."
+
+  expect(await renderMarkdown(markdown)).toMatchInlineSnapshot(`
+    "
+    Visit https://example.com today."
+  `)
+})
+
+test("named link still shows href in concealed mode", async () => {
+  const markdown = "[click here](https://example.com)"
+
+  expect(await renderMarkdown(markdown)).toMatchInlineSnapshot(`
+    "
+    click here (https://example.com)"
+  `)
+})
+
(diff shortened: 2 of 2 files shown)


ash543210:
Closing this — I realize after opening it that #1081 already proposed essentially the same fix for #1050 and was closed without merging. I don't want to re-litigate a change a maintainer has already declined. I'll ask on the issue itself about the preferred direction before reopening any work here. Apologies for the duplicate.

PR #1128, opened 2026-05-31
ref(grapheme): harden pool slot lifetime handling

simonklee:
Reject invalid and oversized graphemes before they can corrupt slab state or access unallocated storage. Keep allocation and reference transitions consistent on failure so stale IDs, double frees, and OOM paths cannot leave the pool in an invalid state.


simonklee:

PR #1129, opened 2026-05-31
renderer: prevent calls after destruction

simonklee:
Deferred teardown of a detached focused textarea can run after the renderer frees its native state. Ignore renderer operations once destruction starts so cleanup cannot call through a freed pointer and crash on shutdown


simonklee:
https://github.com/anomalyco/opentui/commit/c127f724a014f4b7cb7226a5f946224d39e6bd87

PR #1133, opened 2026-06-01
feat(core): double-click to select the word under the cursor

mhafaneh:
## Summary
Adds double-click word selection to the renderer mouse handling.

A left-button `down` within **500ms** at the same `(x, y)` cell as the previous
click is treated as a double-click. On a double-click, the selection expands to
the **contiguous run of printable cells** (codepoint > 32) around the clicked
cell, scanning left/right in the current render buffer to find word boundaries.

## Behavior
- Triggers only on `MouseButton.LEFT`, no `ctrl` modifier.
- Only fires on a `selectable`, non-destroyed renderable (consistent with the
  existing single-click selection gate).
- No-op when the clicked cell is whitespace/empty (codepoint <= 32).
- Falls through to the normal single-click selection path otherwise.

## Changes
- `packages/core/src/renderer.ts`
  - Add `_lastClickTime / _lastClickX / _lastClickY` tracking fields.
  - Add the double-click expansion block before the single-click selection
    handler in `handleMouseEvent`.

## Notes
- The 500ms threshold could be made configurable on `CliRendererConfig` if desired.

---
Discussion: #1138

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -819,6 +819,11 @@ export class CliRenderer extends EventEmitter implements RenderContext {
   private lastOverRenderableNum: number = 0
   private lastOverRenderable?: Renderable
 
+  // Double-click tracking for word selection (see handleMouseEvent).
+  private _lastClickTime: number = 0
+  private _lastClickX: number = -1
+  private _lastClickY: number = -1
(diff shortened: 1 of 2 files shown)


mhafaneh:

PR #1135, opened 2026-06-01
perf(core): throttle tree-sitter re-highlighting while streaming

mhafaneh:
## Summary
Adds an optional throttle for tree-sitter re-highlighting while a
`CodeRenderable` is streaming.

Today, with `streaming: true`, every content update re-runs a full tree-sitter
parse + highlight of the **entire** document. For long streamed responses this
is O(n^2) on the main thread and shows up as growing per-frame cost / jank as
the document grows.

This change coalesces those re-highlights to at most one per
`streamingHighlightThrottleMs` window (default **100ms**) once past the initial
chunk. Between highlights the buffer is still kept visually current:
- when `drawUnstyledText` is true, the raw text is pushed every frame so prose
  flows live;
- the trailing scheduled highlight then applies the styled result.

## Behavior
- New option `CodeOptions.streamingHighlightThrottleMs` (default `100`).
- `0` disables throttling (current behavior: re-highlight on every update).
- Only active while `streaming === true` and after the initial content.
- Pending timer is cancelled when streaming is toggled and on destroy.

## Changes
- `packages/core/src/renderables/Code.ts`
  - Add `streamingHighlightThrottleMs` option + default.
  - Add `_scheduleThrottledHighlight()` / `_cancelThrottledHighlight()`.
  - New throttled branch in `renderSelf`.
  - Override `destroySelf()` to clear any pending timer.

## Notes
- Backwards compatible; default preserves visible behavior while capping cost.

---
Discussion: #1139

### DIFF
--- a/packages/core/src/renderables/Code.streaming-throttle.test.ts
+++ b/packages/core/src/renderables/Code.streaming-throttle.test.ts
@@ -1,0 +1,151 @@
+import { test, expect, beforeEach, afterEach } from "bun:test"
+import { CodeRenderable } from "./Code.js"
+import { SyntaxStyle } from "../syntax-style.js"
+import { RGBA } from "../lib/RGBA.js"
+import { createTestRenderer, type TestRenderer, MockTreeSitterClient } from "../testing.js"
+import type { SimpleHighlight } from "../lib/tree-sitter/types.js"
+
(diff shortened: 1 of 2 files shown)


mhafaneh:

PR #1136, opened 2026-06-01
perf(core): avoid always-pending re-render in ScrollBox.recalculateBarProps

mhafaneh:
## Summary
`recalculateBarProps()` unconditionally scheduled a
`process.nextTick(() => this.requestRender())` at the end of every call. During
chat-style streaming with sticky-scroll-to-bottom and a continuously growing
child, this left a render permanently pending between content updates — the
renderer never settled to idle, burning CPU.

This change records the content/viewport dimensions seen at each call and only
schedules the deferred re-render when it can actually change output:
- **viewport dimensions changed** — preserves the resize workaround the existing
  comment describes; or
- **sticky scroll is active and content grew** — preserves auto-scroll-to-bottom
  during streaming.

The four `_lastBarRecalc*` fields initialize to `-1`, so the first call always
fires.

## Changes
- `packages/core/src/renderables/ScrollBox.ts`
  - Add `_lastBarRecalcScrollHeight/Width` and `_lastBarRecalcViewportHeight/Width`.
  - Gate the trailing `process.nextTick(requestRender)` on the conditions above.

## Notes
- Behavior-preserving for the cases the original `nextTick` was protecting
  (resize + sticky streaming); it only removes redundant always-pending renders.

### DIFF
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -138,6 +138,13 @@ export class ScrollBoxRenderable extends BoxRenderable {
   private _stickyScrollBottom: boolean = false
   private _stickyScrollLeft: boolean = false
   private _stickyScrollRight: boolean = false
+
+  // Snapshot of content/viewport size at the last recalculateBarProps() so the
+  // deferred re-render below can be skipped when nothing relevant changed.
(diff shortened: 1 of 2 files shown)


mhafaneh:

Issue #1139, opened 2026-06-01
Feature: opt-in throttle for tree-sitter re-highlighting while streaming


mhafaneh:
## Summary
Add an opt-in `CodeOptions.streamingHighlightThrottleMs` (default `0` = disabled) to coalesce tree-sitter re-highlights while `streaming: true`.

## Motivation
With `streaming: true`, every content update re-parses and re-highlights the entire document. For long streamed responses this is O(n^2) on the main thread and shows up as growing per-frame cost. Throttling caps re-highlights to at most one per window once past the initial chunk, while keeping raw text current each frame (when `drawUnstyledText`).

## Why default 0
Defaulting to disabled preserves current behavior exactly (no surprise change, no impact on existing streaming tests). Consumers opt in with e.g. `streamingHighlightThrottleMs: 100`.

## Open questions for discussion
- Is an opt-in option the right surface, or should throttling be on by default with a sensible value?
- Preferred default window if enabled by default?

Opening per CONTRIBUTING.md (feature suggestions discussed via issues). Draft implementation in #1135.

Issue #1137, opened 2026-06-01
Feature: opt out of synchronous terminal setup via setupTerminal config (tmux -CC)


mhafaneh:
## Summary
Proposing a `CliRendererConfig.setupTerminal?: boolean` option (default `true`). When `false`, `createCliRenderer` would skip its synchronous `await renderer.setupTerminal()` call so an outer layer that already owns the terminal can drive raw mode / alternate-screen setup itself.

## Motivation
Running OpenTUI inside a tmux `-CC` (control mode) host: tmux owns terminal setup, and a second synchronous setup from OpenTUI corrupts terminal state. The caller wants to construct the renderer and perform setup at the appropriate time.

## Proposed API
```ts
const renderer = await createCliRenderer({ setupTerminal: false })
// caller drives setup later / outer layer owns it
```
Backwards compatible: default `true` preserves current behavior.

## Notes
Opening per CONTRIBUTING.md (feature suggestions discussed via issues). Draft implementation in #1132. Happy to adjust the API shape (e.g. naming, or a callback hook instead of a boolean) based on maintainer preference.

PR #1140, opened 2026-06-01
fix(core): respect useKittyKeyboard: null to disable the Kitty protocol

mhafaneh:
## Summary
Fix `useKittyKeyboard: null` being ignored.

`const kittyConfig = config.useKittyKeyboard ?? {}` collapses a caller-supplied
`null` to `{}`, which has two consequences:

- `buildKittyKeyboardFlags({}) === 0b101` (Kitty **enabled**) instead of
  `buildKittyKeyboardFlags(null) === 0` (disabled) — so passing
  `useKittyKeyboard: null` does **not** disable the Kitty keyboard protocol.
- It defeats the downstream `const useKittyForParsing = kittyConfig !== null`
  gate, which can then never be `false`.

Switching to an explicit `=== undefined` check preserves a caller-supplied
`null` so Kitty is actually disabled.

## Changes
- `packages/core/src/renderer.ts`: `useKittyKeyboard ?? {}` →
  `useKittyKeyboard === undefined ? {} : useKittyKeyboard`.
- `packages/core/src/tests/renderer.kitty-config.test.ts`: new tests asserting
  `null` disables and `undefined` / `{}` enable Kitty (via
  `renderer.useKittyKeyboard`).

## Notes
- Bug fix, so opening directly for review per CONTRIBUTING.md.
- Split out of #1132 (which now only contains the unrelated `setupTerminal` feature).

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -1047,7 +1047,10 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     if (process.platform === "linux") config.useThread = false
     lib.setUseThread(rendererPtr, config.useThread)
 
-    const kittyConfig = config.useKittyKeyboard ?? {}
+    // Explicit `undefined` check (not `??`) so a caller-supplied `null` is
+    // preserved rather than collapsed to `{}`. `null` means "disable Kitty";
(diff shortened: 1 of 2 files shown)


mhafaneh:

PR #1141, opened 2026-06-01
fix(core): allow _hasManualScroll reset when reaching bottom during concurrent operations (#1088 follow-up)

malventano:
## Related Issues
- anomalyco/opencode#29992 (regression discovered after #1088)
- #1088 (original sticky scroll fix)

## Issue
After merging #1088 and releasing in v0.2.16, testing revealed a related issue where sticky auto-scroll fails to re-engage when users scroll to the absolute bottom of content and then new content arrives (e.g., tool call results).

Root cause: In `updateStickyState()`, the guard `!this._isApplyingStickyScroll` prevents `_hasManualScroll` from being reset when the user reaches bottom **if** any sticky operation happened to be running concurrently. This creates a race condition that #1088 didn't address.

## Fix
Remove the overly conservative `_isApplyingStickyScroll` guard from the bottom-reset logic in `updateStickyState()`. When the user physically reaches the bottom of scrollable content, we should always recognize that state regardless of timing with concurrent operations.

The actual position-setting logic lives in `applyStickyStart()` which already has its own try/finally protection, so removing this guard doesn't introduce race conditions. It just allows proper state recognition.

## Testing
Added regression test verifying `_hasManualScroll` resets correctly even when `_isApplyingStickyScroll` is true during concurrent operations. Pre-patch test fails, post-patch all 42 tests pass.

## Impact
Minimal risk. Single condition removal in one code path.

### DIFF
--- a/packages/core/src/renderables/ScrollBox.ts
+++ b/packages/core/src/renderables/ScrollBox.ts
@@ -225,7 +225,7 @@ export class ScrollBoxRenderable extends BoxRenderable {
     } else if (this.scrollTop >= maxScrollTop) {
       this._stickyScrollTop = false
       this._stickyScrollBottom = true
-      if (!this._isApplyingStickyScroll && this._stickyStart === "bottom") {
+      if (this._stickyStart === "bottom") {
         this._hasManualScroll = false
       }
     } else {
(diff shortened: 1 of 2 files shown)


kommander:
I think I indirectly fixed that here https://github.com/anomalyco/opentui/pull/1144

PR #1144, opened 2026-06-02
Fix sticky scroll re-engagement and document render hook limits

kommander:
- Centralizes ScrollBox manual-scroll tracking with `syncManualScrollState()`.
- Re-enables sticky bottom when users scroll back to the sticky edge via mouse wheel or keys.
- Prevents programmatic sticky/layout recalculations from accidentally changing manual-scroll state.
- Documents `renderBefore`/`renderAfter` as draw-only hooks that run after layout and culling.
- Notes that culled ScrollBox children do not run render hooks, so layout/state must not depend on them.

Additionally:
- Incorporates PR #1141’s sticky-bottom regression test.
- Extends the centralized manual-scroll sync to also clear manual mode when sticky sync is active but the viewport is already back at the sticky edge.
- Confirms this branch covers the #1141 follow-up case without reintroducing programmatic resize/recalculation regressions.


malventano:
This seems like a solid fix, but in testing here, it looks to run into the [same issue](https://github.com/anomalyco/opencode/issues/29992#issuecomment-4608485542) I saw with my fix in #1141 , which is that if the user scrolls down to the bottom after a new tool call block is opened, but before that block is expanded out (from the result of the call), the tool call block will expand past the bottom and continue to leave the current scroll location behind (even though it was supposed to be at the bottom). It's a bit tricky to repro, but worth keeping an eye on.

If I had to guess, I believe the edge detection somehow fails to trigger if scrolling down to a half-opened tool call block, or something along those lines.

malventano:
On further use, this PR seems to have regressed the behavior a bit. I've run into several cases where scrolling back to bottom and then entering a new prompt resulted in the pane remaining at its scrolled position (should have stuck at the bottom but didn't).

malventano:
I figured it out. My earlier build with #1144 did not include my prune fix PR https://github.com/anomalyco/opencode/pull/29126 , and my test session was >100 messages long, which triggered escapes from bottom even with #1144 applied. Hopefully that other PR gets merged at some point to better address this interaction.

PR #1149, opened 2026-06-05
opentui: support Node.js 26 alongside Bun

### LINKED ISSUE #2: Support Node/Deno (NAPI instead of bun:ffi)
Have you thought about using NAPI to make the project run both in Bun and Node? 

There is this package to create automatic NAPI bindings from Zig code https://github.com/cztomsik/napigen

simonklee:
Bun remains the primary runtime; Node.js 26 is now a validated
second target, requiring --experimental-ffi for native rendering.
The published packages ship runtime-conditioned exports, the core
test suite runs under both lanes through a repo-owned bun:test
compatibility module, and the keymap and Solid packages have
packed-consumer smoke tests under Node.

Runtime-plugin, Solid preload, and Bun-plugin subpaths stay Bun-only for now.

Fix #2


mathieudutour:
Thanks for adding support to node! But it seems like this hasn't fixed https://github.com/anomalyco/opentui/pull/511. Maybe I'm missing something?

PR #1175, opened 2026-06-14
Document SSH package

neriousy:
Summary:
- Add a web docs SSH overview for @opentui/ssh, covering setup, sessions, framework hand-offs, auth, middleware, host keys, and shutdown.
- Add SSH to the web docs navigation/index and skill routing metadata.
- Update root/core/package READMEs so package listings and docs links include the current SSH and website docs surfaces.

Validation:
- bun run --cwd packages/web validate:docs
- git diff --check


simonklee:
This is outdated after doc changes

PR #1180, opened 2026-06-16
feat: add allowedLinkSchemes to text renderables & others to prevent unnecessary OSC8 events

### LINKED ISSUE #864: Security: file:// hyperlinks execute files — need link click handler or reveal mode
## Problem

`<a href="file:///path/to/file">` hyperlinks rendered via OSC8 will **execute** the file when clicked if it's executable (e.g. `.sh`, binaries). This is a security concern for any TUI that renders user-controlled or LLM-generated file paths as clickable links.

The current OSC8 implementation (#735, PR #736) correctly emits hyperlinks, but there's no way to control what happens when a link is clicked — the terminal's default handler runs, which for `file://` URLs means "open with default app."

## Use case

In opencode, tool output (Read, Write, Edit, Glob, Grep, etc.) includes file paths. We want these to be visually distinct and clickable, but clicking should **reveal the file in the system file manager** (Finder/Explorer/Nautilus), not execute it.

- macOS: `open -R /path/to/file`
- Windows: `explorer /select,\path\to\file`
- Linux: `xdg-open` (parent directory) or `nautilus --select`

### Option A: Link click callback on the renderer

Allow the application to register a handler that intercepts link clicks before the default behavior:

```typescript
const renderer = createCliRenderer({
  onLinkClick(url: string) {
    if (url.startsWith("file://")) {
      const path = url.slice(7)
      exec(`open -R ${JSON.stringify(path)}`)
      return true // handled, suppress default
    }
    return false // fall through to default (open in browser)
  }
})
```

### Option B: `onMouseUp` / `onClick` on `<a>` elements

Once #650 (DOM-like capture/bubble events) lands with mouse dispatch, `<a>` elements could support event handlers that let the application override default behavior:

```tsx
<a href={url} onMouseUp={(e) => {
  e.preventDefault()
  revealInFinder(path)
}}>
  {filename}
</a>
```

### Option C: Custom URL scheme support

Allow applications to register scheme handlers in the renderer, so `file://` can be remapped without changing the `<a>` element API.

---

Option A or B would unblock safe file path linking in opencode. Happy to help with implementation if there's a preferred direction.

viralcodex:
Fixes: #864  (in a way)

hi @kommander / @simonklee , this is trying to solve this issue by adding a allow list for what OSC8 links can be opened or not from the terminal.
Since Opentui cannot stop the event, we need to manually remove the OSC8 characters from the link chunk to limit the behaviour.
My next step would be add a native side hit calculation with the link spans (ids) so that when the mouse click happens we can capture the event and use custom behaviour like the issue describes.

Would love to have to a discussion on this. Thanks


simonklee:
Thanks. I think we send the link URL that the application
provides. The terminal and operating system control link activation. A default
scheme restriction would disable valid file and custom links. Applications must
select safe URLs or handle mouse events directly. This behavior is outside the
scope of OpenTUI, as explained in #864.

PR #1181, opened 2026-06-16
fix: TUI leaves terminal in mouse-reporting/raw mode; keystrokes / mo…

gmedici:
…use events print as ^[[<... after exit or crash [#10610](https://github.com/anomalyco/opencode/issues/10610)

Also related to: https://github.com/anomalyco/opencode/issues/30133


bbartels:
Running into this issue as well, would be great to see this fixed :)

simonklee:
ffb3c96adfe41a49ecb6fe82648153721dcb8560

PR #1192, opened 2026-06-21
Select: add showSelectionIndicator option

leonid-shutov:
## Summary

Add a `showSelectionIndicator` boolean option to `SelectRenderable` that controls whether the `▶` glyph is rendered next to the currently selected item.

## API

```ts
const select = new SelectRenderable(renderer, {
  width: 40,
  height: 10,
  options,
  showSelectionIndicator: false, // hides ▶ and reclaims its 2-column gutter
})
```

**Default:** `true` (no behaviour change for existing consumers).

When set to `false`, the indicator glyph **and** its 2-column gutter are both removed, giving item text two extra columns of horizontal space.

## Screenshots

Image

Image

## Incidental cleanup

Two refactors that aren't part of the feature but were needed to keep the render loop coherent once the configurable indicator was introduced:

1. Removed const contentX = 0 / const contentY = 0. Always 0, added as-is to every coordinate (a no-op).
2. Renamed descX → textX. It was seeded as the description's x-position but then reused as the start column for the item name too.

### DIFF
--- a/packages/core/src/renderables/Select.test.ts
+++ b/packages/core/src/renderables/Select.test.ts
@@ -91,6 +91,7 @@ describe("SelectRenderable", () => {
       expect(select.showScrollIndicator).toBe(false)
       expect(select.showDescription).toBe(true)
       expect(select.wrapSelection).toBe(false)
+      expect(select.showSelectionIndicator).toBe(true)
     })
 
     test("should initialize with custom selected index", async () => {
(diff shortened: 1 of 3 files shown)


simonklee:
The main branch already includes this feature, its tests, and its
documentation. Commit 4a247a0c implements the feature and credits you as a
co-author. We are closing this pull request because the feature is already
available.

PR #1195, opened 2026-06-23
tests: add a failing test for streaming markdown in scrollbox

simonklee:
Add tests for streaming Markdown should synchronously display the current styled source while asynchronous highlighting is pending.

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -874,7 +874,13 @@ export class MarkdownRenderable extends Renderable {
     id: string,
   ): boolean {
     if ((token.type === "text" || token.type === "paragraph") && renderable instanceof CodeRenderable) {
-      this.applyMarkdownCodeRenderable(renderable, this.normalizeScrollbackMarkdownBlockRaw(token.raw), 0)
+      this.applyMarkdownCodeRenderable(
+        renderable,
+        this.normalizeScrollbackMarkdownBlockRaw(token.raw),
+        0,
+        undefined,
+        this.createInitialStyledText(token),
+      )
       return true
     }
 
--- a/packages/core/src/renderables/__tests__/Markdown.streaming.snapshot.test.ts
+++ b/packages/core/src/renderables/__tests__/Markdown.streaming.snapshot.test.ts
@@ -1,0 +1,85 @@
+import { afterEach, expect, test } from "bun:test"
+import { SyntaxStyle } from "../../syntax-style.js"
+import { createTestRenderer, MockTreeSitterClient, type TestRenderer } from "../../testing.js"
+import { BoxRenderable } from "../Box.js"
+import { CodeRenderable } from "../Code.js"
+import { MarkdownRenderable } from "../Markdown.js"
+import { ScrollBoxRenderable } from "../ScrollBox.js"
+
+let renderer: TestRenderer | undefined
+
+afterEach(() => {
+  renderer?.destroy()
+  renderer = undefined
+})
+
+const pendingHighlights = (markdown: MarkdownRenderable) => {
+  const children = [...markdown.getChildren()]
+  const pending: CodeRenderable[] = []
+  while (children.length > 0) {
+    const child = children.pop()!
+    if (child instanceof CodeRenderable && child.isHighlighting) pending.push(child)
+    children.push(...child.getChildren())
+  }
+  return pending
+}
+
+test("streaming ordered list keeps growing text visible while markdown highlighting is pending", async () => {
+  const setup = await createTestRenderer({ width: 100, height: 16 })
+  renderer = setup.renderer
+  const treeSitterClient = new MockTreeSitterClient()
+  treeSitterClient.setMockResult({ highlights: [] })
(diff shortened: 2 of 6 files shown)


simonklee:
# Deferred Lifecycle Investigations

## Tree-sitter one-shot requests can remain pending forever

Status: unresolved. Content liveness for streaming Markdown list prose is protected by synchronously rendering `initialStyledText`, but the underlying request lifecycle still needs hardening.

### Failure modes

- `TreeSitterClient.highlightOnce()` has no request-level timeout or cancellation.
- A worker-side `ONESHOT_HIGHLIGHT` exception is converted to a generic `ERROR` without the request `messageId`. The
  client emits that error but does not settle the matching promise.
- The worker abstraction observes `message` and `error`, but not a clean or otherwise silent worker exit.
- A latest request that neither resolves nor rejects leaves `CodeRenderable.highlightingDone` pending indefinitely.
- Generic `CodeRenderable` instances using `streaming: true` and `drawUnstyledText: false` retain the previous text
  buffer until highlighting completes. If the latest request hangs, their visible content remains stale indefinitely.

Relevant code:

- `src/lib/tree-sitter/client.ts`: `highlightOnce`, `messageCallbacks`, `handleWorkerFailure`, and the generic `ERROR` handler.
- `src/lib/tree-sitter/parser.worker.ts`: `handleOneShotHighlight` and the outer message-handler catch.
- `src/platform/worker.ts`: worker event abstraction.
- `src/renderables/Code.ts`: streaming content setter and `startHighlight`.

### Exact deterministic reproduction

1. Create a `CodeRenderable` with a previous highlighted value, `streaming: true`, a filetype, and `drawUnstyledText: false`.
2. Use a `TreeSitterClient` whose next `highlightOnce()` returns a promise that never settles.
3. Assign new content and render once so the new request starts.
4. Observe that `code.content` is new while `code.plainText` and the frame retain the previous value.
5. Additional renders do not recover because `_highlightsDirty` was consumed when the hanging request started.

For the list-prose manifestation, `src/renderables/__tests__/Markdown.streaming.snapshot.test.ts` deliberately leaves
replacement highlights pending. Removing the current-token `initialStyledText` argument from
`MarkdownRenderable.applyListChildRenderable()` makes its pending snapshot display stale prefixes indefinitely.

### Separate rejection control

Ordinary rejection is already handled correctly. Run:

```sh
bun test src/renderables/Code.test.ts --test-name-pattern "falls back to unstyled text when highlights fail"
```

The catch path writes current plain text, updates layout, and requests a render. The unresolved problem is a request
that never rejects or resolves.

### Candidate hardening

- Include `messageId` in one-shot worker errors and reject the matching callback.
- Observe worker exit and reject all pending requests.
- Add request cancellation or a bounded timeout policy.
- Track multiple outstanding highlights accurately rather than exposing only the latest promise and one shared boolean.

## Markdown style refresh can emit a blank frame

Status: unresolved. This was exposed by the original Solid streaming-list fixture but is independent of Solid,
scrollboxes, culling, and highlighting.

### Root cause

Changing `MarkdownRenderable.syntaxStyle` marks `_styleDirty`. During `renderSelf()`, `rerenderBlocks()` can force
`updateBlocks(true)` to destroy and recreate children. This happens after Yoga layout and render-command collection.
Commands for destroyed children are skipped, while newly created children are absent from the current render list,
producing a blank frame.

The original Solid test triggered this accidentally by writing `syntaxStyle={SyntaxStyle.fromTheme([])}` inline beside reactive `content={content()}`. Every content update created a new style identity. Hoisting one stable `SyntaxStyle` makes the list test exercise only content updates and pass.

### Exact deterministic reproduction

1. Create a top-level `MarkdownRenderable` with visible content and render it.
2. Assign `markdown.syntaxStyle = SyntaxStyle.fromTheme([])` with a new identity.
3. Optionally assign updated content in the same turn.
4. Render once and capture the frame: replaced Markdown children are absent.
5. Render again: the recreated children appear.

### Candidate hardening

- Move style reconciliation to a pre-layout phase.
- Apply compatible style updates in place without rebuilding the child tree.
- Do not mutate renderable structure from `renderSelf()`.

simonklee:
Little frank having a look at the life-cycle things that may cause this in application code: https://github.com/anomalyco/opentui/pull/1195#issuecomment-4776646926

simonklee:
I just focused on list and paragraph in pr so far. but blockquotes inside lists may instead cause the list row to be reconstructed because applyListChildRenderable() does not have an in-place blockquote branch, but their newly created text may still wait for highlighting because createBlockquoteRenderable() also lacks an initial styled fallback etc. so if we want to fix it following the direction of this pr i'll go over more edge cases

pisze-programy:
Confirmed the same bug and the fix resolves it. Thanks for working on this. 

I ran into exactly this issue: while the agent was streaming an ordered list lines were cut off and yet opencode.db contained the full text. Closing/reopening the TUI — or opening the Web UI — always showed the complete list, so this is purely a streaming-rendering problem, not a data problem

Environment I tested with:
- opencode 1.18.9 (Homebrew)
- @opentui/core 0.4.5 (bundled)
- macOS
- model: ollama/gemma4:latest (RTX 4090)

PR #1203, opened 2026-06-25
fix(core): support web-tree-sitter 0.26 wasm path

### LINKED ISSUE #1201: MarkdownRenderable silently renders raw markup (conceal stops working) with `web-tree-sitter@>=0.26` — hardcoded `web-tree-sitter/tree-sitter.wasm` was renamed upstream
## Summary

`@opentui/core`'s tree-sitter worker hardcodes an import of `web-tree-sitter/tree-sitter.wasm`. In `web-tree-sitter@0.26.0` that file was **renamed** to `web-tree-sitter.wasm` (and the package `exports` no longer expose `./tree-sitter.wasm`). When the resolved `web-tree-sitter` is `>=0.26`, the tree-sitter runtime fails to initialize, so the markdown grammar never loads and **`MarkdownRenderable` renders raw markup** (`##`, `**`, `` ` `` etc. are no longer concealed/styled). The failure is effectively silent — there is only a `console` warning; rendering itself doesn't throw.

## Environment

- `@opentui/core`: `0.4.1` (also reproduces on `0.4.2`; both `peerDependencies.web-tree-sitter = "0.25.10"`)
- `web-tree-sitter`: `0.26.3` (broken) vs `0.25.10` (works)
- Runtime: Node and Bun both affected (it's a module-resolution problem, not runtime-specific)

## Root cause

`parser.worker.js` resolves the web-tree-sitter runtime wasm by name:

```js
// @opentui/core/parser.worker.js
let treeWasm = await resolveBundledFilePath(
  () => import("web-tree-sitter/tree-sitter.wasm", { with: { type: "wasm" } }),
  () => import.meta.resolve("web-tree-sitter/tree-sitter.wasm"),
  import.meta.url,
);
```

`web-tree-sitter`'s `exports` map by version:

| version  | wasm subpath exported            |
| -------- | -------------------------------- |
| `0.25.x` | `./tree-sitter.wasm`             |
| `0.26.x` | `./web-tree-sitter.wasm` (renamed) |

So on `0.26.x`, both the dynamic `import(...)` and `import.meta.resolve("web-tree-sitter/tree-sitter.wasm")` fail with `Cannot find module 'web-tree-sitter/tree-sitter.wasm'`. The TreeSitter client init throws, the markdown grammar is never available, and conceal becomes a no-op (raw markup is shown).

Observed warning:

```
TreeSitter client initialization failed: ResolveMessage: Cannot find module 'web-tree-sitter/tree-sitter.wasm'
```

## Minimal reproduction

```ts
import { createCliRenderer, MarkdownRenderable, SyntaxStyle, resolveRenderLib } from "@opentui/core";

const lib = resolveRenderLib();

const md = new MarkdownRenderable(renderer, {
  content: "## Heading\n\n**bold** and `code`",
  syntaxStyle: new SyntaxStyle(lib, lib.createSyntaxStyle()),
  conceal: true,
});
renderer.root.add(md);

// web-tree-sitter@0.25.10  -> "Heading  bold and code"        (markers concealed) ✅
// web-tree-sitter@0.26.x   -> "## Heading  **bold** `code`"   (raw markup)        ❌
```

Toggle the result by installing `web-tree-sitter@0.25.10` vs `web-tree-sitter@0.26.3`.

## Expected vs actual

- **Expected:** markdown markers are concealed/styled regardless of which supported `web-tree-sitter` is installed.
- **Actual:** with `web-tree-sitter@>=0.26`, the grammar fails to load and markup renders raw.

## Why "just pin `web-tree-sitter@0.25.10`" isn't a great workaround

`@opentui/core` declaring the peer as `0.25.10` pushes the constraint onto every consumer. Apps that also depend on `web-tree-sitter@0.26+` directly (other tree-sitter grammars, newer API) end up with a hard version conflict, and resolvers (pnpm/npm) may satisfy `@opentui/core`'s peer with the consumer's `0.26.x`, silently re-triggering this bug. There's also no public way to override the runtime wasm path: the `dataPath` option only points at the **grammar** asset directory, not the `web-tree-sitter` runtime wasm.

## Suggested fixes (any one)

1. **Resolve both filenames** in the worker — try `web-tree-sitter/web-tree-sitter.wasm` (0.26+) first, fall back to `web-tree-sitter/tree-sitter.wasm` (≤0.25). Smallest change, supports both majors:
   ```js
   const treeWasm = await resolveBundledFilePath(
     () => import("web-tree-sitter/web-tree-sitter.wasm", { with: { type: "wasm" } })
            .catch(() => import("web-tree-sitter/tree-sitter.wasm", { with: { type: "wasm" } })),
     () => { try { return import.meta.resolve("web-tree-sitter/web-tree-sitter.wasm"); }
             catch { return import.meta.resolve("web-tree-sitter/tree-sitter.wasm"); } },
     import.meta.url,
   );
   ```
2. **Bump the peer range** to support `web-tree-sitter@^0.26` and use the new filename.
3. **Expose an override** for the runtime wasm path (e.g. extend `dataPath`/a `wasmPath` option) so consumers can point at the correct file.
4. **Surface the init failure** (throw or a prominent error) instead of silently degrading to raw markup — this took a while to diagnose precisely because rendering "works" but conceal silently stops.

snvtac:
## Summary

- Resolve `web-tree-sitter/web-tree-sitter.wasm` first for `web-tree-sitter@0.26+`.
- Fall back to `web-tree-sitter/tree-sitter.wasm` for `web-tree-sitter@0.25`.
- Add focused coverage for the preferred and fallback wasm path resolution.

Closes #1201.

## Testing

- `bun test src/lib/tree-sitter/parser.worker.test.ts src/lib/tree-sitter/client.test.ts`
- `bun test src/renderables/__tests__/Markdown.test.ts`
- `bun run fmt:check`
- `bun run lint`
- `bunx tsc -p packages/core/tsconfig.build.json --noEmit --pretty false`
- `bun test`
- Temporary `web-tree-sitter@0.26.3` install with `--no-save`: `bun test src/lib/tree-sitter/parser.worker.test.ts src/lib/tree-sitter/client.test.ts`

Not run: `bun run test:native` because `zig` is not installed in this environment.

### DIFF
--- a/packages/core/src/lib/tree-sitter/parser.worker.test.ts
+++ b/packages/core/src/lib/tree-sitter/parser.worker.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, test } from "bun:test"
+
+import { resolveWebTreeSitterWasmPath } from "./web-tree-sitter-wasm.js"
+
+describe("parser worker web-tree-sitter wasm resolution", () => {
+  test("prefers the web-tree-sitter 0.26 wasm filename", () => {
+    const resolvedSpecifiers: string[] = []
+
+    const resolved = resolveWebTreeSitterWasmPath((specifier) => {
+      resolvedSpecifiers.push(specifier)
+      return `/node_modules/${specifier}`
+    })
+
+    expect(resolved).toBe("/node_modules/web-tree-sitter/web-tree-sitter.wasm")
+    expect(resolvedSpecifiers).toEqual(["web-tree-sitter/web-tree-sitter.wasm"])
+  })
+
+  test("falls back to the web-tree-sitter 0.25 wasm filename", () => {
+    const resolvedSpecifiers: string[] = []
+
+    const resolved = resolveWebTreeSitterWasmPath((specifier) => {
+      resolvedSpecifiers.push(specifier)
+      if (specifier === "web-tree-sitter/web-tree-sitter.wasm") {
+        throw new Error("new wasm filename is not exported")
+      }
+
+      return `/node_modules/${specifier}`
+    })
+
+    expect(resolved).toBe("/node_modules/web-tree-sitter/tree-sitter.wasm")
+    expect(resolvedSpecifiers).toEqual(["web-tree-sitter/web-tree-sitter.wasm", "web-tree-sitter/tree-sitter.wasm"])
+  })
+})
--- a/packages/core/src/lib/tree-sitter/parser.worker.ts
+++ b/packages/core/src/lib/tree-sitter/parser.worker.ts
@@ -15,6 +15,7 @@ import type {
 } from "./types.js"
 import { DownloadUtils } from "./download-utils.js"
 import { isBunfsPath, normalizeBunfsPath } from "../bunfs.js"
+import { importWebTreeSitterWasm, resolveWebTreeSitterWasmPath } from "./web-tree-sitter-wasm.js"
 import { resolveBundledFilePath } from "../../platform/runtime.js"
 import {
   isWorkerRuntime,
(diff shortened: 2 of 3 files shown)


denghongcai:
Should update peerDependencies either?

Issue #975, opened 2026-04-26
0.1.103: DSR 997 (color-scheme updates) creates an infinite OSC 10/11 query loop under tmux on macOS


lavaman131:
## Summary

Upgrading from 0.1.102 → 0.1.103 causes any `createCliRenderer`-based app running inside a tmux pane on macOS to enter an infinite escape-sequence loop the moment the renderer initializes. The terminal fills with repeating bytes; keyboard input never reaches the app; only killing the terminal or `tmux kill-server` recovers.

The regression is in #969 ("more reliable theme mode"). Pinning to `0.1.102` fully resolves the symptom.

## Repro

- macOS (Terminal.app / iTerm2 — both reproduce)
- tmux 3.6a
- `@opentui/core` and `@opentui/react` at `0.1.103`

```ts
import { createCliRenderer } from "@opentui/core";

await createCliRenderer({
  exitOnCtrlC: false,
});
// keep the process alive — any live OpenTUI renderer reproduces it
await new Promise(() => {});
```

Run that script inside a tmux session (`tmux new -s repro 'bun script.ts'`, then `tmux attach -t repro`). The pane immediately starts emitting a continuous stream of escape sequences and never settles.

Outside tmux: fine. With `0.1.102` inside tmux: also fine.

## Root cause

In `packages/core/src/zig/terminal.zig`, `setupTerminal` calls:

```zig
try self.setColorSchemeUpdates(tty, true);   // CSI ? 2031 h
```

unconditionally, opting the terminal into DSR 997 (`CSI ? 997 ; {1|2} n`) theme-change notifications.

Then `packages/core/src/renderer-theme-mode.ts:handleSequence` reacts to every DSR 997 by re-firing OSC 10/11 queries:

```ts
if (sequence === "\x1b[?997;1n" || sequence === "\x1b[?997;2n") {
  this.requestThemeOscColors();   // host.queryThemeColors()
  return { handled: true, changedMode: null };
}
```

Under tmux on macOS, the OSC 10/11 reply round-trip (or the alt-screen flip / pane redraw) is enough to cause the terminal to emit another DSR 997, which re-arms another OSC 10/11 query, and so on. The existing tmux guard

```zig
self.theme_queries_pending = !self.in_tmux;
```

only suppresses the *initial* startup query — it doesn't prevent subsequent terminal-initiated DSR 997s from re-triggering the loop.

## Suggested fixes (any one of these would close the loop)

[...]

Issue #513, opened 2026-01-11
Automatically adapt appearance (light/dark) to that of the system


ccjmne:
Hey folks,
I'd like to have OpenCode look bright during the day and dark at night.

I already played around with similar features in Ghostty, tmux, Neovim and I monkey-patched the code base tonight, I can attest to its fairly straightforward, and even simple feasibility.

However, there's still a few other topics I haven't begun considering:

- desirability: **would you agree it'd be a nice thing to have?**
- configuration: I can make "the thing", but I haven't even looked at or begun thinking of how we'd let the user toggle that behaviour on and off.
- testability: I have in particular no idea about the React and Solid bits, nor about testing in this project in general nor for that feature in particular. I'd have to enlist the LLM for support on that front, but I'd do it responsibly (if I can't understand precisely what it does and subsequently own it, I'd rather offer nothing and ask for help). I can however already commit to copious manual testing ;)

**My idea**

My plan would be to emit a `CSI` sequence `? 2031 h` when this "automatic appearance toggle" mechanism would be turned on, and listed for the `DECSET`, `DECRST` status reports through `997`. It's what modern terminal emulators, multiplexers and TUIs use.

A short and clear overview of the principles that go into implementing such support (as well as a table outlining the adoption state across the most prominent terminal tools) is available at https://contour-terminal.org/vt-extensions/color-palette-update-notifications.

**TL;DR:**

I implemented some proof of concept on my end, which works splendidly; now I'm thinking of contributing upstream and figured, from the first line of the `CONTRIBUTING` guidelines, that I should start a conversation here before anything.

So here it goes: what do you think? If we want it, I'll get started with making it work and making it configurable.

kommander:
`2031` is general color scheme updates, right? In `terminal.zig` it already tries to detect that capability. I would activate/deactivate it there in `enableDetectedFeatures` and `resetState` respectively. In `renderer.ts` I would then add an input handler when it is supported, which can emit a `color-sheme-changed` event or something similar. It can then fetch bg/fg terminal palette colors and check if they flipped, to additionally emit and dark/light mode event. 

If you have a better idea, go for it. Happy to bring this in.

ccjmne:
Hey @kommander,
I am dealing with some personal emergency and haven't spent any more time looking into this, but I still firmly intend to put something together in some near future.

[...]

PR #1213, opened 2026-07-01
fix(core): honor late OSC replies + bound re-query on CSI ?997 (runtime theme switching on non-2031 terminals)

DEAN-Cherry:
## Problem

Runtime light/dark theme switching fails intermittently on terminals that do **not** support DEC mode 2031 (notably **WezTerm**), even when those terminals inject `CSI ?997;{1,2}n` themselves to signal the change. opentui receives the notification and re-queries OSC 10/11, but the mode often does not flip — the reply is discarded and the mode stays on the old value until the next launch.

This surfaces downstream as:
- opencode [#23196](https://github.com/anomalyco/opencode/issues/23196) — "Themes always render light variant … regardless of terminal dark background (OSC 11 detection broken)"
- opencode [#21870](https://github.com/anomalyco/opencode/issues/21870) — "Theme detection fails to use dark variant despite OSC 11 returning dark background color"
- opencode [#19254](https://github.com/anomalyco/opencode/issues/19254) — auto theme mode not working in multiplexer popups
- opentui [#513](https://github.com/anomalyco/opentui/issues/513) — auto-adapt appearance (the upstream feature this unblocks on non-2031 terminals)

## Root cause

Two issues in `packages/core/src/renderer-theme-mode.ts`:

**1. Late OSC 10/11 replies are silently dropped.**
`requestThemeOscColors()` arms a `QUERY_TIMEOUT_MS = 250ms` window. If the terminal answers slower than 250ms, `completeThemeQuery()` sets `themeQueryPending = false`, and `handleSequence` then returns `{ changedMode: null }` for the delayed reply — the inferred mode is never applied. WezTerm's OSC 11 reply can lag ~1s behind a `color_scheme` change (the terminal needs to apply the new scheme before its OSC 11 answer reflects the new background), so the answer regularly arrives after the timeout and is thrown away.

**2. `CSI ?997` received while a refresh is in flight is dropped with no follow-up.**
The debounce guard added in f0733f58 (#975) — `if (themeRefreshTimeoutId !== null) return` — correctly prevents the tmux feedback loop from recurring, but as a side effect it drops every notification that lands inside the 250ms window. Terminals that retry `?997` within that window (e.g. WezTerm configs that re-inject at 0.3/0.8/1.5/2.5s to paper over the OSC-11 settle delay) lose the notification entirely; the next switch has to wait for a fresh `?997` from the terminal.

## Fix

**Late replies (bug 1):** `handleSequence` now applies the inferred mode once both `themeOscForeground` and `themeOscBackground` have arrived, regardless of `themeQueryPending`. This is safe because:
- `requestThemeOscColors` resets fg/bg to `null` at the start of each query, so stale values from a prior query cannot leak in.
- `applyThemeMode`'s `if (!changed) return null` short-circuits redundant emits when the reply confirms the current mode.
- No new `queryThemeColors()` is issued, so this cannot create a feedback loop.

**Bounded re-query (bug 2):** `requestThemeOscColors` records a single-slot `requeryPending` flag instead of dropping the notification, and the 250ms timeout callback starts exactly one follow-up query when that flag is set. The bound is a boolean, not a counter — worst case is one extra query per notification burst, never an unbounded loop — so #975 cannot recur.

## Tests

Two new cases in `packages/core/src/tests/renderer.input.test.ts`:

- `late OSC 11 reply arriving after the query timeout still applies the theme mode` — drives the 250ms timeout to fire with no reply, then delivers the reply and asserts the mode flips. Fails on `main`, passes with the fix.
- `CSI 997 received during an in-flight refresh triggers a bounded re-query` — sends a second `?997` inside the in-flight window and asserts exactly one follow-up query runs after the window ends; also verifies the bound holds (no further query spawns when no additional notification arrives). Fails on `main`, passes with the fix.

### Verification

```
# new tests
bun test packages/core/src/tests/renderer.input.test.ts \
  --test-name-pattern "late OSC 11 reply arriving after the query timeout|CSI 997 received during an in-flight refresh"
# 2 pass

# full theme/997/wait suite (13 pre-existing + 2 new)
bun test packages/core/src/tests/renderer.input.test.ts --test-name-pattern "997|theme|themeMode|wait"
# 15 pass, 0 fail

# entire renderer.input.test.ts
bun test packages/core/src/tests/renderer.input.test.ts
# 98 pass, 0 fail

# terminal-capability-detection (untouched, sanity)
bun test packages/core/src/lib/terminal-capability-detection.test.ts
# 24 pass, 0 fail

# lint
oxlint packages/core/src/renderer-theme-mode.ts packages/core/src/tests/renderer.input.test.ts
# 0 warnings, 0 errors
```

The pre-existing tmux debounce test (`CSI 997 does not query theme colors again while a refresh is already pending`) still passes — the first in-flight notification still does not trigger an immediate re-query; only the bounded follow-up after the window ends is new.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

## Notes for reviewers

- The `themeQueryPending` field is kept (still used by `cancelRefresh`/`dispose` semantics and the in-flight guard); only the `if (!themeQueryPending) return` drop in `handleSequence` is removed.
- The debounce guard from #975 is preserved verbatim in spirit — an in-flight notification never starts an immediate query. The only addition is the single-slot `requeryPending` that turns the silent drop into a bounded follow-up.
- This does **not** add DEC mode 2031 support to opentui (#513 remains the tracking issue for that); it makes the existing OSC 10/11 fallback path reliable at runtime on terminals that already inject `?997` (or where users inject it themselves).

### DIFF
--- a/packages/core/src/renderer-theme-mode.ts
+++ b/packages/core/src/renderer-theme-mode.ts
@@ -44,7 +44,6 @@ export class RendererThemeMode {
   private static readonly QUERY_TIMEOUT_MS = 250
 
   private _themeMode: ThemeMode | null = null
-  private themeQueryPending = true
   private themeOscForeground: string | null = null
   private themeOscBackground: string | null = null
   private themeRefreshTimeoutId: TimerHandle | null = null
@@ -89,7 +88,6 @@ export class RendererThemeMode {
 
     this.clock.clearTimeout(this.themeRefreshTimeoutId)
     this.themeRefreshTimeoutId = null
-    this.themeQueryPending = false
   }
 
   public dispose(): void {
@@ -130,10 +128,19 @@ export class RendererThemeMode {
       return { handled: false, changedMode: null }
     }
 
-    if (!this.themeQueryPending) {
-      return { handled: true, changedMode: null }
-    }
-
+    // Apply the inferred mode once both foreground and background replies have
+    // arrived, even if the 250ms query timeout already fired. A terminal that
+    // answers OSC 10/11 slower than QUERY_TIMEOUT_MS — common under WezTerm,
+    // where a color-scheme change can take ~1s to settle in the OSC 11 reply —
+    // would otherwise have its late answer silently dropped, leaving the mode
+    // stuck on the old value. requestThemeOscColors resets fg/bg to null at
+    // the start of each query, so stale values from a prior query cannot leak
+    // in here.
+    //
+    // Note this means unsolicited OSC 10/11 replies can mutate the theme mode
+    // at any time (fg/bg persist after a completed query). After suspend(),
+    // inertness relies on the stdin data listener being detached — not on an
+    // internal pending flag.
     if (!this.themeOscForeground || !this.themeOscBackground) {
       return { handled: true, changedMode: null }
     }
@@ -156,23 +163,22 @@ export class RendererThemeMode {
 
   private completeThemeQuery(): void {
     this.clearThemeRefreshTimeout()
-    this.themeQueryPending = false
   }
 
   private requestThemeOscColors(): void {
     // Ignore repeated ?997 notifications while the current OSC refresh is
-    // still in flight.
+    // still in flight: under tmux on macOS the OSC 10/11 round-trip itself
+    // provokes another ?997, so starting a new query per notification would
+    // recreate the feedback loop from #975.
     if (this.themeRefreshTimeoutId !== null) {
       return
     }
 
-    this.themeQueryPending = true
     this.themeOscForeground = null
     this.themeOscBackground = null
 
     this.host.queryThemeColors()
 
-    this.clearThemeRefreshTimeout()
     this.themeRefreshTimeoutId = this.clock.setTimeout(() => {
       this.completeThemeQuery()
     }, RendererThemeMode.QUERY_TIMEOUT_MS)
--- a/packages/core/src/tests/renderer.input.test.ts
+++ b/packages/core/src/tests/renderer.input.test.ts
@@ -2019,6 +2019,41 @@ test("waitForThemeMode resolves with current theme mode when renderer is destroy
   await expect(themeModePromise).resolves.toBeNull()
 })
 
+test("late OSC 11 reply arriving after the query timeout still applies the theme mode", async () => {
+  const { renderer, queryThemeColorsCalls, clock } = await createThemeQueryRenderer()
+  const themeModes: string[] = []
+  renderer.on("theme_mode", (mode) => {
+    themeModes.push(mode)
+  })
+
+  // Establish an initial dark mode.
+  renderer.stdin.emit("data", Buffer.from("\x1b]10;#ffffff\x07"))
+  renderer.stdin.emit("data", Buffer.from("\x1b]11;#000000\x07"))
+  advanceClock(clock)
+  expect(renderer.themeMode).toBe("dark")
+
+  // Terminal reports a theme change. opentui re-queries OSC 10/11.
+  renderer.stdin.emit("data", Buffer.from("\x1b[?997;2n"))
+  advanceClock(clock, 249)
+  expect(queryThemeColorsCalls.count).toBe(1)
+
+  // Let the 250ms query timeout fire with no reply yet. Waiters stop waiting,
+  // but a delayed reply must still be honored rather than silently dropped.
+  advanceClock(clock, 1)
+  await flushMicrotasks()
+  expect(renderer.themeMode).toBe("dark")
+
+  // The terminal finally answers with a light background after the timeout.
+  renderer.stdin.emit("data", Buffer.from("\x1b]10;#000000\x07"))
+  renderer.stdin.emit("data", Buffer.from("\x1b]11;#ffffff\x07"))
+  advanceClock(clock)
+
+  expect(renderer.themeMode).toBe("light")
+  expect(themeModes).toEqual(["dark", "light"])
+
+  renderer.destroy()
+})
+
 test("pixel resolution response should not trigger keypress", async () => {
   const keypresses: KeyEvent[] = []
   currentRenderer.keyInput.on("keypress", (event) => {


kommander:
I verified this PR before review: both bugs are real, and both new tests genuinely fail on main and pass here. But the re-query mechanism reintroduces the #975 feedback loop it claims to prevent, so this can't merge as-is.

## Blocker: the "bounded" re-query is only bounded when notifications are external — in the #975 environment they aren't

#975's documented root cause: under tmux on macOS, the OSC 10/11 round-trip **itself provokes the next `?997`**. Mode 2031 is still enabled unconditionally (`terminal.zig:294`, no `in_tmux` gate), so the in-flight drop from #976 is the *only* thing breaking the cycle. This PR turns that drop into a deferred re-query, which re-closes it:

query → provoked `?997` lands in the 250ms window → `requeryPending = true` → timeout → re-query → provoked `?997` → …

The boolean bounds queries per notification, but each re-query *causes* the next notification, so the loop is self-sustaining at one query per 250ms. Simulated with the existing `ManualClock` harness (each query provokes a `?997` 50ms later, mirroring #975): **main settles at 1 query; this branch runs 41 queries in 10s of virtual time and never stops.**

The pre-existing #976 test still passing is misleading: `advanceClock` defaults to 20ms, so it never crosses the 250ms boundary. Its guarantee ("in-flight `?997` never causes another query") is hollowed out, not preserved.

<details>
<summary>Loop simulation test (fails on this branch: 41 queries, passes on main: 1 query)</summary>

```ts
import { test, expect } from "bun:test"
import { Buffer } from "node:buffer"
import { createTestRenderer } from "../testing/test-renderer.js"
import { createTerminalCapabilities } from "../testing/terminal-capabilities.js"
import { ManualClock } from "../testing/manual-clock.js"

// #975: "the OSC 10/11 reply round-trip ... is enough to cause the terminal
// to emit another DSR 997, which re-arms another OSC 10/11 query, and so on."
test("tmux #975 simulation: query-provoked ?997 must not sustain a query loop", async () => {
  const clock = new ManualClock()
  const stdout = { isTTY: true, columns: 80, rows: 24, write: () => true, getColorDepth: () => 24 } as any
  const { renderer } = await createTestRenderer({ clock, stdout, useThread: false })

  // @ts-expect-error - mocking for test
  renderer.lib.processCapabilityResponse = () => {}
  // @ts-expect-error - mocking for test
  renderer.lib.getTerminalCapabilities = () => createTerminalCapabilities()

  let queryCount = 0
  // @ts-expect-error - mocking for test
  renderer.lib.queryThemeColors = () => {
    queryCount += 1
    clock.setTimeout(() => renderer.stdin.emit("data", Buffer.from("\x1b[?997;1n")), 50)
  }

  renderer.stdin.emit("data", Buffer.from("\x1b[?997;1n"))

  for (let t = 0; t < 10_000; t += 10) {
    clock.advance(10)
    await Promise.resolve()
  }

  expect(queryCount).toBeLessThanOrEqual(2)
  renderer.destroy()
})
```

</details>

## The re-query also misses its own target case

The only consumer of `requeryPending` is the timeout callback — but `completeThemeQuery()` called from `handleSequence` (full reply arrives *before* the timeout) cancels that timeout, stranding the flag. On any terminal that answers within 250ms: `?997` #2 arrives in-flight, the terminal answers quickly with pre-settle colors → mode sticks on the stale value and the promised follow-up never runs. That is exactly the settle-lag race this PR describes, and it's still dropped. Verified with a second simulation (query count stays 1 over 5s; mode stuck).

## Minor

- After removing its only read, `themeQueryPending` is write-only. The reviewer note ("still used by cancelRefresh/dispose semantics and the in-flight guard") is incorrect — the in-flight guard checks `themeRefreshTimeoutId`, and `cancelRefresh`/`dispose` only write the field. Keep the gate or delete the field.
- With the gate gone, unsolicited OSC replies mutate theme mode at any time — including a lone OSC 11, since fg/bg persist after a completed query (they're only nulled at query start). Probably acceptable, but it means `cancelRefresh()`'s post-suspend inertness (renderer.ts:4014/4171) now relies solely on the stdin listener being detached. Worth stating explicitly.

## Path forward

1. **Split out fix 1 (honor late replies) — it can merge on its own.** It's sound (fg/bg reset at query start prevents stale leaks; `applyThemeMode` dedupes) and addresses the WezTerm symptom by itself: the late reply gets applied, and injections at 0.8/1.5/2.5s land *outside* the 250ms window, so they already start fresh queries on main.
2. **Fix 2 needs a real loop damper before any re-query can replace the #976 drop:**
   - a cooldown — at most one follow-up query per N seconds regardless of how many notifications arrive, and/or
   - gating 2031/`?997` handling under tmux at the Zig layer (options 1/3 from #975, never implemented), and
   - consuming `requeryPending` on early completion, not only in the timeout callback.
3. Any rework should include a regression test that models **query-provoked** notifications (the #975 shape), not just externally-timed ones. Happy to share both simulation tests.

DEAN-Cherry:
Thank you for the thorough review — the loop simulation made the flaw unambiguous, and you were right on all counts: the "bounded" re-query was only bounded for externally-timed notifications, and the flag was stranded on early completion.

Done as suggested:

1. **This PR is now fix 1 only** (honor late OSC replies), rebased onto main. Since removing the guard left `themeQueryPending` write-only — as you noted — the field is deleted entirely, and the comment now states explicitly that unsolicited OSC 10/11 replies can mutate the mode and that post-suspend inertness relies on the stdin listener being detached.
2. **Fix 2 moved to #1228** (stacked on this one), reworked with a real damper:
   - `requeryPending` is consumed on **both** completion paths (250ms timeout and early completion), fixing the stranded-flag case;
   - a **5s cooldown** allows at most one follow-up query regardless of notification count;
   - an idle `?997` within **1s** of the last query activity is treated as a follow-up candidate under the same cooldown — without this, a terminal that answers inside the 250ms window completes the query before its provoked `?997` lands, and the loop reopens through the "fresh query" path (your simulation with fast replies catches exactly this).
3. I adapted your tmux #975 simulation as a regression test (queries ≤ 2 over 10s of virtual time), plus the fast-reply variant and a settle-lag early-consumption test. Thanks for sharing it — it's the anchor test for the damper.

Happy to adjust the cooldown/window constants if you have preferences.

PR #1215, opened 2026-07-01
clipboard: add tmux transport mode

simonklee:
Let users choose how OpenTUI sends OSC 52 through tmux. Setups that disable passthrough can use tmux set-clipboard, and locked-down sessions can disable clipboard output instead of emitting unsupported escapes.

### DIFF
--- a/packages/core/src/zig/terminal.zig
+++ b/packages/core/src/zig/terminal.zig
@@ -74,6 +74,12 @@ pub const Osc52Support = enum(u8) {
     unsupported,
 };
 
+const TmuxClipboardMode = enum {
+    passthrough,
+    raw,
+    off,
+};
+
 const NOTIFICATION_QUERY_ID = "opentui-notifications";
 pub const SCREEN_PASSTHROUGH_CHUNK_SIZE = 252;
 pub const CLIPBOARD_PAYLOAD_SIZE_MAX = std.math.maxInt(u32);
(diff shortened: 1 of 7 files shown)


packages/core/src/zig/terminal.zig
```diff
@@ -1694,11 +1711,28 @@ fn writeClipboardBytes(writer: anytype, bytes: []const u8, escape: bool) !void {
...
+    };
+    defer if (env_map_storage) |*map| map.deinit();
+
+    if (env_map) |map| {
+        if (map.get("OPENTUI_TMUX_CLIPBOARD_MODE")) |value| {
+            if (std.mem.eql(u8, value, "raw")) return .raw;
+            if (std.mem.eql(u8, value, "off")) return .off;
+        }
+    }
+    return .passthrough;
```
ayubun, line 1728: i feel as though defaulting to passthrough is a bit of a crime, given the tmux maintainer said it is not required: https://github.com/tmux/tmux/issues/4214#issuecomment-2439605818
and tmux's own wiki documents using `set-clipboard on`: https://github.com/tmux/tmux/wiki/Clipboard

i would prefer this config to default to `raw`

ayubun (changes requested):
i still believe just sending the regular osc52 sequence as i proposed resolves this: https://github.com/anomalyco/opencode/pull/30472

but if you insist on making this a config, i'd settle for `raw` being the default

Issue #1184, opened 2026-06-18
`modifyOtherKeys` mode should be 2 instead of 1


abhinavnatarajan:
During terminal setup, `modifyOtherKeys` is the fallback when the terminal supports it and the kitty keyboard protocol is not supported:

https://github.com/anomalyco/opentui/blob/71b129abdc0854ba5153486b5f29356488223006/packages/core/src/zig/terminal.zig#L362-L381

Opentui requests `modifyOtherKeys` mode 1:

https://github.com/anomalyco/opentui/blob/71b129abdc0854ba5153486b5f29356488223006/packages/core/src/zig/ansi.zig#L394

This should be changed to mode 2, because mode 1 does not correctly disambiguate ordinary keys with ctrl+shift modifiers.
For example `ctrl+g` and `ctrl+shift+g` are the same in mode 1. This can be tested for example with `printf '\033[>4;1m'` followed by typing `ctrl+v`+`ctrl+g` or `ctrl+v`+`ctrl+shift+g`. 
This is pertinent to tmux users, since tmux does not yet support the kitty keyboard protocol. For example, with opencode running in tmux, several of the default keybindings like `ctrl+shift+a` for `input_select_line_home` do not work.

To check that `tmux` is running under mode 1, when running open code one can execute the tmux command `display -p '#{pane_key_mode}'` with the result `Ext 1`. By comparison, Neovim (for example) runs in `Ext 2` inside tmux.

The fix is very simple: simply change to

```zig
pub const modifyOtherKeysSet = "\x1b[>4;2m";
```

avncharlie:
Hi, I can confirm this issue as well. Setting modifyOtherKeys to 2 allows Ctrl-letter keybindings to work inside tmux for opentui apps. 
I hit this attempting to bind `ctrl+j` / `ctrl+k` in opencode's TUI, which won't work without setting modifyOtherKeys to 2.

PR #1220, opened 2026-07-02
fix: request modifyOtherKeys level 2 to support more key combos under tmux.

avncharlie:
This is fix for the issue described here: https://github.com/anomalyco/opentui/issues/1184

Tmux does not implement the kitty keyboard protocol, so opentui falls back to modifyOtherKeys. Level 1 (CSI > 4 ; 1 m) does not encode standard control characters like Ctrl-J (0x0a), so tmux delivers a raw byte indistinguishable from a newline. Level 2 (CSI > 4 ; 2 m) encodes these as CSI-u sequences (e.g. \x1b[106;5u), which the parser already decodes correctly — the fix is solely in what level we request.

### DIFF
--- a/packages/core/src/lib/parse.keypress-kitty.test.ts
+++ b/packages/core/src/lib/parse.keypress-kitty.test.ts
@@ -38,6 +38,27 @@ test("parseKeypress - Kitty keyboard ctrl+a", () => {
   expect(result.meta).toBe(false)
 })
 
+test("parseKeypress - Kitty keyboard uppercase primary codepoint normalizes to lowercase name", () => {
+  // tmux (and plain xterm) can put the *shifted* codepoint (65 = 'A') as the
+  // primary field rather than the base one. The name must still normalize to
+  // lowercase so `ctrl+shift+a` bindings match (issue #1184). Shift comes from
+  // the modifier mask. See parse.keypress-kitty.ts.
+  const options: ParseKeypressOptions = { useKittyKeyboard: true }
+
+  // Ctrl+Shift+A as tmux csi-u: \x1b[65;6u (mod 6 => shift+ctrl, primary 65='A')
+  const ctrlShiftA = parseKeypress("\x1b[65;6u", options)!
+  expect(ctrlShiftA.name).toBe("a")
+  expect(ctrlShiftA.shift).toBe(true)
+  expect(ctrlShiftA.ctrl).toBe(true)
+  expect(ctrlShiftA.meta).toBe(false)
+
+  // Shift+Z with uppercase primary: \x1b[90;2u
+  const shiftZ = parseKeypress("\x1b[90;2u", options)!
+  expect(shiftZ.name).toBe("z")
+  expect(shiftZ.shift).toBe(true)
+  expect(shiftZ.ctrl).toBe(false)
+})
+
 test("parseKeypress - Kitty keyboard alt+a", () => {
   const options: ParseKeypressOptions = { useKittyKeyboard: true }
   const result = parseKeypress("\x1b[97;3u", options)!
--- a/packages/core/src/lib/parse.keypress-kitty.ts
+++ b/packages/core/src/lib/parse.keypress-kitty.ts
@@ -373,7 +373,12 @@ export function parseKittyKeyboard(sequence: string): ParsedKey | null {
   } else {
     // It's a Unicode character
     if (codepoint > 0 && codepoint <= 0x10ffff) {
-      const char = String.fromCodePoint(codepoint)
+      // Normalize uppercase letters to lowercase so `name` reflects the
+      // physical key (the shift flag is derived from the modifier mask below).
+      // Terminals differ on whether they encode the shifted (65='A') or base
+      // (97='a') codepoint here, so this keeps `ctrl+shift+a` matching either way.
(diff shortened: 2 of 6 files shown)


kommander:
Reviewed this in depth, including verifying the parser against the exact bytes tmux (`input-keys.c`/`tty-keys.c`) and xterm emit at level 2. Direction is right (vim/nvim also request level 2, and it aligns tmux/xterm with how kitty-protocol terminals already behave in opentui), but a few things are worth noting before merge:

**1. Blast radius is wider than tmux.** `enableDetectedFeatures` sends `modifyOtherKeysSet` unconditionally, gated only on kitty not being active (`terminal.zig:372-374`) — no capability check. xterm, mintty, older iTerm2 etc. all move to level 2 too.

**2. Behavior changes for legacy control aliases (intentional, but unstated).** Level 2 encodes the combos level 1 exempts: Ctrl+J/M/H/I/[ stop arriving as `linefeed`/`return`/`backspace`/`tab`/`escape` and become letter+ctrl. Concrete fallout under tmux/xterm: Ctrl+J no longer triggers the default `linefeed` bindings (Textarea newline `Textarea.ts:96`, Input submit `Input.ts:82-84`, Select/TabSelect confirm `Select.ts:36`, `TabSelect.ts:33`); Ctrl+H no longer deletes; Ctrl+[ no longer escapes. Matches kitty-terminal behavior, but deserves a release note.

**3. The #1184 headline case (`ctrl+shift+a`) is only half-fixed.** Several stacks send the *shifted* codepoint, and neither extended parse path lowercases letters (unlike the raw-char path at `parse.keypress.ts:394-397`):
- `\x1b[27;6;97~` (kitty outer → tmux) → name `a`+ctrl+shift → matches ✅
- `\x1b[27;6;65~` (plain xterm, always) and `\x1b[65;6u` (tmux csi-u) → name `A`+ctrl+shift → does **not** match a `ctrl+shift+a` binding ❌

tmux explicitly punts C-S-A vs C-A disambiguation to applications (`tty-keys.c`). Fix: lowercase A–Z (keeping `shift: true`) in `parse.keypress.ts:347-355` and `parse.keypress-kitty.ts:373-387`.

**4. Regression: `useKittyKeyboard: null` + CSI-u tmux.** That config disables CSI-u parsing (`renderer.ts:1057,1174,1197`), and tmux 3.2–3.3a hardcodes CSI-u format (the xterm-format default only arrived in 3.4). Every modified key then parses to name `""` — including Ctrl+C, so `exitOnCtrlC` never fires (verified). At level 1 those tmux versions sent raw `0x03`, which worked.

**5. Unclean exits get worse.** `resetState` restores `>4;0m` (`terminal.zig:229-231`), but a crash/SIGKILL now leaks level 2 instead of level 1: the shell then receives `\x1b[27;5;99~` instead of `0x03` for Ctrl+C — the user can't even interrupt. Pre-existing leak paths, higher severity.

**6. No test coverage.** Nothing asserts the requested level, and the mock keyboard only emits xterm-format with lowercase codepoints (and maps `\n`→13, `mock-keys.ts:241`), so none of the newly-reachable encodings are exercised.

Suggested asks:
1. Release note for the ctrl+j/m/h/i/[ alias changes under tmux/xterm.
2. Lowercase-normalize shifted letters in both extended parse paths (or a linked follow-up) — otherwise `ctrl+shift+a` stays broken on plain xterm.
3. Tests: assert `\x1b[>4;2m` is emitted + parser round-trips for shifted codepoints (`27;6;65~`, `65;6u`).
4. Consider decoupling CSI-u parsing from the `useKittyKeyboard: null` protocol opt-out.

avncharlie:
Thanks for the thorough review

I've pushed a follow-up commit addressing the concrete correctness gap (your point 3): both extended parse paths now lowercase A–Z into `name` and report capitalization via the `shift` flag, mirroring the existing raw single-char path at `parse.keypress.ts:394-397`. So `ctrl+shift+a` now matches regardless of whether the terminal encodes the shifted codepoint (`65`) or the base one (`97`) — covering plain xterm's `\x1b[27;6;65~` and tmux csi-u's `\x1b[65;6u`, which were the two failing encodings. `baseCode` is left as-is since it feeds the separate layout-fallback matcher.

Tests added for both paths asserting the round-trip on the exact sequences you flagged (`27;6;65~`, `65;6u`), plus the lowercase-codepoint variant so the either-encoding guarantee is pinned.

On the remaining points:
- **1 & 2 (release note for the ctrl+j/m/h/i/[ alias changes):** agreed this is a behavior change worth documenting — happy to add a changelog/release note. Where would you prefer it lives?
- **4 (`useKittyKeyboard: null` + tmux 3.2–3.3a csi-u → Ctrl+C dead):** good catch. Worth noting this is a pre-existing coupling between csi-u *parsing* and the kitty-protocol *opt-out* that this change only surfaces on those specific (pre-3.4) tmux versions with that non-default config. It's orthogonal to the level-1→2 request and, given how narrow the trigger is, I don't think it should block this PR, and maybe could be fixed in a seperate PR?
- **5 (leaked level-2 mode on crash):** noted; perhaps a fix could be attempting to restore modifyOtherKeys on panics? But this still wouldn't catch all crashes (e.g SIGKILL)

simonklee:

PR #1232, opened 2026-07-06
fix(core): respect useKittyKeyboard: null to disable kitty keyboard protocol

posva:
`config.useKittyKeyboard ?? {}` coerced `null` to `{}`, enabling the default flags (disambiguate + alternateKeys) instead of disabling the protocol as the type comment and docs state. `buildKittyKeyboardFlags` already handled `null`; it just never received it, and `useKittyForParsing` could never be `false`.

Fix preserves `null` (only `undefined` falls back to `{}`), so `null` now yields flags 0 and disables kitty stdin parsing.

Added config-level regression tests to `renderer.kitty-flags.test.ts` asserting the public `renderer.useKittyKeyboard` getter through `createTestRenderer` for `null` (disabled), omitted (defaults on), and `{ events: true }` — the `null` case failed before the fix.

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -1054,7 +1054,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     if (process.platform === "linux") config.useThread = false
     lib.setUseThread(rendererPtr, config.useThread)
 
-    const kittyConfig = config.useKittyKeyboard ?? {}
+    // Only undefined falls back to the defaults: null explicitly disables the protocol
+    const kittyConfig = config.useKittyKeyboard === undefined ? {} : config.useKittyKeyboard
     const kittyFlags = buildKittyKeyboardFlags(kittyConfig)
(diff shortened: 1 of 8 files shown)


packages/core/src/renderer.ts
```diff
@@ -1054,7 +1054,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     if (process.platform === "linux") config.useThread = false
     lib.setUseThread(rendererPtr, config.useThread)
 
-    const kittyConfig = config.useKittyKeyboard ?? {}
+    // Only undefined falls back to the defaults: null explicitly disables the protocol
+    const kittyConfig = config.useKittyKeyboard === undefined ? {} : config.useKittyKeyboard
```
posva, line 1059: This is just to follow what's said in https://opentui.com/docs/core-concepts/keyboard/ but maybe keeping the original code and updating the docs to say to use `false` makes more sense? That would also require updating the existing tests for `buildKittyKeyboardFlags`

PR #1235, opened 2026-07-07
feat(forms): add @opentui/forms package with RadioButton

### LINKED ISSUE #1222: [New component] Radio buttons in OpenTUI
Hi @kommander / @kommander,

It will be great to have radio buttons in opentui to enhance its feature-set & flexibility in different use-cases.

I have already raised the PR, this along with some other components can be easily used in @opentui/forms package down the line.

Discussion is welcome.

Thanks 😄

viralcodex:
hi @simonklee | @kommander ,

Fixes: https://github.com/anomalyco/opentui/issues/1222

This is a proposal to add new component in OpenTUI - the radio button inside a new Package - forms.
Basic implementation is that the whole renderable has multiple groups mapped by unique names or ids (Map<ID, Set), all of them having 1 or more radio buttons that act as options to a certain choice, each of them groups can be cycled through either with TAB or ARROWS (up and down).

There's also a prop for users to add their own radio button styles, they need to pass a tuple [checked, unchecked] to handle both states. [extensibility]

## Doubts
what will be a better UX? change selection on changing focus? or keep moving and focus with an enter press? Currently tab only changes focus but arrow keys change focus + selection.
should I delegate handling groups of radio-buttons to user and how they want it instead of keeping track internally?

Currently this package isn't added to be built and published in the timeline and there could be issues i may have missed while creating a new package, please let me know.

Next up in line is checkboxes with hierarchies.

## Demo

https://github.com/user-attachments/assets/789af044-a308-404f-b85a-96ece294415c


simonklee:
This looks good @viralcodex — i'm out today but i'll try to get this merged soon.

viralcodex:
> This looks good @viralcodex — i'm out today but i'll try to get this merged soon.

sure @simonklee , once this gets merged, I will get the checkboxes ready

simonklee:
1. Put framework support in `forms` through `@opentui/forms/react` and `@opentui/forms/solid`. That's how we did it in the qrcode package. Remove the forms imports and default catalog entries from React and Solid so you must register the optional package explicitly. The adapters should provide typed module augmentation and event behavior. Solid can continue to support `on:changed` and `on:selected`. React needs typed callbacks or a something that doesn't hardcode forms into its reconciler.

2. Scope group state to a renderer instead of the JavaScript process. Keep the existing string API, backed something like a `WeakMap<RenderContext, Map<string, Set<...>>>`. Add a proper `group` setter that atomically moves a button between registries while stahying mutual exclusive.

3. Define a deterministic rule for initial selection. Maybe let the last registered button with `checked: true` win. Do not emit user-action events during construction.

4. For keyboard behavior, the Up and Down keys should move focus and selection together. Space should select the focused radio. Leave Tab traversal to the application or shared focus-routing layer. Forms should not depend on an unintentionally public `InternalKeyBinding` contract. Either make the required helpers part of the intentional public API or use the a keymap abstraction.

5. Build should externalize `@opentui/core`, emit declarations, rewrite workspace versions, and produce a publishable `dist/package.json`. Then include forms in the root build, publish and pre-publish scripts, preview packages, release artifacts, dry-run publication, and packed-consumer validation, etc.

viralcodex:
> 1. Put framework support in `forms` through `@opentui/forms/react` and `@opentui/forms/solid`. That's how we did it in the qrcode package. Remove the forms imports and default catalog entries from React and Solid so you must register the optional package explicitly. The adapters should provide typed module augmentation and event behavior. Solid can continue to support `on:changed` and `on:selected`. React needs typed callbacks or a something that doesn't hardcode forms into its reconciler.
> 2. Scope group state to a renderer instead of the JavaScript process. Keep the existing string API, backed something like a `WeakMap<RenderContext, Map<string, Set<...>>>`. Add a proper `group` setter that atomically moves a button between registries while stahying mutual exclusive.
> 3. Define a deterministic rule for initial selection. Maybe let the last registered button with `checked: true` win. Do not emit user-action events during construction.
> 4. For keyboard behavior, the Up and Down keys should move focus and selection together. Space should select the focused radio. Leave Tab traversal to the application or shared focus-routing layer. Forms should not depend on an unintentionally public `InternalKeyBinding` contract. Either make the required helpers part of the intentional public API or use the a keymap abstraction.
> 5. Build should externalize `@opentui/core`, emit declarations, rewrite workspace versions, and produce a publishable `dist/package.json`. Then include forms in the root build, publish and pre-publish scripts, preview packages, release artifacts, dry-run publication, and packed-consumer validation, etc.

made the changes accordingly now.

forms are now wired in just like qrcode pkg with react and solid reconcilers.

tab and enter are removed and up/down will do focus + select. space selects the focused option. used handleKeyPress instead of the keymap intrinsics

last selected option wins, on moving one option from one group to another, it keeps the selected state and the destination groups are cleared of their selection.

WeakMap in added for registries for context -> radio groups.

forms are now added to build + publish release as a package on new version.

kommander:
@viralcodex is the CI failure related to the changes or something else?

viralcodex:
> @viralcodex is the CI failure related to the changes or something else?

unsure, it happened after taking the pull from main. before it was passing. it's in ssh build and test, so shouldn't be coz of my changes

viralcodex:
> @viralcodex is the CI failure related to the changes or something else?

passed, it was flaky test for windows afaik.

PR #1240, opened 2026-07-11
fix(core): force full repaint on terminal resize

### LINKED ISSUE #1110: alternate-screen: rightmost column stays stale after horizontal resize (macOS Terminal.app)
### Summary

After resizing the terminal window horizontally, the **rightmost character column** (immediately left of the scrollbar) often keeps **stale content** from a previous frame. The rest of the TUI updates, but that column is not repainted and does not recover when resizing stops; multiple resizes may be needed before it clears.

This is **not** logo-specific: any UI that previously occupied the right edge (menus, borders, text) can leave fragments there.

### Environment

- **OS:** macOS (darwin)
- **Terminal:** Terminal.app (`TERM_PROGRAM=Apple_Terminal`, `TERM=xterm-256color`, `COLORTERM=truecolor`)
- **Consumer:** OpenCode CLI TUI (also reproduced on DevEco Code fork; same `@opentui/core`)
- **OpenTUI version:** 0.2.14 (bundled with opencode)
- **Screen mode:** default `alternate-screen`, `externalOutputMode: passthrough`

Image

**Note:** Warp does **not** show this issue on the same machine → suggests Terminal.app does not erase cells OpenTUI does not overwrite, combined with incremental resize handling.

### Steps to reproduce

1. Launch a full-screen OpenTUI app (e.g. `opencode` home screen).
2. Use a moderately wide window so the right edge shows dense UI (logo, hints, etc.).
3. **Horizontally shrink** the window (drag edge). Do **not** suspend with Ctrl+Z.
4. Observe the column directly left of the scrollbar.

### Expected behavior

After resize settles, every cell in `1..cols` × `1..rows` should match the current frame. Columns beyond the new width (if any were visible before shrink) should be cleared.

### Actual behavior

- The rightmost column shows **ghost characters** from the previous layout.
- Releasing the mouse does **not** auto-fix.
- Repeated resizes may eventually force a full enough redraw to clear it.

### Why this looks like an OpenTUI renderer issue

- Default TUI uses **`alternate-screen`**. In `processResize`, **`split-footer`** clears stale rows/columns when width shrinks (`moveCursorAndClear` + buffer clear), but **`alternate-screen` does not** — it only resizes buffers and calls `requestRender()` without `forceFullRepaintRequested`.
- Incremental render (`lib.render(..., force=false)`) can skip cells that compare equal in the render buffer, while **physical terminal cells** (especially on Terminal.app) remain dirty.
- Resume path was fixed in #1008 / #922 (force full repaint after `resume()`); **resize path has no equivalent**.

### Related links

**OpenTUI**
- #778 — resize layout / weird characters (may overlap)
- #922 / PR #1008 — stale cells due to diff render after resume (fixed on resume, not resize)
- #812 — hit grid not updated on shrink (different symptom, same resize class)

**OpenCode (consumer / prior art)**
- https://github.com/anomalyco/opencode/pull/25606 — Draft workaround for Apple Terminal resize (closed unmerged; scoped `TERM_PROGRAM=Apple_Terminal`)
- https://github.com/anomalyco/opencode/issues/10053 — manual resize to refresh; Warp OK (closed)
- https://github.com/anomalyco/opencode/issues/16327 — stale frame after Ctrl+Z + resize + fg (closed; resume fix in #1008)

**Similar mechanism elsewhere**
- https://github.com/anthropics/claude-code/issues/46898 — stale characters at row edges during partial updates

### Suggested fix (optional)

On `processResize` for `alternate-screen` (and possibly `main-screen`):

1. When `width < prevWidth`, emit terminal clear for exposed columns/rows (similar to split-footer), **or**
2. Set `forceFullRepaintRequested = true` for the next frame after resize debounce settles.

Regression test: shrink width in alternate-screen mode; assert no non-background glyphs remain in columns `> width` (or force=true called once per resize burst).

### Screenshots

Image

YayoRazo:
### Issue for this PR

Closes #1110

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Shrinking the terminal and growing it back (or shrinking width in
alternate-screen) leaves stale cells on screen — ghost columns/rows — until
you resize again. `processResize` resizes the renderer and requests a render
but never sets `forceFullRepaintRequested`, so the next frame diffs against a
stale buffer and skips cells the terminal still shows as dirty. Every other
screen-invalidating transition (applyScreenMode, resume in #1008/#922,
capability change, buffer reset) already forces a full repaint; resize was the
exception. This implements the fix suggested in #1110 (option 2): force a full
repaint after a resize. TypeScript-only, no native build.

Added two regression tests (main-screen and alternate-screen) asserting a
resize sets the full-repaint flag; both fail before the change and pass after.
`bun test packages/core/src/tests/renderer.control.test.ts` → 30 pass;
`oxfmt --check` and `oxlint --deny-warnings` clean. The fix reuses the same
repaint path the existing `resume()` tests already cover. I did not record a
live before/after TUI clip.

### Screenshots / recordings

N/A (internal render path).

### Checklist

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -3785,6 +3785,11 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     this._console.resize(this.width, this.height)
     this.root.resize(this.width, this.height)
     this.emit(CliRenderEvents.RESIZE, this.width, this.height)
+    // A terminal resize invalidates the on-screen contents the diff renderer
+    // compares against, so force the next frame to fully repaint. Without this a
+    // grow-back after a shrink leaves stale cells until another resize, matching
+    // every other geometry/mode transition (applyScreenMode, resume, etc.).
+    this.forceFullRepaintRequested = true
     this.requestRender()
   }
 
--- a/packages/core/src/tests/renderer.control.test.ts
+++ b/packages/core/src/tests/renderer.control.test.ts
@@ -144,6 +144,28 @@ test("resume() forces the next alternate-screen render to fully repaint", async 
   await expectStartedResumeForcesNextRender("alternate-screen")
 })
 
+async function expectResizeForcesNextRender(screenMode: "main-screen" | "alternate-screen"): Promise<void> {
+  renderer.destroy()
+  ;({ renderer, mockInput, mockMouse, renderOnce } = await createTestRenderer({ screenMode, width: 80, height: 24 }))
+  ;(renderer as any)._terminalIsSetup = true
+  ;(renderer as any).forceFullRepaintRequested = false
+
+  renderer.resize(120, 40)
+
+  // A terminal resize invalidates what the diff renderer believes is on screen,
+  // so the next frame must fully repaint. Without this, growing back after a
+  // shrink leaves stale cells until another resize (see processResize).
+  expect((renderer as any).forceFullRepaintRequested).toBe(true)
+}
+
+test("resize() forces the next main-screen render to fully repaint", async () => {
+  await expectResizeForcesNextRender("main-screen")
+})
+
+test("resize() forces the next alternate-screen render to fully repaint", async () => {
+  await expectResizeForcesNextRender("alternate-screen")
+})
+
 test("stop() transitions to EXPLICIT_STOPPED and stops rendering", () => {
   renderer.start()
   expect(renderer.isRunning).toBe(true)


simonklee:
I’m closing the PR because the patch does not change the behavior of the reported resize path.

For both main-screen and alternate-screen, a terminal dimension change already calls native resizeRenderer(). Native resize clears currentRenderBuffer with CLEAR_CHAR and initializes nextRenderBuffer with normal cells, so every viewport cell already differs and is emitted on the next frame with force=false.

I also retested the regression in on a Mac and it's still there so I don't think this is the right path.

PR #1246, opened 2026-07-11
fix(solid): fall back to append when anchor is not found in reconciler

### LINKED ISSUE #1245: bug(solid): reconciler passes anchorIndex -1 to parent.add when anchor not found
## Description

In the Solid reconciler, when an anchor node is not found in the parent's children list (`children.indexOf(anchor)` returns -1), the code logs a warning but still passes `-1` to `parent.add(node, anchorIndex)`. This silently inserts the node at an incorrect position.

## Steps to reproduce

1. Use Solid JSX with conditional rendering that inserts/removes nodes
2. When the reconciler tries to insert a node before an anchor that doesn't exist in the parent (e.g., due to slot resolution or stale references)
3. The node gets appended/indexed incorrectly instead of either inserting at the correct position or gracefully falling back

## Expected behavior

When the anchor cannot be found, the node should be appended at the end as a safe fallback.

## Actual behavior

`parent.add(node, -1)` is called, which due to internal handling in `Renderable.add()` results in the node being placed at an unexpected position.

YayoRazo:
### Issue for this PR

Closes #1245

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

In the Solid reconciler's `_insertNode` function, when `children.indexOf(anchor)` returns -1 (anchor not found in parent's children), the previous code logged a warning but still called `parent.add(node, anchorIndex)` with `anchorIndex = -1`. This silently placed the node in an incorrect position.

The fix appends the node without an anchor (`parent.add(node)`) and returns early, which is the safe fallback when the expected anchor position cannot be determined.

Reviewed the `_insertNode` function flow and confirmed that `parent.add(node)` (no index) correctly appends to the end of the children list, matching the behavior of `add()` when called without an index argument.

### Screenshots / recordings

N/A

### Checklist

### DIFF
--- a/packages/solid/examples/repro-stale-slot-anchor.tsx
+++ b/packages/solid/examples/repro-stale-slot-anchor.tsx
@@ -1,0 +1,96 @@
+import { onMount } from "solid-js"
+import { createSlotNode, insertNode, render, useRenderer } from "@opentui/solid"
+import type { TextRenderable } from "@opentui/core"
+import { TextNodeRenderable } from "@opentui/core"
+
+process.env.DEBUG = "true"
+
+/**
+ * Repro for #1245 / #1246: reconciler passes anchorIndex -1 to
+ * `parent.add()` when the anchor can't be found in the parent's children.
+ *
+ * Root cause (packages/solid/src/elements/slot.ts, `SlotRenderable`):
+ * a slot's per-parent placeholder (`layoutNodesByParent`/`textNodesByParent`)
+ * only gets forgotten in `didRemoveSlotChild` when the slot is still attached
+ * to *another* parent. When a <For>/<Show> expression's placeholder is torn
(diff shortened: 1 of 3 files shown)


Adictya:
The behavior fix makes sense, however i am still concerned about this edgecase being hit. I could not get a repro with #1245 , could you help me with a component that reproduces it similar to how we have other repros in the repo and can you add some tests as well ?

simonklee:
Thanks for adding the example and tests. I reviewed and reproduced the sequence, but it requires manually removing a slot placeholder and then reusing that detached slot as an anchor through low-level renderer primitives. The example also notes that compiled JSX does not naturally produce this relationship. Current control-flow and slot tests, including empty/non-empty transitions, do not reach the missing-anchor path.

PR #1257, opened 2026-07-11
fix(core): guard setRawMode in resume() so input always re-attaches

### LINKED ISSUE #1255: bug: resume() leaves input dead on Windows when setRawMode throws
On Windows, `CliRenderer.resume()` can leave the terminal permanently unresponsive — no input, no output — while the render loop keeps running.

`resume()` re-arms input in this order: `setRawMode(true)` → drain buffered bytes → re-attach the stdin `data` listener → `stdin.resume()`. On Windows, `uv_tty_set_mode` can throw (`EINVAL` / `ENOTTY` / `EBADF`) when the console handle is in a transient state right after a child process (editor / pager / git) released it. The throw is unguarded, so it unwinds `resume()` before the `data` listener is re-attached, and input is never restored. (`finalizeDestroy()` already wraps its `setRawMode(false)` in try/catch; `resume()` does not.)

### Repro

Run an opentui app on Windows that suspends and resumes the renderer around a shell-out (e.g. opencode's `/editor`). After the child returns, the terminal is frozen — no input, no output — until the process is killed.

### Suggested fix

Wrap the `setRawMode(true)` call in try/catch so the `data`-listener re-attach and `stdin.resume()` always run; fall back to cooked mode if raw mode can't be restored.

YayoRazo:
## Summary

On Windows, `resume()` can leave the terminal permanently unresponsive — no input, no output — while the render loop keeps running. `resume()` re-arms input as `setRawMode(true)` → drain buffered bytes → re-attach the stdin `data` listener → `stdin.resume()`. On Windows `uv_tty_set_mode` can throw (`EINVAL` / `ENOTTY` / `EBADF`) when the console handle is in a transient state right after a child process (editor / pager / git) released it. The unguarded throw unwinds `resume()` before the `data` listener is re-attached, so input is never restored. `finalizeDestroy()` already wraps its `setRawMode(false)` in try/catch; `resume()` did not.

This PR wraps the `setRawMode(true)` call in try/catch (logging on failure) so the `data`-listener re-attach and `stdin.resume()` always run. If raw mode can't be restored, input falls back to cooked mode instead of dying.

## Repro

On Windows, run an opentui app that suspends and resumes the renderer around a shell-out (e.g. opencode's `/editor`). After the child returns, the terminal is frozen — no input, no output — until the process is killed. Likely the opentui-side cause of anomalyco/opencode#11877.

## Verification

`bun test` — `renderer.focus` / `renderer.focus-restore` / `renderer.lifecycle` / `renderer.idle`: 41 pass, 0 fail. `oxlint` clean.

Closes #1255

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -4024,7 +4024,16 @@ export class CliRenderer extends EventEmitter implements RenderContext {
 
   public resume(): void {
     if (this.stdin.setRawMode) {
-      this.stdin.setRawMode(true)
+      try {
+        this.stdin.setRawMode(true)
+      } catch (error) {
+        // On Windows, uv_tty_set_mode can throw (EINVAL/ENOTTY/EBADF) when the
+        // console handle is in a transient state — e.g. right after shelling
+        // out to a child (git/editor/pager) that grabbed the console. Never let
+        // that skip re-attaching the "data" listener + resume() below, or input
+        // stays permanently dead while the render loop keeps running.
+        console.error("Failed to restore raw mode on resume:", error)
+      }
     }
 
     // Drain any input buffered during suspension before registering the
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -3982,6 +3982,26 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     this.internalPause()
   }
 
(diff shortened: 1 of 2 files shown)


simonklee:
I'm just closing this because i haven't really found that you've tested these things on windows so i cannot really trust this.

PR #1263, opened 2026-07-14
fix(core/audio): Fix audio decoder error reporting race

simonklee:
Preserve the native decoder failure when end-of-input races with the
decoder entering a terminal state, instead of reporting a generic stream
end error.


simonklee:
This is to fix the flaky test as well.

PR #1276, opened 2026-07-17
fix(core): guard non-string bundled asset path

### LINKED ISSUE #1275: BUG: @opentui/core@0.4.5 hard-crashes at import in bun --compile binaries — normalizeLoadedFilePath calls .startsWith on an undefined worker path
## Summary

Since `0.4.5`, importing `@opentui/core` **crashes at module load** inside a `bun build --compile` standalone binary:

```
undefined is not an object (evaluating '$.startsWith')
```

0.4.5 added a module-scope top-level `await resolveBundledFilePath(...)` that eagerly resolves the tree-sitter `parser.worker` path at import time. When the `{ with: { type: "file" } }` import's `.default` is `undefined` in the compiled binary, that `undefined` reaches `normalizeLoadedFilePath`, which unconditionally calls `loadedPath.startsWith("file:")` → `TypeError`. Because the resolution is a top-level await at module scope, the whole `@opentui/core` import throws before `TreeSitterClient.startWorker()`'s existing fallback can run — so it's fatal, not a graceful loss of highlighting.

This is a **regression** of the worker-asset problem in #807: on ≤0.4.3 the worker just failed to spawn and the app degraded to plain text; 0.4.5 makes it a hard startup crash.

## Environment

- `@opentui/core`: `0.4.5`
- Bun: `1.3.14`
- OS: macOS, Apple Silicon (darwin-arm64)
- Consumer: a `bun build --compile` standalone binary (reproduced via opencode — see anomalyco/opencode#37556)

## Root cause (source at `v0.4.5`)

`packages/core/src/platform/runtime-assets.bun.ts:14` — resolution runs at **module scope** via top-level await:

```ts
const bundledTreeSitterWorkerPath = await resolveBundledFilePath(
  PARSER_WORKER_ASSET_KEY,
  () => import("@opentui/core/parser.worker" as string, { with: { type: "file" } }),
  new URL("../lib/tree-sitter/parser.worker.js", import.meta.url),
  import.meta.url,
  { useAssetRoot: false },
)
```

`packages/core/src/platform/runtime.ts:72` — `.default` is passed through unchecked:

```ts
return normalizeLoadedFilePath((await loadBundledFile()).default, metaUrl)
```

`packages/core/src/platform/runtime.ts:80` — and dereferenced with no guard (annotated `: string`, but `undefined` at runtime):

```ts
function normalizeLoadedFilePath(loadedPath: string, baseUrl: string): string {
  if (loadedPath.startsWith("file:")) {   // ← TypeError when loadedPath is undefined
```

## Why `.default` is `undefined` (tested)

In a `bun build --compile`, if the consumer's build graph emits `@opentui/core/parser.worker` as its **own entrypoint/chunk**, the `import(..., { with: { type: "file" } })` specifier resolves to that JS module namespace instead of a `type: "file"` asset — so `.default` is `undefined` instead of a `/$bunfs/...` path. Confirmed with a minimal `Bun.build` probe:

- with `parser.worker` as an entrypoint → the file-import's `.default` is `undefined`
- without it → `.default` is a valid `/$bunfs/root/parser.worker-*.js` path

opencode hits exactly this (its `build.ts` lists `parser.worker.js` as a `Bun.build` entrypoint — anomalyco/opencode#37556). But regardless of *why* the asset doesn't resolve (this, or #807's relocated-binary case), opentui shouldn't hard-crash at import over it.

## Reproduction

Build any `@opentui/core@0.4.5` app with `bun build --compile` where `parser.worker` ends up as its own chunk (e.g. opencode `dev` at anomalyco/opencode@a46374e90e, before their workaround: `bun packages/opencode/script/build.ts --single`), run the binary → crash at launch. `--version` / non-TUI paths that never import the render graph are unaffected.

Note: opencode has since worked around this on its side (anomalyco/opencode@ba3f966ed0) by injecting the worker source as a virtual `Bun.build` entrypoint so the real file no longer collides with the `type: "file"` import — so current opencode `dev` no longer reproduces. The library-side crash remains for any other compiled consumer that hits a non-string `default`, which is what #1276 guards.

## Proposed fix

Guard the deref in `resolveBundledFilePath`'s Bun branch: if the bundled module's `default` is not a path string, resolve the source fallback path (exactly what the non-Bun branch already does) instead of throwing. Implemented in #1276, with a regression test.

A single guard is sufficient — no lazy-resolution refactor needed — because the graceful-degradation machinery already exists downstream: `TreeSitterClient.startWorker()` installs `worker.onerror` handling that logs, emits, and degrades to plain text. The import-time crash simply fired before that machinery could run; the guard reconnects it. Consumers that set `OTUI_TREE_SITTER_WORKER_PATH` (env or build-time define) bypass the default resolution path entirely and keep full highlighting. Making the module-scope resolution lazy instead would push async through the synchronous `resolveWorkerPath()` chain in `client.ts` and both `runtime-assets` platform variants, and it would resolve to the same fallback path anyway.

Related: #807 (pre-0.4.5 silent version), anomalyco/opencode#37556 (downstream trigger).

CasualDeveloper:
## Problem

Since `@opentui/core@0.4.5`, importing the package inside a `bun build --compile` binary crashes at module load with `undefined is not an object (evaluating 'loadedPath.startsWith')`. The Bun branch of `resolveBundledFilePath` dereferences a `type: "file"` import's `.default` unconditionally, but that value can be `undefined` — e.g. when the same file is also emitted as a `Bun.build` entrypoint, so the specifier resolves to a JS module namespace with no path `default`.

## Solution

- Guard the Bun branch of `resolveBundledFilePath`: when the bundled module's `default` is not a string, fall back to the source path via the existing `resolveFallbackFilePath` (mirroring the non-Bun branch) instead of dereferencing `undefined`.
- Add a regression test covering the non-string `default` case.
- This reconnects failures to the existing downstream handling rather than adding new machinery: `TreeSitterClient.startWorker()` already installs `worker.onerror` (logs + emits + degrades to plain text) — the import-time crash just happened before that could run. Net effect: pre-0.4.5 graceful degradation is restored.
- Consumers that set `OTUI_TREE_SITTER_WORKER_PATH` (env or build-time define) regain full highlighting, since it takes precedence in `resolveWorkerPath()` over the default resolution path where the crash lived.

## Notes

- Testing: `bun test src/platform/runtime.test.ts` → 8 pass / 0 fail (includes the new regression test, which fails on the unpatched code with the exact `startsWith` TypeError). `oxfmt --check` and `oxlint` clean on both changed files.
- Scope: guard only, deliberately. Making the module-scope resolution lazy would ripple async through the synchronous `resolveWorkerPath()` chain in `client.ts` and both `runtime-assets` platform variants, and with this guard in place it would resolve to the same fallback path anyway — happy to explore that separately if preferred.
- Fixes #1275. Downstream trigger: anomalyco/opencode#37556 (which has been worked around by downgrading OpenCode to use OpenTUI 0.4.3).
- AI Assistance: OpenCode + Claude Opus 4.8
- Work attribution: Human operator driven and reviewed

### DIFF
--- a/packages/core/src/platform/runtime.test.ts
+++ b/packages/core/src/platform/runtime.test.ts
@@ -84,6 +84,24 @@ describe("platform/runtime", () => {
     expect(fallbackCalled).toBe(!isBun)
   })
 
+  test("falls back to the source path when a bundled asset module's default is not a path string", async () => {
+    const fallbackUrl = new URL("./fallback-tree-sitter.wasm", import.meta.url)
+    let fallbackCalled = false
+
+    const resolved = await resolveBundledFilePath(
+      "web-tree-sitter/tree-sitter.wasm",
+      async () => ({ default: undefined as unknown as string }),
+      () => {
+        fallbackCalled = true
+        return fallbackUrl
+      },
+      import.meta.url,
+    )
+
+    expect(resolved).toBe(fileURLToPath(fallbackUrl))
+    expect(fallbackCalled).toBe(true)
+  })
+
   test("resolves Bun-emitted asset modules when a non-Bun bundle has no source fallback", async () => {
     const bundledUrl = new URL("./bundled-tree-sitter.wasm", import.meta.url).href
     const bundledModuleSpecifier = `data:text/javascript,${encodeURIComponent(
--- a/packages/core/src/platform/runtime.ts
+++ b/packages/core/src/platform/runtime.ts
@@ -69,7 +69,14 @@ export async function resolveBundledFilePath(
     return (await loadBundledFilePath(loadBundledFile, metaUrl, options.loadBundledFileFallback ?? false)) ?? path
   }
 
-  return normalizeLoadedFilePath((await loadBundledFile()).default, metaUrl)
+  // A bundled `type: "file"` import can resolve to a module whose `default` is not a path string
+  // (e.g. when the same file is also emitted as a build entrypoint), so fall back to the source path.
+  const loaded = (await loadBundledFile()).default
+  if (typeof loaded !== "string") {
+    return resolveFallbackFilePath(fallbackPath, metaUrl)
+  }
+
+  return normalizeLoadedFilePath(loaded, metaUrl)
 }
 
 function resolveFallbackFilePath(fallbackPath: FilePathFallback, metaUrl: string): string {


CasualDeveloper:
Superseded by #1293, which implements the same guard and regression test. Thanks!

PR #1295, opened 2026-07-28
fix(edit-buffer): extract ranges using configured width method

### LINKED ISSUE #1287: bug(core): EditBuffer.getText()/getTextRange() silently truncate at 1 MiB with no way to detect it
## Summary

`EditBuffer.getText()` reads into a fixed 1 MiB buffer and returns whatever fits, with no signal that truncation occurred. The buffer itself holds the full content — `lineCount` and rendering are correct — so the loss only appears when reading text back out.

For an editor built on `TextareaRenderable`, `plainText` is the natural thing to write to disk. That makes this a silent data-loss path: open/paste something over 1 MiB, save, and the file is cut to 1048576 bytes with no error anywhere.

## Source

`packages/core/src/edit-buffer.ts`:

```ts
public getText(): string {
  this.guard()
  const maxSize = 1024 * 1024 // 1MB max
  const textBytes = this.lib.editBufferGetText(this.bufferPtr, maxSize)
  if (!textBytes) return ""
  return this.lib.decoder.decode(textBytes)
}
```

`getTextRange()` has the same `const maxSize = 1024 * 1024`.

The native side writes up to `maxLen` and returns the byte count, so truncation is not distinguishable from a complete read:

```zig
export fn editBufferGetText(edit_handle: NativeHandle, outPtr: ?[*]u8, maxLen: u32) u32 {
    const object_ptr = acquireEditBuffer(edit_handle) orelse return 0;
    if (maxLen == 0) return 0;
    const out = outPtr orelse return 0;
    const outBuffer = out[0..@as(usize, maxLen)];
    return @intCast(object_ptr.getText(outBuffer));
}
```

## Reproduction

```ts
import { createTestRenderer } from "@opentui/core/testing"
import { TextareaRenderable } from "@opentui/core"

const { renderer } = await createTestRenderer({ width: 80, height: 24 })
const ta = new TextareaRenderable(renderer, { flexGrow: 1 })

const source = Array.from({ length: 60_000 }, (_, i) => `line ${i} with some filler text`).join("\n")
ta.setText(source)

console.log("source length   :", source.length)
console.log("plainText length:", ta.plainText.length)
console.log("lineCount       :", ta.lineCount)
```

Output:

```
source length   : 1968889
plainText length: 1048576
lineCount       : 60000
```

`lineCount` reports all 60000 lines, so the data is in the buffer; only the readout is cut. Nothing throws.

## Expected

Any of:

- `getText()` returns the full contents, sizing the output buffer from the actual length; or
- the API exposes the true byte length so callers can size their own read; or
- truncation is signalled (throw, or a documented `{ text, truncated }` shape) instead of returning a short string that looks complete.

## Secondary: errors are indistinguishable from empty results

```zig
return @intCast(object_ptr.getTextRange(start_offset, end_offset, outBuffer) catch 0);
```

`catch 0` collapses a native error into `0`, which the JS wrapper turns into `null` and then `""`. A caller cannot tell "this range failed" from "this range is empty":

```ts
const long = new TextareaRenderable(renderer, { flexGrow: 1 })
long.setText("x".repeat(1_500_000))
const eb = long.editBuffer as any
eb.getTextRange(0, 1000)              // "xxxxxxxxxx..."  fine
eb.getTextRange(1_048_576, 1_048_586) // ""               error or empty? unknowable
```

This also defeats the obvious workaround of reading the buffer in sub-1 MiB chunks: the loop terminates early on a `""` that it has to treat as end-of-content. Chunked reads do work for buffers of short lines, but stop at 1048576 once individual lines get large, which made the limit look buffer-wide rather than per-call while debugging.

## Environment

- `@opentui/core` 0.4.4; constants unchanged on `main` as of filing (0.4.5 latest)
- Bun 1.3.14, Windows 11 x64

simonklee:
getTextRange always measured with Unicode rules, so wcwidth text
past 65535 columns and ranges starting inside a grapheme returned
the wrong bytes.

Honor the configured width method while preserving grapheme
snapping, stop traversal at the range end, avoid copying exact-sized
FFI output buffers, and widen rope balance math so large weights do
not wrap.

Fixes https://github.com/anomalyco/opentui/issues/1287


simonklee:

PR #1304, opened 2026-07-30
text-buffer: widen chunk width past u16

simonklee:
Lines longer than 65535 columns saturated chunk width at u16
max, so cursor, viewport, and rope weight drifted. Store width
as u32 so horizontal extent stays accurate for very long lines.


simonklee:
This change is already included in #1428. The main branch now uses 32-bit text
chunk widths and tests a line with 70,000 columns. Closing because its change
is no longer needed.

PR #1307, opened 2026-07-30
feat(core): add Ghostty terminal runtime

jlongster:
## Problem

OpenTUI does not currently have a native terminal-state runtime for embedding a PTY-backed terminal. Adding the renderable, compositor, FFI, and distribution work in one change would make the core lifecycle and input contract difficult to review independently.

This PR introduces only the runtime boundary around `libghostty-vt`: terminal ownership, VT input, mode-aware input encoding, and terminal-generated PTY responses. It deliberately does not expose a public TypeScript API or render anything yet.

## Changes

- Dynamically load the required `libghostty-vt` runtime symbols from an explicit path or `OPENTUI_GHOSTTY_VT_LIBRARY`.
- Own and clean up the terminal, key encoder/event, and mouse encoder/event across partial initialization failures.
- Support VT writes, resize, viewport scrolling, keyboard, mouse, paste, and focus encoding.
- Capture terminal-generated PTY responses for the host to drain.
- Add a host-libc native test target so Linux integration tests can load a glibc Ghostty library without changing OpenTUI's default musl test target.
- Cover missing-library behavior, lifecycle and validation, mode-aware encoding, and partial response drains.

The loader intentionally does not probe arbitrary system library names. `libghostty-vt` is still an evolving ABI, so a later packaging layer will provide the exact compatible library path.

## Scope

This is the first layer of the embedded-terminal stack. It excludes the OpenTUI buffer compositor, native handle/FFI exports, TypeScript renderable, runtime packaging, benchmarks, and examples.

## Testing

- `bun run test:native` (1709 passed, 8 skipped)
- `OPENTUI_GHOSTTY_VT_LIBRARY=... bun run test:native -Dtest-host-libc=true` (1713 passed, 4 skipped)
- `bun run build`
- `bun run fmt:check`
- `bun run lint`

### DIFF
--- a/packages/core/src/zig/build.zig
+++ b/packages/core/src/zig/build.zig
@@ -76,6 +76,14 @@ fn nativeExecutableTarget(b: *std.Build) std.Build.ResolvedTarget {
     return b.resolveTargetQuery(query);
 }
 
+fn nativeTestTarget(b: *std.Build, use_host_libc: bool) std.Build.ResolvedTarget {
+    if (builtin.os.tag != .linux or !use_host_libc) return nativeExecutableTarget(b);
+    var query = b.graph.host.query;
+    query.abi = .gnu;
+    query.glibc_version = .{ .major = 2, .minor = 17, .patch = 0 };
+    return b.resolveTargetQuery(query);
+}
+
 fn pathExists(path: []const u8) bool {
     if (path.len == 0) return false;
     std.fs.cwd().access(path, .{}) catch return false;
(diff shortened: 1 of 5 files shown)


jlongster:
Superseded by #1317. The replacement foundation pins Ghostty, builds it separately, imports the official C API with `@cImport`, and absorbs the static archive into `libopentui`, removing runtime discovery and ABI drift. This branch remains intact as reference for the terminal lifecycle and input implementation.

PR #1308, opened 2026-07-31
feat(core): add embedded terminal renderable

jlongster:
## Problem

The first stack layer gives OpenTUI an owned `libghostty-vt` terminal runtime, but it cannot display that state or participate in OpenTUI input, focus, layout, and cursor behavior. Landing the renderer together with runtime packaging and screen-observation APIs would make the native composition and lifecycle contract difficult to review.

This PR adds the smallest user-facing layer: a buffered `EmbeddedTerminalRenderable` that translates Ghostty render state directly into an OpenTUI `OptimizedBuffer`. The compatible Ghostty library is still supplied explicitly; distribution is left to the next stack layer.

## Changes

- Bind Ghostty render state, dirty rows, cell styles, colors, wide cells, and cursor state.
- Compose incremental Ghostty updates directly into OpenTUI native buffers.
- Add portable native handles and C/TypeScript FFI methods for lifecycle, composition, cursor queries, input encoding, scrolling, and PTY responses.
- Add `EmbeddedTerminalRenderable` with focus, keyboard, paste, mouse, resize, cursor, and teardown behavior.
- Reject invalid FFI dimensions and nullable pointer/length pairs before crossing native boundaries.
- Retry key encoding using the Ghostty-reported output capacity, cap queued terminal responses at 1 MiB, and keep callback failures from blocking native cleanup.
- Cover native composition, clipping, dirty rows, wide characters, cursor state, input modes, resize, focus cleanup, long input, and idempotent destruction.

## Scope

This is the second layer of the embedded-terminal stack and targets `ghostty-runtime`. It intentionally excludes Ghostty runtime packaging and asset resolution, screen snapshots/change notifications, benchmarks, and examples. Callers must provide `libraryPath` or `OPENTUI_GHOSTTY_VT_LIBRARY`.

## Testing

- `bun run test:native` (1709 passed, 10 skipped)
- `OPENTUI_GHOSTTY_VT_LIBRARY=... bun run test:native -Dtest-host-libc=true` (1715 passed, 4 skipped)
- Focused `EmbeddedTerminalRenderable` suite (6 passed)
- `bun run test:js:node` (4518 passed, 6 skipped)
- `bun run build`
- `bun run test:dist`
- `bun run fmt:check`
- `bun run lint`

PR #1317, opened 2026-07-31
build(core): statically link libghostty

jlongster:
## Problem

The first embedded-terminal stack dynamically loaded `libghostty-vt`. That required runtime library discovery, exposed callers to Ghostty ABI drift, split packaging across two native libraries, and left the successful Ghostty integration tests dependent on an externally configured artifact.

This PR replaces that foundation with a pinned, statically linked Ghostty VT build. Ghostty is still compiled separately through its official C API, so its toolchain remains independent from OpenTUI implementation code, but consumers receive one self-contained `libopentui`.

This supersedes the runtime foundation in #1307 and #1308. The original implementation remains available on their branches and in the `terminal-experiment` branch as reference for the next stack layers.

## Changes

- Pin Ghostty to commit `74d0c72fd9318ad3ab95bfb56f6c2d995e267e2e` and Zig 0.15.2.
- Build target-specific PIC static archives from immutable source with baseline CPU features and SIMD disabled, matching the proven prototype artifact.
- Import the official `ghostty/vt.h` API through Zig `@cImport` with `GHOSTTY_STATIC`.
- Link the exact GNU, musl, macOS 13, and Windows GNU archive into each `libopentui` target.
- Hide Ghostty C symbols inside the final shared library so another Ghostty installation cannot interpose them.
- Hash and validate the SDK headers and archives before reuse or external override.
- Include Ghostty, uucode, UTF-8 decoder, and Unicode notices in npm native packages and release ZIPs.

## Scope

This is the replacement first stack layer. It establishes only the reproducible static build and C-import boundary. It does not yet add terminal ownership APIs, OpenTUI composition, TypeScript FFI, `EmbeddedTerminalRenderable`, or synthetic Ghostty behavior tests; those will be added with the real terminal integration in subsequent layers using the original experiment as reference.

The generated Ghostty source checkout, headers, and archives stay in `.cache/ghostty-vt` and are not published. Native packages continue to contain only `libopentui` plus required notices.

## Testing

- `bun run test:native` (1862 passed, 6 skipped)
- `bun run build` from `packages/core`
- `bun run test:dist --skip-build` (packed Node and Bun consumers passed)
- `bun run fmt:check`
- `bun run lint`
- Cross-link verification for Linux GNU, Linux musl, macOS, and Windows GNU
- Linux dependency inspection: no dynamic Ghostty dependency or exported `ghostty_*` API
- Previous GitHub CI passed across Linux, macOS, Windows, Node, Bun, native builds, and packed consumers; the latest removal commit is running the same matrix


packages/core/src/zig/build.zig
```diff
@@ -23,8 +23,8 @@ const SUPPORTED_TARGETS = [_]SupportedTarget{
     .{ .zig_target = "aarch64-linux-gnu.2.17", .output_name = "aarch64-linux", .description = "Linux aarch64" },
     .{ .zig_target = "x86_64-linux-musl", .output_name = "x86_64-linux-musl", .description = "Linux x86_64 (musl)" },
     .{ .zig_target = "aarch64-linux-musl", .output_name = "aarch64-linux-musl", .description = "Linux aarch64 (musl)" },
-    .{ .zig_target = "x86_64-macos", .output_name = "x86_64-macos", .description = "macOS x86_64 (Intel)" },
-    .{ .zig_target = "aarch64-macos", .output_name = "aarch64-macos", .description = "macOS aarch64 (Apple Silicon)" },
+    .{ .zig_target = "x86_64-macos.13.0", .output_name = "x86_64-macos", .description = "macOS x86_64 (Intel)" },
+    .{ .zig_target = "aarch64-macos.13.0", .output_name = "aarch64-macos", .description = "macOS aarch64 (Apple Silicon)" },
```
jlongster, line 27: This does impact opentui: it explicitly supports only macOS 13 and later. Did we intentionally want to support 12 or earlier?

PR #1318, opened 2026-08-02
fix(core): don't warn when tree-sitter highlight fails during client destroy

### LINKED ISSUE #1319: Code highlighting failed, falling back to plain text warning logged on exit (destroy race)
## Summary

When an app using `@opentui/core` exits, `TreeSitterClient.destroy()` rejects any in-flight `highlightOnce()` requests. `CodeRenderable` treats those rejections as genuine highlighting failures and logs a misleading warning:

```text
Code highlighting failed, falling back to plain text: Error: TreeSitter client destroyed
```

This occurs during normal shutdown, so nothing is actually wrong. As a result, applications such as `opencode` print this warning on every exit (see anomalyco/opencode#36454).

## Reproduction

1. Run any OpenTUI app that renders a `CodeRenderable` (or markdown containing code).
2. Exit the application while syntax highlighting is still in progress (e.g. `Ctrl+C` or `/exit`).
3. Observe the warning printed after teardown.

## Root Cause

During shutdown, `TreeSitterClient.destroy()` rejects all pending highlight requests with:

```ts
new Error("TreeSitter client destroyed")
```

`CodeRenderable.startHighlight()` catches every rejected `highlightOnce()` call and assumes it represents a real highlighting failure, logging:

```ts
console.warn(
  "Code highlighting failed, falling back to plain text:",
  error,
);
```

However, in this case the rejection is an expected consequence of teardown, not an actual error.

## Expected Behavior

Shutdown-related cancellations should be treated as expected and should not:

- emit a warning, or
- trigger the normal highlighting failure path.

Actual highlighting failures (for example, parser initialization or worker errors) should continue to log warnings so genuine issues remain visible.

HARIOM-JHA01:
Fixes #1319

## Problem

When @opentui/core is torn down, TreeSitterClient.destroy() rejects any in-flight highlightOnce() promises with a generic `Error("TreeSitter client destroyed")`. CodeRenderable.startHighlight() treats every rejection as a real failure and logs a misleading warning:

    Code highlighting failed, falling back to plain text: Error: ...

This races with client teardown on exit, so opencode users see the warning unexpectedly on shutdown even though nothing is actually broken.

Reported in anomalyco/opencode#36454.

## Fix

- client.ts: introduce a typed `TreeSitterClientDestroyedError` and reject pending requests (and pending initialization) with it in `destroy()`.
- Code.ts: silently bail out when highlighting fails due to client destroy (or the renderable is already destroyed), preserving warnings for genuine worker failures.
- Testing: mock-tree-sitter-client supports `rejectHighlightOnce`; client destroy tests now assert `TreeSitterClientDestroyedError`; 
- Code renderer tests verify no `console.warn` on destroy-triggered failures while a real Error still warns.
## Tests

- bun test packages/core/src/lib/tree-sitter/client.test.ts (49 pass)
- oxlint / oxfmt clean
- Renderer tests (Code.test.ts) require the native lib and run in CI.

### DIFF
--- a/packages/core/src/lib/tree-sitter/client.test.ts
+++ b/packages/core/src/lib/tree-sitter/client.test.ts
@@ -1,5 +1,5 @@
 import { test, expect, beforeEach, afterEach, beforeAll, describe } from "bun:test"
-import { TreeSitterClient } from "./client.js"
+import { TreeSitterClient, TreeSitterClientDestroyedError } from "./client.js"
 import { tmpdir } from "os"
 import { join } from "path"
 import { existsSync } from "fs"
@@ -145,8 +145,7 @@ describe("TreeSitterClient", () => {
       const outcome = await initializeOutcome
       expect(outcome.status).toBe("rejected")
       if (outcome.status === "rejected") {
-        expect(outcome.error).toBeInstanceOf(Error)
-        expect((outcome.error as Error).message).toBe("Client destroyed during initialization")
+        expect(outcome.error).toBeInstanceOf(TreeSitterClientDestroyedError)
       }
       expect(client.isInitialized()).toBe(false)
     } finally {
@@ -1228,8 +1227,8 @@ describe("TreeSitterClient Edge Cases", () => {
     // Immediately destroy
     await client.destroy()
 
-    // Init promise should reject with specific error
-    await expect(initPromise).rejects.toThrow("Client destroyed during initialization")
+    // Init promise should reject with a destroy-specific error
+    await expect(initPromise).rejects.toBeInstanceOf(TreeSitterClientDestroyedError)
     await new Promise((resolve) => setTimeout(resolve, 0))
 
     expect(client.isInitialized()).toBe(false)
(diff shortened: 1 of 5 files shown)


jizhi0v0:
Still reproducing on opencode **1.18.18** (macOS, arm64). That tag pins `@opentui/core` **0.4.5**, and the shipped bundle still has the `console.warn` ahead of the guard:

```
$ strings -a ~/.opencode/bin/opencode | grep -o 'Code highlighting failed.\{0,120\}'
Code highlighting failed, falling back to plain text:",H),this.isDestroyed)return;this.textBuffer.setText($)...
```

Triggered by `Ctrl+C` during a long session. The warning repeats **once per pending highlight request** — `destroy()` → `rejectPendingRequests()` rejects the entire `messageCallbacks` map in one go, so the longer the transcript, the more copies get dumped. Teardown stack:

```
at destroy (chunk-hfnzvvc7.js:64:11892)          <- TreeSitterClient.destroy
at LZ (chunk-hfnzvvc7.js:64:19205)
at finalizeDestroy (chunk-hfnzvvc7.js:93:22271)
at destroy (chunk-hfnzvvc7.js:93:19087)
at gg (chunk-5haa5ka2.js:94:18044)
at exit (chunk-5haa5ka2.js:94:20824)
...
at processParsedKey (chunk-hfnzvvc7.js:26:9634)
at handleStdinEvent (chunk-hfnzvvc7.js:93:5055)
```

One detail worth noting: it lands on the **raw terminal** rather than the TUI console pane, because `finalizeDestroy()` calls `_console.destroy()` → `restoreOriginalConsole()` in the same synchronous block, while the rejections settle on a later microtask — so by the time the warns fire, `global.console` is already the original one.

The typed-`TreeSitterClientDestroyedError` approach here looks like the right call. Guarding on `isDestroyed` alone wouldn't cover it, since the rejection can land on a `CodeRenderable` that hasn't been destroyed itself.

PR #1323, opened 2026-08-03
fix(core): scrub reflowed split-footer rows on resize

### LINKED ISSUE #1322: split-footer resize leaves reflowed footer fragments in scrollback
### Summary

Narrowing a terminal in `split-footer` mode with `externalOutputMode: "capture-stdout"` can leave reflowed fragments of the previous footer on screen and in scrollback.

### Reproduction

Render three full-width footer rows (`AAA…`, `BBB…`, `CCC…`) in a 24-row tmux window, then resize horizontally:

| Resize | Band lands | Leftover fragments |
|---|---|---|
| 100 -> 70 | rows 8-10 | rows 5-7 |
| 140 -> 70 | rows 8-10 | rows 5-7 |
| 80 -> 120 | rows 8-10 | none |

The leftovers are contiguous, immediately above the band, and appear only when narrowing. Their count is:

`N * (ceil(W_old / W_new) - 1)`

### Root cause

`processResize` starts the width-shrink scrub from:

```ts
previousTerminalHeight - visiblePreviousSplitHeight * 2
```

That assumes the footer is bottom-pinned and the reflow ratio is at most two. Under captured stdout, startup cursor seeding can place the footer immediately after existing shell output instead; here the band is at rows 8-10 of a 24-row terminal. The current scrub therefore clears below both the reflowed fragments and the visible band.

The scrub should instead start from the visible surface top (`pendingSplitFooterTransition?.sourceTopLine ?? renderOffset + 1`) minus the measured reflow expansion.

### Local validation

With that calculation, the probe reports zero fragments at 1.4x, 2x, and 2.8x narrowing and across `140 -> 100 -> 70 -> 50`, while preserving all history sentinels in scrollback. Reverting the patch makes the detector report the expected three fragments again.

jsonMartin:
Fixes #1322.

## Summary

- anchor width-shrink cleanup to the visible split-footer surface top
- subtract the exact reflow expansion for the previous width
- cover cursor-seeded and bottom-pinned surfaces at 1.4x, 2x, and 2.8x narrowing

## Reproduction

A probe rendered three full-width footer rows (`AAA…`, `BBB…`, `CCC…`) in a 24-row tmux window:

| Resize | Band lands | Leftover fragments |
|---|---|---|
| 100 -> 70 | rows 8-10 | rows 5-7 |
| 140 -> 70 | rows 8-10 | rows 5-7 |
| 80 -> 120 | rows 8-10 | none |

The old calculation assumed the footer was bottom-pinned and the reflow ratio was at most two. Captured stdout can instead cursor-seed the surface immediately after existing output, so the scrub started below both the reflowed fragments and the visible band.

After this change, the probe reports zero fragments at 1.4x, 2x, and 2.8x narrowing and across `140 -> 100 -> 70 -> 50`, with all history sentinels intact. Reverting the patch makes the detector report the expected three fragments.

## Tests

- `bun test src/tests/renderer.console-startup.test.ts`: 73 passed
- native suite: 1,859 passed, 6 skipped
- `bun run fmt:check`
- `bun run lint`

Terminal reflow is performed by the emulator, so the headless regression tests assert the clear-row calculation for both surface geometries rather than simulating reflow itself.

The complete Bun suite was also attempted twice in Bun 1.3.3 on Linux arm64. It reproducibly crashed inside Bun after unrelated tests (once while visiting `src/lib/singleton.ts`); no test assertion failed before the runtime crash.

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -345,6 +345,16 @@ function normalizeFooterHeight(footerHeight: number | undefined): number {
   return normalizedFooterHeight
 }
 
+function calculateSplitFooterResizeClearStart(
+  previousSurfaceTop: number,
+  visiblePreviousSplitHeight: number,
+  prevWidth: number,
+  width: number,
+): number {
+  const reflowRows = visiblePreviousSplitHeight * (Math.ceil(prevWidth / width) - 1)
+  return Math.max(previousSurfaceTop - reflowRows, 1)
+}
+
 function resolveModes(config: CliRendererConfig): {
   screenMode: ScreenMode
   footerHeight: number
@@ -3845,7 +3855,6 @@ export class CliRenderer extends EventEmitter implements RenderContext {
       this._footerHeight,
     )
     const prevWidth = this._terminalWidth
-    const previousTerminalHeight = this._terminalHeight
     const visiblePreviousSplitHeight =
       pendingSplitFooterTransition?.sourceHeight ?? previousGeometry.effectiveFooterHeight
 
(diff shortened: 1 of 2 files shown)


jsonMartin:
Autogenerated by Claude by mistake while testing a local solution. Please ignore.

PR #1329, opened 2026-08-04
fix(core/editor-view): snap visual cursor columns to grapheme boundaries

### LINKED ISSUE #1289: Textarea vertical navigation can place the cursor inside a wide CJK grapheme
## Summary

In a `TextareaRenderable`, Up/Down navigation preserves a visual terminal column. If the target line has different wide-character boundaries, the preserved column can fall inside a width-2 CJK grapheme. OpenTUI then renders the terminal cursor between the two cells of that grapheme.

This also affects Shift+Up/Down selection because those actions use the same visual movement path.

## Environment

- `@opentui/core` 0.4.3 (reproduced headlessly and in Windows Terminal)
- 0.4.5 is the current npm release; the relevant mapping is unchanged
- current `main` also appears to retain the same mapping in `packages/core/src/zig/editor-view.zig`
- Bun 1.3.x on WSL2

## Minimal reproduction

Use two lines whose valid display-cell boundaries do not align:

```text
的[代码签名政策](
因此签名批准者角色
```

The first line has legal display-column boundaries `0, 2, 3, 5, 7, ...` because the ASCII `[` shifts the following CJK graphemes by one cell. The second line has boundaries `0, 2, 4, 6, ...`.

```tsx
import type { TextareaRenderable } from "@opentui/core";

let textarea: TextareaRenderable | undefined;
const setup = await testRender(
  () => (
    <textarea
      ref={(value: TextareaRenderable) => { textarea = value; }}
      initialValue={"的[代码签名政策](\n因此签名批准者角色"}
      focused
      width={40}
      height={4}
    />
  ),
  { width: 40, height: 4 },
);

await setup.renderOnce();
const editor = textarea!;

// Put the cursor after `代` on line 1: visual row 0, display column 5.
editor.setCursor(0, 3);
editor.moveCursorDown();
await setup.renderOnce();

console.log(editor.editorView.getVisualCursor());
// Actual: visual row 1, visual column 5.
// Column 5 is inside `签`, whose display cells span columns 4..6.

setup.renderer.destroy();
```

The same problem can be seen interactively by placing the cursor after `代` on the first line and pressing Down once.

## Expected vs actual

- **Expected:** vertical navigation snaps to a valid grapheme/display-cell boundary on the target line (column 4 or 6, according to the intended editor policy).
- **Actual:** the desired visual column, 5, is committed directly. The terminal cursor is drawn between the two cells occupied by `签`.

Text insertion from this invalid state did not split the UTF-8 sequence in my test; it snapped before the affected glyph. However, cursor placement is visibly wrong, and Shift+Up/Down can produce an incorrect selection endpoint.

## Suspected area

`TextareaRenderable.moveCursorDown()` delegates to `editorView.moveDownVisual()`. In `packages/core/src/zig/editor-view.zig`, vertical movement preserves `desired_visual_col`, while `visualToLogicalCursor()` derives the target with logic equivalent to:

```zig
logical_col = vline.source_col_offset + clamped_visual_col;
```

That target is not snapped to a valid grapheme/display-cell boundary before the cursor is committed. Rendering subsequently places the terminal cursor at `visualCursor.visualCol`, exposing the interior column.

A fix likely needs to resolve the requested visual column to a legal boundary on the target visual line before updating the logical cursor. The same correction should cover Up/Down, Shift+Up/Down, and soft-wrapped visual lines.

3aKHP:
Fixes #1289

## Rebase note (2026-08-27)

Rebased onto current `main` across the #1391 source move (`packages/core/src/zig` → `packages/native/src`) and the #1428 wrapped-layout rewrite (`VirtualLine.source_col_offset` → `source_col_start`; the unsnapped arithmetic this PR targets is unchanged, now at `editor-view.zig:655`). `getVisualEOL` keeps the newer viewport/occupancy computation from #1393 and only routes its final visual→logical mapping through the snapped path. One test expectation was aligned with #1393 semantics: under `cell` occupancy a forward inclusive selection stores the *exclusive* end, so the assertion is `sel.end == cursor.offset + width(签)` rather than `sel.end == cursor.offset`.

Sensitivity on this base: without the fix, 8 of the 10 new tests fail; the two selection-extension tests pass because #1393's occupancy machinery already snaps the selection-sync path — the direct-move path (`moveUpVisual`/`moveDownVisual`) remains broken. With the fix, all 10 pass.

## Summary

Vertical cursor movement (`moveUpVisual`/`moveDownVisual`) preserves a desired visual column measured in display cells. When the target row's grapheme boundaries differ from the row the column was measured on — e.g. width-2 CJK graphemes shifted by an odd-width ASCII char — the desired column can fall strictly inside a grapheme. `visualToLogicalCursor` committed it verbatim:

```zig
const clamped_visual_col = @min(visual_col, vline.width_cols);
const logical_col = vline.source_col_start + clamped_visual_col;
```

The terminal caret then renders between the two cells of a wide glyph, and Shift+Up/Down (same code path) can produce a selection endpoint inside a grapheme. All column/offset units here are display-cell widths (`coordsToOffset` = `line_start_weight + col`), so an interior column is not a legal cursor position — it conflates display cells with grapheme boundaries.

Reproduction (headless) is in #1289: buffer `的[代码签名政策](` / `因此签名批准者角色`, cursor after `代` (visual col 5), move down → col 5 lands inside `签` (cells 4..6).

## Fix

Snap the requested visual column to the **leading (line-start-side) boundary** of the containing grapheme cluster before committing the cursor:

- New iterator helper `snapColToGraphemeBoundary` (`text-buffer-iterators.zig`), placed next to `getGraphemeWidthAt`/`getPrevGraphemeWidth` whose segment-walk pattern it mirrors. It reuses `utf8.findPosByWidth(..., include_start_before=false, width_method)`, so it is grapheme-cluster-aware (combining marks, emoji/ZWJ) and width-method-aware; no new width tables. One subtlety: `findPosByWidth`'s final-cluster tail path reports the *trailing* boundary when the column sits inside the last cluster of a chunk — the helper detects `columns_used > local_col` and steps back one cluster width via `utf8.getPrevGraphemeStart`.
- `visualToLogicalCursor` snaps after clamping (central fix; covers Up/Down, Shift+Up/Down, and mouse/TS callers).
- `getVisualEOL` and `makeCursorVisible` had the same unsnapped col arithmetic; both now route through the snapped mapping.

Leading-boundary policy is deterministic, keeps the caret on the grapheme it was aimed at, and composes with `clampVisualColToStayOnVisualRow` (#1078), whose one-cell step back from a wrap boundary can itself land inside a trailing wide grapheme — leading snap resolves that back onto the same visual row instead of drifting onto the next row's start.

Legal boundaries, column 0, end-of-line, and ASCII-only text are identity mappings (verified by the unchanged pre-existing suite).

## Tests

10 new tests in `packages/native/src/tests/editor-view_test.zig`:

- exact issue repro: move down onto interior column snaps to `签`'s leading boundary (col 4), plus the direct `visualToLogicalCursor(1, 5)` mapping
- move up with the same boundary mismatch
- Shift+Up/Down selection extension ends on a legal grapheme boundary (mirrors the TS path: `moveCursorUp/Down` → `getVisualCursor` → `setLocalSelection`/`updateLocalSelection` → `syncCursorToSelectionFocus`)
- vertical moves between soft-wrapped visual rows of one logical line never split wide graphemes
- short and empty target rows: clamping without splitting, desired-col persistence, round trip
- combining-mark cluster and emoji/ZWJ cluster boundaries
- down-down-up-up round trip returns to the original column without drift
- inserting at the snapped cursor keeps graphemes intact (full-text equality)

`zig build test -Dtest-filter="EditorView"`: **118/118 passed**. Full `zig build test`: 2099/2108 (8 skipped; the one failure is an environment-dependent X11 clipboard test that fails identically on pristine `main`). `zig fmt` clean. TS: `bun test src/editor-view.test.ts` **73/73** against the rebuilt native library (including the aligned emoji vertical-nav expectation from the companion commit: col 8 sits inside 🌟 cells 7–8, the snapped leading boundary is 7); `edit-buffer`/`buffer`/`text-buffer` TS suites: 218 passed.

## Out of scope (deliberately; happy to follow up)

- **Mouse selection**: `coordsToCharOffset` (`text-buffer-view.zig`) can commit a raw interior-cell offset into a selection, and `syncCursorToSelectionFocus` (`editor-view.zig`) can then put the cursor there when clicking the second cell of a wide grapheme. Not changed here: snapping in `coordsToCharOffset` would alter selection-range semantics (selection offsets deliberately snap at extraction time), and snapping only the cursor would decouple `cursor.offset` from the selection focus offset the TS layer relies on. Shift+arrow selection **is** covered by this PR because it anchors from the already-snapped committed cursor.
- **Logical-line moves**: `EditBuffer.moveUp`/`moveDown` (`edit-buffer.zig:515,538`) contain the same `@min(desired_col, line_width)` unsnapped arithmetic. `EditBufferRenderable` uses `moveUpVisual`/`moveDownVisual`, so the reported defect path is unaffected; the same one-line helper applies if you want a follow-up.

### DIFF
--- a/packages/core/src/editor-view.test.ts
+++ b/packages/core/src/editor-view.test.ts
@@ -523,7 +523,8 @@ describe("EditorView", () => {
       view.moveDownVisual()
       cursor = view.getVisualCursor()
       expect(cursor.visualRow).toBe(1)
-      expect(cursor.visualCol).toBe(8)
+      // Col 8 sits inside 🌟 (cells 7-8); vertical movement snaps to the leading boundary.
+      expect(cursor.visualCol).toBe(7)
 
       buffer.moveCursorLeft()
       cursor = view.getVisualCursor()
--- a/packages/native/src/editor-view.zig
+++ b/packages/native/src/editor-view.zig
@@ -178,7 +178,16 @@ pub const EditorView = struct {
                 const target_logical_row = @as(u32, @intCast(target_vline.source_line));
 
                 const line_width = iter_mod.lineWidthAt(self.edit_buffer.tb.rope(), target_logical_row);
-                const target_col = @min(cursor.col, line_width);
+                // The cursor col is a display width measured on the source line;
+                // snap it to a grapheme boundary on the target line.
+                const target_col = iter_mod.snapColToGraphemeBoundary(
+                    self.edit_buffer.tb.rope(),
+                    self.edit_buffer.tb.memRegistry(),
+                    target_logical_row,
+                    @min(cursor.col, line_width),
+                    self.edit_buffer.tb.tabWidth(),
+                    self.edit_buffer.tb.widthMethod(),
+                );
 
                 if (self.edit_buffer.cursors.items.len > 0) {
                     const offset = iter_mod.coordsToOffset(self.edit_buffer.tb.rope(), target_logical_row, target_col) orelse return;
@@ -653,14 +662,29 @@ pub const EditorView = struct {
 
         const vline = &vlines[visual_row];
         const clamped_visual_col = @min(visual_col, vline.width_cols);
-        const logical_col = vline.source_col_start + clamped_visual_col;
         const logical_row = @as(u32, @intCast(vline.source_line));
+
+        // A visual column can fall strictly inside a wide grapheme when the
+        // target row's grapheme boundaries differ from the row the column was
+        // measured on. Snap to the cluster's leading boundary so the committed
+        // cursor never splits a grapheme. This also composes with
+        // clampVisualColToStayOnVisualRow, whose one-cell step back from a wrap
+        // boundary can land inside a trailing wide grapheme.
+        const logical_col = iter_mod.snapColToGraphemeBoundary(
+            self.edit_buffer.tb.rope(),
+            self.edit_buffer.tb.memRegistry(),
+            logical_row,
+            vline.source_col_start + clamped_visual_col,
+            self.edit_buffer.tb.tabWidth(),
+            self.edit_buffer.tb.widthMethod(),
+        );
+        const snapped_visual_col = logical_col -| vline.source_col_start;
 
         const offset = iter_mod.coordsToOffset(self.edit_buffer.tb.rope(), logical_row, logical_col) orelse 0;
 
         return .{
             .visual_row = visual_row,
-            .visual_col = clamped_visual_col,
+            .visual_col = snapped_visual_col,
             .logical_row = logical_row,
             .logical_col = logical_col,
             .offset = offset,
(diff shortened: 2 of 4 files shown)


simonklee:
`VESICLE_FORK.md` ..

3aKHP:
> `VESICLE_FORK.md` ..

Sorry for my carelessness — that file is internal to our fork and should not have been part of this PR. My mistake. The branch has been rebased onto the pristine upstream tag, and the diff now contains only the intended Zig changes. Thank you for the time you spent looking at it.

3aKHP:
Rebased onto current main (53291770, including the Zig 0.16 migration) without conflicts. The internal fork file noted earlier remains absent from the diff. This is ready for review when you have time; thank you.

PR #1330, opened 2026-08-04
fix(core/text-buffer-view): fill partially wrapped lines across chunk boundaries

### LINKED ISSUE #1288: Textarea: inserting into a soft-wrapped line anchors the wrap boundary at the edit offset
## Summary

With `wrapMode="word"`, inserting a character into the middle of the first visual row of an already soft-wrapped long line moves the first wrap boundary to the insertion point. The stale wrap persists for that logical line and only a full `setText`/`replaceText` recomputes it.

## Environment

- `@opentui/core` 0.4.3 **and** 0.4.5 (both reproduce)
- `@opentui/solid` same versions, also reproducible in a real terminal (Bun 1.3.14, Linux)

## Minimal reproduction (headless)

```tsx
import type { TextareaRenderable } from "@opentui/core";

let ed: TextareaRenderable | undefined;
const setup = await testRender(
  () => (
    <box width="100%" height="100%">
      <line_number fg="#555555" minWidth={4} paddingRight={1}>
        <textarea
          ref={(r: TextareaRenderable) => { ed = r; }}
          initialValue={"a".repeat(100)}
          focused
          width="100%"
          height="100%"
          wrapMode="word"
        />
      </line_number>
    </box>
  ),
  { width: 60, height: 15 },
);
await setup.renderOnce();
const ta = ed!;
console.log("before:", ta.editorView.getTotalVirtualLineCount()); // 2 — first row full width

ta.setCursor(0, 20);               // middle of the first visual row
await setup.mockInput.typeText("X");
await setup.renderOnce();
console.log("after:", ta.editorView.getTotalVirtualLineCount());  // 3 — BUG
console.log(setup.captureCharFrame());
// Row 1 now ends right after the inserted X (col 21) instead of filling
// the 56-col viewport; the remaining text wraps full-width below it.
setup.renderer.destroy();
```

## Expected vs actual

- **Expected:** the logical line (101 chars) re-wraps completely — the first visual row stays full-width, virtual line count stays 2.
- **Actual:** virtual line count becomes 3; the first visual row ends exactly at the edit offset. Every subsequent edit on that line re-anchors the boundary at its own edit point. In a real terminal the first visual row of the edited logical line is permanently truncated at the cursor.

## No viable consumer-side workaround

Nothing in the public API triggers a full re-wrap of existing text:

- `editorView.setViewportSize` (same size, or wiggled ±1 and back) — no effect
- `editorView.setWrapMode` toggle, renderable-level `wrapMode` toggle — no effect
- `editorView.setViewport(...)` — no effect
- `editBuffer.setText` — recomputes, but clears the undo stack
- `editBuffer.replaceText` — recomputes, but resets the cursor to (0,0) and pushes an undo entry; worse, the stale wrap cache seems bound to undo state, so `replaceText` followed immediately by `undo()` (a zero-cost repair attempt) restores the broken wrap again

## Suspected area

The incremental wrap recompute in the native edit buffer appears to start from the edit offset and keep it as a segment boundary, instead of re-wrapping the whole logical line (or re-flowing from the line start). Happy to test a fix against the repro above.

3aKHP:
Fixes #1288

## Summary

With `wrapMode="word"`, inserting into the middle of an already soft-wrapped logical line anchors the first wrap boundary at the edit offset. The affected row stays truncated — e.g. `initialValue = "a" * 100` at wrap width 56, insert `"X"` at col 20: the virtual line count becomes 3 with row 1 ending at col 21, where a fresh `setText` of the same 101 characters yields 2 full-width rows. Reproduction is in #1288.

## Root cause

This is not stale caching — all wrapping is recomputed from scratch per update by `calculateVirtualLinesGeneric`. It is deterministic chunk-segmentation divergence:

- Rope edits split chunks at the edit offset: after the repro edit the line is chunks `["a"*20]["X"]["a"*80]` instead of one chunk.
- In the word-mode walker (`text-buffer-view.zig`), when a chunk arrives at a partially filled line and no wrap break has been recorded on that line (`remaining_in_chunk > remaining_on_line`, no `last_wrap_that_fits`, `last_wrap_chunk_count == 0`), the final else branch called `commitVirtualLine()` unconditionally — treating the chunk boundary as a wrap point — then filled the *next* line at full width.
- A single-chunk buffer with identical text takes the `line_position == 0` hard-fill branch instead and wraps correctly. Hence "only broken after editing".

## Fix

One hunk in the else branch: a chunk boundary is not a wrap point. Hard-fill the remaining space on the current line from the chunk via `utf8.findWrapPosByWidth(remaining_bytes, remaining_on_line, ...)`, mirroring the `line_position == 0` path. Commit only when:

- the line is already full (`remaining_on_line == 0`), or
- the next grapheme does not fit the remaining space (`columns_used == 0`, e.g. 1 column left and a width-2 CJK grapheme next) — then the short row commits and the next iteration hard-fills a fresh line. The force-one-grapheme guard is kept only for the full-line case, exactly as before; applying it mid-line would have split wide graphemes across rows and changed single-chunk behavior too (single-chunk walks also reach this branch after the first hard-filled row).

All other branches — fit, `last_wrap_that_fits`, `line_position == 0`, and the `last_wrap_chunk_count > 0` backtrack (#1078 machinery) — are byte-identical. This branch only runs when no wrap break was recorded on the current line, so word-wrap backtracking is unaffected.

**Intended behavior change:** an unbreakable run arriving at a partially filled line no longer pushes to the next row. Example: chunks `["hello"]["x"*100]` at wrap 56 (e.g. insert `"hello"` at offset 0 of an x-run) — old: rows `[5, 56, 44]`; new: `[56, 49]`, identical to a fresh `setText` of the same content. Edge behavior preserved: a wide grapheme that doesn't fit the remaining space still wraps whole (`["a"*54]["世"*20]` at wrap 55 → `[54, 40]`, matching single-chunk behavior before and after; pinned by test).

## Tests

11 new tests in `packages/core/src/zig/tests/word-wrap-editing_test.zig` (file already registered in `test.zig`). The central oracle is `expectSegmentationMatchesFreshSet`: read the edited buffer's full text, `setText` it into a fresh buffer + view with the same wrap mode/width, and assert identical virtual-line segmentation (count, `source_line`, `source_col_offset`, `width_cols`) — segmentation must depend only on text content, never on edit history.

- exact issue repro: insert into soft-wrapped unbreakable line reflows (`expected 2, found 3` pre-fix)
- insert / delete / paste-like multi-char insert at several offsets (line start, mid-first-row, row boundary, mid-second-row), oracle-checked
- word boundaries still backtrack after a mid-row edit (spaces; explicit widths + oracle)
- unbreakable CJK line edited mid-row (asserts even row widths = no grapheme split)
- mixed widths with combining marks, oracle at 4 offsets
- five sequential single-char inserts, oracle after each
- undo → redo restores fresh-set segmentation
- wrap width change after edit (56→40→56), oracle-checked
- wide grapheme that does not fit the remaining space is not split (guard for the new branch)

Each was run individually and fails pre-fix except the two behavior guards. `zig build test`: **1877 passed, 6 skipped, 0 failed**. `zig fmt` clean. `wrapMode="char"`/`"none"` untouched — the change is entirely inside the word-mode arm and their existing tests pass unchanged.

## Performance

No asymptotic change: per loop iteration the work is identical — one `findWrapPosByWidth` scan over the same bytes, only the width argument differs, and the tracked `byte_offset` invariant (the O(n²) guard) holds. A commit is *removed* from the common path; the new commit-and-continue is bounded by the number of virtual lines. `zig build bench` ("TextBuffer Wrapping", word wrap avg, before → after): multi-line w40 27.69→18.60 ms, w80 21.52→15.00 ms, w120 17.44→13.19 ms; single-line w40 28.29→25.18 ms, w80 27.74→26.00 ms, w120 26.34→27.22 ms — no regression (differences are machine noise in our favor).

### DIFF
--- a/packages/core/src/zig/tests/word-wrap-editing_test.zig
+++ b/packages/core/src/zig/tests/word-wrap-editing_test.zig
@@ -496,3 +496,378 @@ test "Word wrap - incremental character edits near boundary" {
     try std.testing.expectEqual(@as(u32, 14), vlines[0].width_cols);
     try std.testing.expectEqual(@as(u32, 6), vlines[1].width_cols);
 }
+
+/// Assert that the virtual-line segmentation of `view` (backed by an edited,
+/// potentially multi-chunk buffer) is identical to the segmentation of a fresh
+/// single-chunk buffer set to the same text content. Wrap segmentation must
+/// depend only on text content, never on how the text was edited.
+fn expectSegmentationMatchesFreshSet(
+    pool: *gp.GraphemePool,
+    link_pool: *link.LinkPool,
+    eb: *EditBuffer,
+    view: *TextBufferView,
+    wrap_width: u32,
+) !void {
+    var text_buf: [16384]u8 = undefined;
+    const text_len = eb.getText(&text_buf);
+
+    var fresh_eb = try EditBuffer.init(std.testing.allocator, pool, link_pool, .wcwidth, null);
+    defer fresh_eb.deinit();
+
+    var fresh_view = try TextBufferView.init(std.testing.allocator, fresh_eb.getTextBuffer());
+    defer fresh_view.deinit();
+
+    fresh_view.setWrapMode(.word);
+    fresh_view.setWrapWidth(wrap_width);
+    try fresh_eb.setText(text_buf[0..text_len]);
+
+    const actual = view.getVirtualLines();
+    const expected = fresh_view.getVirtualLines();
+
+    try std.testing.expectEqual(expected.len, actual.len);
+    for (expected, actual) |exp, act| {
+        try std.testing.expectEqual(exp.source_line, act.source_line);
+        try std.testing.expectEqual(exp.source_col_offset, act.source_col_offset);
+        try std.testing.expectEqual(exp.width_cols, act.width_cols);
+    }
+}
+
+test "Word wrap - insert into soft-wrapped unbreakable line reflows (issue 1288)" {
+    const pool = gp.initGlobalPool(std.testing.allocator);
+    defer gp.deinitGlobalPool();
+    const link_pool = link.initGlobalLinkPool(std.testing.allocator);
+    defer link.deinitGlobalLinkPool();
+
+    var eb = try EditBuffer.init(std.testing.allocator, pool, link_pool, .wcwidth, null);
+    defer eb.deinit();
+
+    var view = try TextBufferView.init(std.testing.allocator, eb.getTextBuffer());
+    defer view.deinit();
+
+    view.setWrapMode(.word);
+    view.setWrapWidth(56);
+
+    try eb.setText("a" ** 100);
+
+    var vlines = view.getVirtualLines();
+    try std.testing.expectEqual(@as(usize, 2), vlines.len);
+
+    try eb.setCursor(0, 20);
+    try eb.insertText("X");
+
+    vlines = view.getVirtualLines();
+    try std.testing.expectEqual(@as(usize, 2), vlines.len);
+    try std.testing.expectEqual(@as(u32, 56), vlines[0].width_cols);
+    try std.testing.expectEqual(@as(u32, 45), vlines[1].width_cols);
+
+    try expectSegmentationMatchesFreshSet(pool, link_pool, eb, view, 56);
+}
+
+test "Word wrap - insert at various offsets matches fresh set" {
+    const pool = gp.initGlobalPool(std.testing.allocator);
+    defer gp.deinitGlobalPool();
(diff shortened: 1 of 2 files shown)


3aKHP:
Rebased onto current main (53291770, including the Zig 0.16 migration) without conflicts. This is ready for review when you have time; thank you.

simonklee:
Thank you for this contribution. We are closing this pull request because #1428
replaced the affected word-wrap implementation and added a test for words split
across text chunks. The files changed here no longer exist on main.

PR #1345, opened 2026-08-06
feat: allow passing marked extensions to markdown lexer

Cl-VII:
## Summary

Allow callers to pass marked extensions to the markdown lexer, so custom inline/block tokenizers (e.g. LaTeX math) can be applied during incremental parsing.

## Motivation

The OpenCode TUI renders markdown via MarkdownRenderable. To render LaTeX math as Unicode, a marked extension ($...$ tokenizer) must reach the Lexer.lex call inside parseMarkdownIncremental. Currently the lexer is always invoked without options, so extensions registered via marked.use() are ignored (marked 17+ drops global extensions when explicit options are passed) and there is no way to inject them.

## Changes

- packages/core/src/renderables/markdown-parser.ts
  - parseMarkdownIncremental accepts an optional extensions: MarkedExtension[] parameter
  - New normalizeExtensions() converts MarkedExtension[] into the shape the marked Lexer expects (tokenizers under extensions.inline/extensions.block, start fns under startInline/startBlock) - same logic as marked.use
  - Lexer.lex is called with { gfm: true, extensions } when extensions are provided
- packages/core/src/renderables/Markdown.ts
  - MarkdownOptions gains extensions?: MarkedExtension[], stored and forwarded to the incremental parser
- packages/core/src/renderables/__tests__/markdown-parser.test.ts
  - New test verifying an inline math extension produces a custom token through the incremental parser
  - New test verifying no-extensions path still uses gfm defaults

## Compatibility

- Without extensions the behavior is unchanged (Lexer.lex(src) with defaults, gfm on).
- Type-only addition; the native renderer is untouched.

### DIFF
--- a/packages/core/src/renderables/Markdown.ts
+++ b/packages/core/src/renderables/Markdown.ts
@@ -5,7 +5,7 @@ import type { TextChunk } from "../text-buffer.js"
 import { createTextAttributes } from "../utils.js"
 import type { BorderStyle } from "../lib/border.js"
 import { RGBA, parseColor, type ColorInput } from "../lib/RGBA.js"
-import { Lexer, type MarkedToken, type Token, type Tokens } from "marked"
+import { Lexer, type MarkedExtension, type MarkedToken, type Token, type Tokens } from "marked"
 import { CodeRenderable, type OnChunksCallback } from "./Code.js"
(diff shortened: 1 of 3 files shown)


Cl-VII:
Author requested removal of all latex-related uploads.

PR #1346, opened 2026-08-07
feat(core): wrapIndent for soft-wrap continuations

Hyuuh:
## Summary
- Add `wrapIndent` (`none` | `same`, default `none`) so soft-wrap continuations keep the logical line's leading indent (VS Code `wrappingIndent: "same"`).
- Zig layout stores `VirtualLine.pad_cols` and narrows continuation wrap width; paint offsets content and hit-testing/cursor stay consistent with painted geometry.
- Wired through FFI → `TextBufferView` / `EditorView` → `TextBufferRenderable` / `EditBufferRenderable` (`wrapIndent` prop). No breaking change when omitted.

Note: orthogonal to #1330 (chunk-boundary reflow after edits).

## Test plan
- [x] `bun run test:native -Dtest-filter="wrap indent"` (13/13)
- [x] Related filters: wrapping / word wrap / moveDownVisual / occupancy
- [x] TS smoke: Textarea `wrapIndent` get/set
- [ ] Manual: indented line + `wrapMode: "word"|"char"` + `wrapIndent: "same"` — continuations paint under indent; click pad → caret at content start; ↓ from col 0 → content start of continuation


simonklee:
Just a few notes from grok llm looking through the PR:

If you omit wrapIndent, the default stays none and nothing changes. wrapIndent "same" breaks Up and Down.

getVisualCursor adds pad_cols so the column matches paint. moveUpVisual and moveDownVisual still store desired_visual_col as offset minus document_cell_offset. They pass that content column to visualToLogicalCursor.
Continuation text paints under the first-line text. As a result, the caret moves right by the indent.

Repro: indent of 4 spaces, wrap width 20, a long run of a. Put the caret on logical column 4 (the first a, painted column 4). Down puts the caret on painted column 8.

renderCursor and shift-selection both read getVisualCursor().visualCol.

The wrap-indent movement test only does Down from visual column 0. At column 0, content-sticky and paint-sticky with clamp-out-of-pad give the same result. The older test "moveDownVisual across empty line preserves desired column" already uses public visual_col as the sticky X.

Keep internal translation content-relative if you want. Convert at the movement boundary:

- desired_paint = content + src.pad
- target_content = desired_paint -| dst.pad

Then clamp the caret out of the pad.

Add a native test that fails on this head. Down from the first non-indent cell must keep public visual_col.

Hyuuh:
Hi, @simonklee! I appreciate you taking the time to review my PR. I'll get right on checking what you've mentioned.

PR #1363, opened 2026-08-13
fix(core): refresh dimensions after resize debounce

### LINKED ISSUE #1344: SIGWINCH-based window resize silently breaks when process.stdout is not the user's real terminal (e.g. wrapped by 1Password's op run)
## Title

SIGWINCH-based window resize silently breaks when `process.stdout` is not the user's real terminal (e.g. wrapped by 1Password's `op run`, screen recorders, or any stdio-proxying wrapper)

## Summary

`CliRenderer` only registers its `SIGWINCH` listener when `_usesProcessStdout` is true (`renderer.ts:1205`). `_usesProcessStdout` is set to `stdout === process.stdout` (line 1039), which is identity, but the documented rationale at line 985-986 is "terminal-driven resize only fires for `process.stdout`."

When the *child process* is launched by a wrapper that proxies stdio — most commonly 1Password CLI's `op run`, which **always** pipes the child's stdout/stderr (even with `--no-masking`) — the renderer's stdout is no longer attached to the user's terminal. In that case:

1. `process.on("SIGWINCH", …)` is gated by `_usesProcessStdout`. **However**, the gate is `stdout === process.stdout`, which is *true* even when stdout is a pipe. So the listener *is* registered in the wrapper case.
2. On every SIGWINCH, the handler reads `this.stdout.columns` / `this.stdout.rows` (`renderer.ts:891-892`). On a pipe (non-TTY) those are `undefined`, so the handler falls back to `80×24` and calls `handleResize(80, 24)`. The renderer resizes to a tiny hardcoded grid on every window resize — which from the user's perspective looks like "the TUI doesn't resize when I resize the window" (layout collapses to ~80×24, sometimes interpreted as "stuck" if the original was wider).
3. The initial size at startup is *correct* because `CliRenderer` is constructed with explicit `width`/`height` from the caller (line 1046). Only subsequent resizes are wrong.

The net effect: **no TUI using `CliRenderer` correctly resizes when launched through any wrapper that proxies stdio** — `op run`, `script -f` (script), `asciinema`, screen recorders, etc.

## Reproduction (opencode 1.18.x on macOS, Bun 1.3.14 standalone)

Minimal: launch `~/.opencode/bin/opencode` (any modern opencode TUI which uses OpenTUI) via 1Password's `op run`:

```sh
op run --env-file ~/.config/opencode/qwen.env -- ~/.opencode/bin/opencode
```

Inside the running TUI, resize the terminal window: the layout does **not** reflow. Manually sending `kill -WINCH <opencode-pid>` produces the same effect (no real resize).

Confirm the cause:

```sh
lsof -a -d 1,2 -p <opencode-pid>    # fd 1,2 are PIPE under op run, not the TTY
stty -f /dev/ttysNNN size            # PTY size is updated correctly by the host
```

Other TUI apps launched the same way (`htop`, `vim`) resize fine in the same terminal, so the kernel signal path works — the bug is specific to OpenTUI's resolver.

## Expected

The TUI should detect window resizes and reflow correctly regardless of whether the parent process proxied stdio, as long as at least one of stdin/stdout is the user's TTY (typically stdin remains a TTY under `op run`).

## Proposed fix

In `packages/core/src/renderer.ts`:

1. When `this.stdout.isTTY` is false, fall back to reading the TTY size from `process.stdin` (which is usually still a TTY under wrappers like `op run`). `process.stdin.rows` / `columns` reflect the user's terminal.

2. Either:
   - register `SIGWINCH` whenever any of `process.stdout.isTTY` or `process.stdin.isTTY` is true, or
   - keep the existing `_usesProcessStdout` identity gate (for custom Writable cases) but in the handler, fall back to `process.stdin.columns/rows` when `this.stdout` columns/rows are `undefined`.

3. Keep the current guard for non-process custom Writables (tests, embedded use), where SIGWINCH is genuinely meaningless.

A small change, likely under 10 lines, and would unblock every Bun-OpenTUI app run through `op run` (or any stdio proxy) — an increasingly common deployment shape for credential-bearing dev tools.

## Environment

- OpenTUI: latest (`anomalyco/opentui`, checked out 2026-08-05)
- opencode: 1.18.13 (Bun 1.3.14 standalone, macOS aarch64, Ghostty/cmux)
- Repro: 1Password CLI `op run` (any modern version)

## Related

- opencode issue #3328 (closed) — same symptom in the prior opentui incarnation, root-caused at the time to an upstream Bun SIGWINCH delivery bug. The window-resize plumbing has since been rewritten in the new TS renderer, and this stub remains.
- opentui #1187 (renderer diff desync), #1110 (Terminal.app stale columns) — adjacent stale-repaint issues, possibly the same family of bugs once `handleResize` is producing bad sizes.

dajiaohuang:
## Summary

- refresh cached TTY dimensions when a debounced `SIGWINCH` is handled
- sample the refreshed dimensions when the debounce expires
- use `stdin` as the size source when it is the TTY and `stdout` is piped
- cover stale cached dimensions and piped-stdout setups with regression tests

Browser-hosted PTY bridges can deliver a resize signal before the runtime's cached `stdout.columns` and `stdout.rows` reflect the new kernel window size. The existing resize path captured those cached values immediately, so initial attach and shrink events could keep rendering at the previous width.

OpenTUI also always read dimensions from `stdout`, but some launchers pipe stdout while keeping stdin attached to the controlling terminal. In that setup the resize signal arrives, yet stdout has no terminal dimensions. The resize handler now refreshes and samples the active TTY stream after the debounce expires.

This keeps the existing debounce behavior and is related to anomalyco/opencode#42225.

Closes #1344

## Testing

- `bun run test:js` (5,397 passed, 23 skipped)
- `bun run test:js:node` with Node.js 26.4.0 (4,664 passed, 6 skipped)
- `bun run build:lib`
- `bun run fmt:check`
- `bun run lint`
- Linux PTY reproduction covering initial attach, grow, and shrink

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -3909,11 +3909,18 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     if (this._isDestroyed) return
     const resize = () => {
       const stdout = this.stdout as NodeJS.WriteStream & { _refreshSize?: () => void }
+      const stdin = this.stdin as NodeJS.ReadStream & {
+        columns?: number
+        rows?: number
+        isTTY?: boolean
+        _refreshSize?: () => void
+      }
+      const sizeSource = !stdout.isTTY && stdin.isTTY ? stdin : stdout
       // PTY bridges can deliver SIGWINCH before the stream's cached dimensions reflect the new window size.
-      stdout._refreshSize?.()
-      const width = stdout.columns
-      const height = stdout.rows
-      if (width > 0 && height > 0) this.processResize(width, height)
+      sizeSource._refreshSize?.()
+      const width = sizeSource.columns
+      const height = sizeSource.rows
+      if (width && width > 0 && height && height > 0) this.processResize(width, height)
     }
     if (this._splitHeight > 0) {
       resize()
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -916,9 +916,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
 
   private _useConsole: boolean = true
   private sigwinchHandler: () => void = (() => {
-    const width = this.stdout.columns
-    const height = this.stdout.rows
-    if (width > 0 && height > 0) this.handleResize(width, height)
+    this.handleResize()
   }).bind(this)
   private _capabilities: TerminalCapabilities | null = null
   private _latestPointer: { x: number; y: number } = { x: 0, y: 0 }
@@ -3907,10 +3905,18 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     }
   }
 
-  private handleResize(width: number, height: number): void {
+  private handleResize(): void {
     if (this._isDestroyed) return
+    const resize = () => {
+      const stdout = this.stdout as NodeJS.WriteStream & { _refreshSize?: () => void }
+      // PTY bridges can deliver SIGWINCH before the stream's cached dimensions reflect the new window size.
+      stdout._refreshSize?.()
+      const width = stdout.columns
+      const height = stdout.rows
+      if (width > 0 && height > 0) this.processResize(width, height)
+    }
     if (this._splitHeight > 0) {
-      this.processResize(width, height)
+      resize()
       return
     }
 
@@ -3921,7 +3927,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
 
     this.resizeTimeoutId = this.clock.setTimeout(() => {
       this.resizeTimeoutId = null
-      this.processResize(width, height)
+      resize()
     }, this.resizeDebounceDelay)
   }
 
--- a/packages/core/src/tests/renderer.clock.test.ts
+++ b/packages/core/src/tests/renderer.clock.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, beforeEach, expect, test } from "bun:test"
 import { SystemClock } from "../lib/clock.js"
 import { TextRenderable } from "../renderables/Text.js"
+import { CliRenderEvents } from "../renderer.js"
 import { createTestRenderer, type TestRenderer } from "../testing/test-renderer.js"
 import { ManualClock } from "../testing/manual-clock.js"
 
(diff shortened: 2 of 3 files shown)


dajiaohuang:
Rebased onto current main and kept the upstream invalid-dimension guard after refreshing the chosen TTY stream (28c0435). Added a regression for zero dimensions. Both resize/console suites pass: 87 tests; core typecheck and library build pass.

The full core suite initially reported 5,579 passing tests, 36 skipped and one missing generated parser.worker.js asset. After restoring the local bunx launcher and running the library build, the previously failing Node asset test file passes (2/2). No source change was needed for that setup failure.

AmethystLiang:
Another data point from an Electron + xterm.js host, in case it helps this land.

**Environment:** Orca (Electron, xterm.js terminal over a local PTY), macOS 15 / Apple Silicon, opencode 1.18.31 (current latest). Users hit this constantly, because switching worktrees/tabs in the IDE changes the pane geometry while the TUI is in the background.

**What I measured.** I drove a live opencode session over CDP and resized only the PTY, holding the visible xterm grid fixed, to isolate the two directions:

| PTY change | result |
|---|---|
| 212 → 180 cols (shrink) | opencode re-laid out at 180, but cells from the previous 212-col frame stayed painted past the new right edge (visible fragments: `ord reply: ok`, `okens`, `de 1.18.31`) |
| 180 → 212 / any grow | full repaint, screen clean every time |

The recovery asymmetry matches what #42225 describes: growing is the only thing that reliably produces a complete frame. Every workaround our users have found (drag a panel wider, change tab zoom) is a grow in disguise.

**One question about the scope of this fix.** Refreshing the cached dimensions before sampling should fix the case where the re-layout happens at the *wrong* width. It is less clear to me whether it also covers the case where the re-layout happens at the *correct, smaller* width but the region that is no longer part of the grid is never cleared. If OpenTUI's damage tracking only repaints cells inside the current grid, a shrink would still leave the out-of-range cells from the larger previous frame on screen. If that is already handled elsewhere, ignore this; if not, a full clear on shrink may be needed alongside the dimension refresh.

Happy to test a build against our host and report back. We have a reproducible harness on this side.

dajiaohuang:
Thanks — that distinction is valid. I checked the current head: `processResize` now re-reads the selected TTY before re-layout, and the native `resize()` clears/reinitializes the logical buffers so every cell inside the new viewport is repainted. It does not, however, issue an explicit terminal scrub for cells that were outside a narrower viewport.

That matches the outcome of #1240: forcing a full logical repaint was tested and the artifact still reproduced, so a remaining shrink-only symptom would need a separate physical-terminal scrub rather than another buffer repaint. Please try this head with your harness; if the out-of-range fragments remain, the next focused change should target the ANSI scrub on width shrink. I have not changed this PR speculatively.

PR #1379, opened 2026-08-18
fix(core): serialize tree-sitter parse() to avoid Asyncify re-entrancy hang

enkunkun:
Downstream: [anomalyco/opencode#42896](https://github.com/anomalyco/opencode/issues/42896)

## What

`parser.parse()` in web-tree-sitter is Asyncify: the sync JS call unwinds the stack mid-WASM. Under streaming input, a second `parse()` starts before the first resumes and clobbers shared C globals (`currentParseCallback` etc.), and the first never returns. In opencode this shows up as list items and code blocks freezing partway through the LLM stream.

Fix: a per-worker Promise queue that runs at most one `parse()` at a time. All five `parser.parse()` call sites in `parser.worker.ts` go through it.

## Regression test

The existing 5-parallel `highlightOnce` case in `client.test.ts` never overlapped enough to trigger the race. Added a 100-parallel markdown case in the same file. Unpatched build times out; patched build passes in about 4 seconds together with the file's other 49 tests.

## Evidence

Instrumented `@opentui/core@0.4.5` on opencode `v1.18.18`. JSONL of every `parse.begin` / `parse.done` per worker message.

|                  | before  | after   |
|------------------|---------|---------|
| `parse.begin`    | 8168    | 12272   |
| `parse.done`     | 1369    | 12272   |
| exceptions       | 0       | 0       |
| `worker.onerror` | 0       | 0       |

Before: 6799 parses suspended and never resumed. No exception, no worker crash, just a silent hang. After: fully balanced.

### DIFF
--- a/packages/core/src/lib/tree-sitter/client.test.ts
+++ b/packages/core/src/lib/tree-sitter/client.test.ts
@@ -621,6 +621,26 @@ describe("TreeSitterClient", () => {
     expect(queryLoadLogs.length).toBeLessThanOrEqual(1)
   }, 15000)
 
+  test("should not hang under many parallel highlightOnce calls (Asyncify re-entrancy regression)", async () => {
+    await client.initialize()
+
+    const md =
+      "# heading\n\n" +
+      Array.from({ length: 20 }, (_, i) => `- item ${i} with \`inline code\`\n`).join("") +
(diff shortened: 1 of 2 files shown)


simonklee:
We are closing this pull request because the reported failure does not occur in
the current code. The new 100-request test passes without the change in 25
repeated runs.

PR #1381, opened 2026-08-18
fix(core): selection no longer collapses with y below, x left of text

ofekd:
This fixes the selection breaking when `x` is before the text, but `y` below it. Without the fix, the show selection is flipped.


ofekd:
rebased

simonklee (approved):

PR #1384, opened 2026-08-18
fix(core): correlate OSC 66 probe responses

### LINKED ISSUE #1383: OSC 66 probing: any row-1 cursor report is taken as probe evidence, enabling text scaling on terminals that ignore OSC 66
## Environment

- Node 26.7.0 (darwin arm64) : `node --experimental-ffi repro.mjs`
- `@opentui/core` 0.5.3; code cited at `main` @ `ee083522d7317e405accd578be287840d7b038af`
- Originally hit on xfce4-terminal / Rocky Linux 8 (linux x64, Node 26.5.0)

## Summary

`processCapabilityResponse` derives both OSC 66 text-scaling capabilities from cursor-position reports, but it never ties a report back to the query that produced it, and it treats any column at or beyond the expected one as evidence ([terminal.zig#L1262-L1275](https://github.com/anomalyco/opentui/blob/ee083522d7317e405accd578be287840d7b038af/packages/core/src/zig/terminal.zig#L1262-L1275)):

```zig
if (self.startup_cursor_query_pending and !self.startup_cursor_query_captured and row >= 1 and col >= 1) {
    self.setCursorPosition(col, row, self.state.cursor.visible);
    self.startup_cursor_query_captured = true;
    self.startup_cursor_query_pending = false;
}
                                     // <-- no continue: falls through to the check below
if (row == 1) {
    if (col >= 2) {
        self.caps.explicit_width = true;
    }
    if (col >= 3) {
        self.caps.scaled_text = true;
    }
}
```

So **any** report that lands on row 1 at column >= 2 enables the capabilities, regardless of which query produced it. The probe was meant to be the only evidence ([terminal.zig#L344-L351](https://github.com/anomalyco/opentui/blob/ee083522d7317e405accd578be287840d7b038af/packages/core/src/zig/terminal.zig#L344-L351)):

```
ESC [ H     ESC ] 66 ; w=1 ; <space> ESC \     ESC [ 6 n      -> column 2 if supported
ESC [ H     ESC ] 66 ; s=2 ; <space> ESC \     ESC [ 6 n      -> column 3 if supported
```

That design yields two failure modes, both reproduced below.

### 1. A terminal that echoes the probe payload is detected as supporting it

A terminal that does not recognise OSC 66 prints `66;w=1; ` as ordinary text, leaving the cursor well to the **right** of column 2, which the handler misreads as support. The failure is self-reinforcing: the terminals that cannot handle OSC 66 are exactly the ones that report the largest column, and because that column also clears `col >= 3`, `scaled_text` is enabled as well.

Every affected grapheme is then wrapped in `ESC ] 66 ; w=1 ; <grapheme> ESC \` and echoed as literal text. Bytes captured from a renderer in this state, drawing `A●B◉C│D—E▸F`:

```
b'\x1b]66;w=1;\xe2\x97\x8f\x1b\\'   # ●
b'\x1b]66;w=1;\xe2\x97\x89\x1b\\'   # ◉
b'\x1b]66;w=1;\xe2\x94\x82\x1b\\'   # │
b'\x1b]66;w=1;\xe2\x80\x94\x1b\\'   # —
b'\x1b]66;w=1;\xe2\x96\xb8\x1b\\'   # ▸
```

Only East Asian **ambiguous-width** characters are wrapped — precisely the set a TUI uses for status glyphs and box drawing — so in practice every symbol on screen is destroyed.

Reproduced on xfce4-terminal / CentOS 8, where `printf '\033]66;w=1;X\033\\\n'` prints `66;w=1;X` and the ESC bytes themselves render as `001B` tofu cells, making the echoed run wider still.

### 2. An unrelated row-1 cursor report enables the capabilities with no OSC 66 involved

Because every report reaches the `row == 1` branch, a terminal that ignores OSC 66 outright still gets both capabilities enabled whenever any cursor report lands on row 1 at column >= 2. The startup `ESC[s ESC[6n` query is the obvious source: its reply is consumed by the block above, then falls straight through into the branch.

The existing test seems aware of this aliasing and sidesteps it in the fixture rather than in the code — [terminal_test.zig#L508](https://github.com/anomalyco/opentui/blob/ee083522d7317e405accd578be287840d7b038af/packages/core/src/zig/tests/terminal_test.zig#L508) feeds the startup report as `\x1b[7;11R`, on row 7, before the two row-1 probe replies.

## Reproduction

`repro.mjs` (in the attached `osc66-probe-repro.zip`) feeds responses straight into `processCapabilityResponse`. No terminal, pty or rendering involved. Run it from a directory where `@opentui/core` resolves:

```
$ node --experimental-ffi repro.mjs    # or: bun repro.mjs

PASS  "\u001b[1;2R" terminal implements OSC 66: the probe character advanced the cursor by one cell
        expected explicit_width=true  scaled_text=false
        actual   explicit_width=true  scaled_text=false

PASS  "\u001b[1;3R" terminal implements scaled text: the probe character advanced the cursor by two cells
        expected explicit_width=true  scaled_text=true
        actual   explicit_width=true  scaled_text=true

PASS  "\u001b[1;1R" terminal ignores an unknown OSC: the cursor never moved
        expected explicit_width=false scaled_text=false
        actual   explicit_width=false scaled_text=false

FAIL  "\u001b[1;12R" terminal echoed the probe payload as literal text
        expected explicit_width=false scaled_text=false
        actual   explicit_width=true  scaled_text=true

FAIL  "\u001b[1;5R" unrelated row-1 cursor report (e.g. the startup query)
        expected explicit_width=false scaled_text=false
        actual   explicit_width=true  scaled_text=true
```

The three controls are detected correctly; the two `FAIL` rows are the false positives.

`proposed-tests.zig` (also in the zip) carries the same two cases as regression tests in the style of `terminal_test.zig`, if that shape is more useful.

## Suggested fix

Both parts are needed — fixing either alone leaves the other case open.

1. **Correlate each cursor report with the query that produced it.** A pending-probe counter, incremented per emitted `cursorPositionRequest`, lets the handler distinguish the startup report from the two probe replies. Today the handler only pattern-matches `row == 1`.
2. **Require the exact column.** The explicit-width probe is evidence only at `col == 2`, and the scaled-text probe only at `col == 3`. Any other column means the terminal did something other than advance one cell, which is not evidence of support.

## Workaround

`OPENTUI_FORCE_EXPLICIT_WIDTH=0` (or `false`), set **before process start** — the native lib reads the real environ, as #1300 notes. It skips the probe as well as the emission ([terminal.zig#L863-L870](https://github.com/anomalyco/opentui/blob/ee083522d7317e405accd578be287840d7b038af/packages/core/src/zig/terminal.zig#L863-L870)). Only `true`/`1` and `false`/`0` are recognised; any other value (e.g. `off`) matches neither branch and is ignored, leaving detection in charge.

## Related

- #1300 — same emission path, different trigger: `explicit_width` is correctly detected on kitty, but wrapping U+10EEEE breaks image placeholders. That issue is about *what* gets wrapped; this one is about whether the capability should have been enabled at all.
- #1333 — the same class of defect in the opposite direction: a capability gated on a query answer that a terminal never gives (focus reporting on Konsole).

[osc66-probe-repro.zip](https://github.com/user-attachments/files/31174840/osc66-probe-repro.zip)

dajiaohuang:
## Summary

- correlate cursor-position reports with the two emitted OSC 66 probes
- keep the startup cursor report out of capability detection
- require the exact expected columns instead of accepting larger echoed payload widths

Terminals that echo an unsupported OSC 66 payload can leave the cursor well past the expected probe column. Treating any row-1 report at or beyond the expected column therefore enables capabilities the terminal does not support. The probe state now consumes only the two ordered responses emitted by `queryTerminalSend`, while an unrelated cursor report cannot enable text scaling.

Closes #1383.

## Testing

- filtered Zig 0.16 native tests: 2 passed
- full Zig 0.16 native suite: 2,013 passed, 8 skipped
- native hello example
- `bun run build`
- `bun run fmt:check`
- `bun run lint`
- `zig fmt --check packages/native/src/terminal.zig packages/native/src/tests/terminal_test.zig`


simonklee:
Probably close #1401 as well

simonklee (dismissed):

PR #1389, opened 2026-08-19
fix(core): use boundary semantics for backward selection in edit buffers

ofekd:
When selecting backwards using shift+left with a thin cursor placed after character in position `n`, the next character, position `n+1`, is also selected.

The fix is not entirely clear here since some product decisions discretion is required. Please see if  it's to your liking

### DIFF
--- a/packages/core/src/editor-view.test.ts
+++ b/packages/core/src/editor-view.test.ts
@@ -275,6 +275,16 @@ describe("EditorView", () => {
       expect(view.getSelectedText()).toBe("World")
     })
 
+    it("should not include the anchor character in backward selection (boundary semantics)", () => {
+      buffer.setText("Hello World")
+
+      view.setLocalSelection(8, 0, 8, 0)
+      const changed = view.updateLocalSelection(8, 0, 6, 0)
+      expect(changed).toBe(true)
+      expect(view.getSelection()).toEqual({ start: 6, end: 8 })
+      expect(view.getSelectedText()).toBe("Wo")
+    })
+
     it("should handle wrapped lines with updateLocalSelection", () => {
       buffer.setText("ABCDEFGHIJKLMNOPQRST")
 
--- a/packages/core/src/renderables/EditBufferRenderable.ts
+++ b/packages/core/src/renderables/EditBufferRenderable.ts
@@ -714,8 +714,7 @@ export abstract class EditBufferRenderable extends Renderable implements LineInf
 
     if (!select && this.hasSelection()) {
       const selection = this.getSelection()!
-      const targetOffset = this.cursorOffset === selection.start ? selection.end - 1 : selection.end
-      this.editBuffer.setCursorByOffset(targetOffset)
+      this.editBuffer.setCursorByOffset(selection.end)
       this._ctx.clearSelection()
       this.requestRender()
       return true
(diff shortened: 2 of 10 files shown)


simonklee (approved):

simonklee (changes requested):
nvm, need to do some more testing.

simonklee:
Okay, i think all these selections should be inclusive, independent if you select left or right (obviously). Like visual select in vim. Effectively, the character under the cursor should be selected as well, i think this pr does the opposite.

ofekd:
> Okay, i think all these selections should be inclusive, independent if you select left or right (obviously). Like visual select in vim. Effectively, the character under the cursor should be selected as well, i think this pr does the opposite.

Right, there's a gap here. If I implement as you suggest, thin carets will not work as expected:

```
when caret is after b:

ab|cd

shift+left arrow will then highlight as so:

a*bc*d

but the expect highlight for thin caret is:

a*b*cd

```

A fat caret would be on `c` in that situation, so highlighting `bc` makes sense

simonklee:
> > Okay, i think all these selections should be inclusive, independent if you select left or right (obviously). Like visual select in vim. Effectively, the character under the cursor should be selected as well, i think this pr does the opposite.
> 
> Right, there's a gap here. If I implement as you suggest, thin carets will not work as expected:
> 
> ```
> when caret is after b:
> 
> ab|cd
> 
> shift+left arrow will then highlight as so:
> 
> a*bc*d
> 
> but the expect highlight for thin caret is:
> 
> a*b*cd
> ```
> 
> A fat caret would be on `c` in that situation, so highlighting `bc` makes sense

This is how vim does it. Alternative is to have multiple modes but not a fan of the complexity added

ofekd:
> > > Okay, i think all these selections should be inclusive, independent if you select left or right (obviously). Like visual select in vim. Effectively, the character under the cursor should be selected as well, i think this pr does the opposite.
> > 
> > 
> > Right, there's a gap here. If I implement as you suggest, thin carets will not work as expected:
> > ```
> > when caret is after b:
> > 
> > ab|cd
> > 
> > shift+left arrow will then highlight as so:
> > 
> > a*bc*d
> > 
> > but the expect highlight for thin caret is:
> > 
> > a*b*cd
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > A fat caret would be on `c` in that situation, so highlighting `bc` makes sense
> 
> This is how vim does it. Alternative is to have multiple modes but not a fan of the complexity added

Neovim at least has 2 thin caret options. In normal or visual mode it is before the character it refers, and in insert mode is it after it.

If you use it in insert mode, then turn on shift+arrow highlight, it will behave as this PR suggests.

This indicates a bug in another level -- OpenTUI thin cursor is positioned like in Vim's normal and visual modes but behavior is a mix of insert mode. I don't know the code well enough to say, but we may be able to fix the cursor position, then get all the behavior for free.

Since all of those components in OpenTUI are "insert mode"-like, I believe it's correct to do.

simonklee:
Thanks for the input on this — i ended up doing a quiet extensive pass on selection and found a bunch of tiny issues all around.

ofekd:
> Thanks for the input on this — i ended up doing a quiet extensive pass on selection and found a bunch of tiny issues all around.

Thank you for addressing it so quickly

PR #1394, opened 2026-08-20
fix(core): pass transient FFI buffers directly

Hona:
## Summary

- use `buffer` ABI parameters and pass typed-array owners directly for transient synchronous Bun FFI calls
- lower portable `buffer` parameters to Node's stable `pointer` path while still passing owner objects directly
- keep `ptr` only for nullable, mixed native-pointer, raw `ArrayBuffer`, callback, and retained-memory cases
- stabilize retained text-memory views through `.buffer` before resolving their native address
- preserve raw-pointer compatibility while allowing direct buffers in supersample, packed-buffer, matrix, and grayscale APIs
- document the Bun 1.3.14 and Bun 1.4+ ownership rules in `AGENTS.md`

## Why

A Bun `FastTypedArray` may store its data inline. Calling `ptr(view)` captures that address, but a later first access to `view.buffer` can move the data into separate `ArrayBuffer` storage and leave the native address stale. Passing the owner directly lets the FFI backend borrow the current storage for synchronous calls and allows Bun's `buffer` fast path to keep the owner visible to the JIT.

This keeps nullable and true native-pointer parameters on `ptr`, where `buffer` cannot represent the ABI, while avoiding pre-resolved addresses for transient memory. Retained pointers explicitly materialize stable backing storage before `ptr(view)` and keep the view alive for the native lifetime.

Node 26.4's Linux optimized `buffer` trampoline delivered a null pointer to a multi-argument audio call in CI. OpenTUI therefore maps its portable `buffer` descriptor to Node `pointer`, whose documented owner-borrowing path passes the same views correctly. Bun continues to receive the `buffer` descriptor.

Related Bun investigation: oven-sh/bun#32054 and oven-sh/bun#32055.

## Testing

- `bunx bun@1.3.14 test src/tests/ffi-borrowed-pointer-callsites.test.ts` (23 passed)
- `bun test src/tests/ffi-borrowed-pointer-callsites.test.ts` on Bun 1.4 canary (23 passed)
- `bun run test:js` (5,397 passed, 23 skipped)
- `bun run test:js:node` with Node 26.4.0 (4,665 passed, 6 skipped)
- `bun run test:dist`
- `bun run build:lib`
- `bun run fmt:check`
- `bun run lint`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,10 +40,16 @@ terminal or platform is unavailable locally.
 - Portable symbol signatures must stay within the `node:ffi`/`bun:ffi` intersection. Use explicit widths such as
   `u32`/`u64`, not backend-only ABI names such as `usize`, `napi_env`, or `napi_value`; represent `i64`/`u64` as `bigint`,
   native booleans as `0`/`1`, and shared pointers as `number | bigint`.
-- Pass transient `ArrayBuffer` values or views directly to synchronous pointer parameters so the backend borrows the
-  owner. Do not pre-resolve them with `ptr()`.
-- Use `ptr(view)` only for addresses stored in structs or retained by native code, and keep the backing buffer alive for
-  the complete native lifetime.
+- Default to `buffer` for transient, non-null `TypedArray` parameters and pass the view directly. Do not call `ptr()`.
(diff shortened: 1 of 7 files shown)


kommander (approved):
Approved.

simonklee (approved):

PR #1396, opened 2026-08-20
fix(native): enable focus tracking without DECRQM

### LINKED ISSUE #1333: Focus reporting gated on a DECRQM answer: never enabled on Konsole, which reports focus but does not answer the query
`@opentui/core` sends `CSI ? 1004 $ p` (DECRQM) at startup and only enables focus reporting if the terminal answers. Konsole sends focus events correctly but does not answer that query, so the mode is never enabled and `CliRenderEvents.FOCUS` / `BLUR` never fire — on a terminal that supports the feature.

Reporting a mode and answering a *query about* that mode are independent in practice. Gating on the answer means the feature silently degrades on any terminal that implements one but not the other, with no way for the app to tell.

## Reproducing

Konsole 25.08.1, Linux.

**1. Konsole does report focus.** In a Konsole tab (non-canonical mode matters — `cat` in canonical mode buffers until a newline and hides this):

```sh
stty -icanon -echo; printf '\033[?1004h'; timeout 15 cat -v; printf '\033[?1004l'; stty sane; echo
```

Click away and back: `^[[O` and `^[[I` arrive as expected.

**2. Konsole does not answer the query.**

```sh
stty -icanon -echo; printf '\033[?1004$p'; timeout 3 cat -v; stty sane; echo
```

Silent.

**3. opentui therefore never enables it.** Capturing everything an opentui app writes to a pty, the only occurrence of 1004 is the query:

```
sent:            [?1004$p
modes enabled:   [?1000h [?1002h [?1003h [?1006h [?1049h [?2004h [?2026h [?2027h [?2031h [?25h
```

Feeding a synthetic `\033[?1004;2$y` reply back into the pty produces exactly one `[?1004h`, confirming the gate.

**4. The parser does not need the gate.** With focus reporting never enabled, writing a bare `\033[O` into the pty still makes the app react — an app that slows its spinner on blur went from 29.2% to 11.2% of a core. So `FOCUS`/`BLUR` are dispatched whether or not opentui enabled the mode; only the enabling is missing.

## Suggestion

Enable 1004 unconditionally rather than gating on the DECRQM answer. Terminals that don't support the mode ignore the DECSET, which is the normal cost of the sequence, whereas the current gate silently disables a working feature on terminals that stay quiet.

If the probe is wanted for other reasons, treating "no answer" as "enable anyway" would have the same effect — the failure mode of enabling on an unsupporting terminal is nothing, while the failure mode of not enabling on a supporting one is a dead event.

## Workaround

Setting the mode from the application works, since the parser already handles the events:

```ts
process.stdout.write("\x1b[?1004h")   // after createCliRenderer
process.stdout.write("\x1b[?1004l")   // on teardown
```

Measured 30.8% -> 14.3% on blur with that in place and no reply to the probe. It does reach around the renderer's terminal-state ownership though, and it is likely lost across `terminal_suspend` resume, which is why this feels like it belongs in opentui.

Found via opencode, whose `attention.ts` gates notifications on the same events (`when: "blurred"`), so those never suppress on Konsole either.

Versions: `@opentui/core` 0.4.5, Konsole 25.08.1, Linux, bun 1.3.13.

dajiaohuang:
## Summary

- enable focus tracking during normal feature activation even when DECRQM has no response
- keep the detected capability flag tied to an actual terminal response
- cover the no-DECRPM path with a native regression test

Some terminals, including Konsole, implement focus reporting but do not answer the `CSI ? 1004 $ p` query. Query silence therefore cannot be used as evidence that mode 1004 is unsupported. The renderer now requests the existing DEC private mode unconditionally, while the existing terminal state continues to own teardown and focus-in restoration.

The compatibility assumption is that terminals without mode 1004 support ignore the unknown DECSET sequence, matching normal DEC private-mode behavior. No configuration default or public interface changes.

Closes #1333

## Testing

- filtered Zig 0.16 native tests: 2 passed
- full Zig 0.16 native suite: 2,013 passed, 8 skipped
- native hello example
- `bun run build`
- `bun run fmt:check`
- `bun run lint`
- `zig fmt --check packages/native/src/terminal.zig packages/native/src/tests/terminal_test.zig`
- `git diff --check`

### DIFF
--- a/packages/native/src/terminal.zig
+++ b/packages/native/src/terminal.zig
@@ -426,9 +426,7 @@ pub fn enableDetectedFeatures(self: *Terminal, tty: anytype, use_kitty_keyboard:
         try self.setBracketedPaste(tty, true);
     }
 
-    if (self.caps.focus_tracking) {
-        try self.setFocusTracking(tty, true);
-    }
+    try self.setFocusTracking(tty, true);
 
     // queryTerminalSend already enabled mode 2031 during normal startup.
     if (!self.state.color_scheme_updates) {
--- a/packages/native/src/tests/terminal_test.zig
+++ b/packages/native/src/tests/terminal_test.zig
@@ -1004,9 +1004,11 @@ test "isXtversionTmux - detects tmux from xtversion" {
     try testing.expect(term.isXtversionTmux());
 }
 
-// ============================================================================
+// =====================================================================}
+
 // GRAPHEME CURSOR POSITIONING CAPABILITY TESTS
-// ============================================================================
+// =====================================================================}
+
 
 test "processCapabilityResponse - tmux sets explicit_cursor_positioning" {
     var term: Terminal = .{};
@@ -1109,9 +1111,11 @@ test "processCapabilityResponse - XTGETTCAP Ms only establishes positive support
     try testing.expectEqual(Terminal.Osc52Support.unknown, unknown.osc52_support);
 }
 
-// ============================================================================
+// =====================================================================}
+
 // CLIPBOARD (OSC 52) TESTS
-// ============================================================================
+// =====================================================================}
+
 
 test "writeClipboard - generates basic OSC52 sequence" {
     if (builtin.os.tag == .windows) return error.SkipZigTest;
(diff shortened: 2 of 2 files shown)


dajiaohuang:
Rebased onto current main (06f597f), preserving both the new upstream width-profile tests and this focus-tracking regression. The focused native test passes; the complete native suite reports 2,088 passed and 60 skipped.

The required root build compiled the native library but stopped at Windows release-symbol packaging (Missing DLL PDB GUID with the available LLVM 13 tools). I have not claimed that packaging check passed.

PR #1406, opened 2026-08-22
fix: disable OSC 66 for Ghostty

### LINKED ISSUE #1401: Malayalam graphemes render at incorrect cell positions
## Description

Malayalam grapheme clusters can render at incorrect horizontal positions in static text. In the observed case, the final glyphs appeared near or beyond the right edge of the containing panel instead of remaining contiguous with the rest of the title.

The exact string was:

```text
OpenCode search configuration പരിശോധിക്കൽ
```

This was rendered in an OpenCode session tab using a fixed-width `<text>` with `wrapMode="none"`. The title was not being scrolled.

## Expected

The Malayalam word renders contiguously after the ASCII prefix and remains within the text bounds.

## Actual

Parts of the Malayalam word render as separated glyphs at incorrect cell positions, including near the right edge of the panel.

## Environment

- OpenTUI: `0.5.6`
- Terminal: Ghostty `1.3.1`
- OS: macOS `25.5.0` (arm64)
- Consumer: OpenCode `0.0.0-beta-17728`

This may share a root cause with #479, but that report concerns corruption while scrolling Thai text. This reproduction is static Malayalam text on a newer OpenTUI version.

dajiaohuang:
## Summary

- ignore OSC 66 width-probe false positives when XTVERSION identifies Ghostty
- preserve `OPENTUI_FORCE_EXPLICIT_WIDTH=true` as an explicit escape hatch
- document the Ghostty compatibility behavior

## Why

OpenTUI's buffer places the reported Malayalam graphemes contiguously, but Ghostty currently parses OSC 66 without applying its sizing metadata (ghostty-org/ghostty#10333). The probe's plain space therefore advances the cursor to column 2 and incorrectly enables explicit-width output. Subsequent graphemes are wrapped in OSC 66 even though Ghostty does not implement the rendering side of the protocol.

This assumes Ghostty should remain on standard Unicode-width output until its open implementation issue is completed. The exact XTVERSION identity guard is small and reversible, and the force-on environment override remains available for testing future releases.

Fixes #1401.

## Validation

- regression test failed before the fix and passes after it
- `bun run test:native` (1988 passed, 53 skipped)
- `bun run build`
- `bun run fmt:check`
- `bun run lint`

### DIFF
--- a/packages/core/docs/development.md
+++ b/packages/core/docs/development.md
@@ -151,4 +151,4 @@ OPENTUI_FORCE_EXPLICIT_WIDTH=false
 - Falls back to standard width calculation
 - No visual artifacts on unsupported terminals
 
-**Modern Terminals:** If your terminal supports OSC 66 (Kitty, Ghostty, WezTerm, Alacritty, iTerm2), you don't need this setting - they work correctly by default.
+**Modern Terminals:** If your terminal supports OSC 66 (Kitty, WezTerm, Alacritty, iTerm2), you don't need this setting - they work correctly by default. OpenTUI automatically disables OSC 66 for Ghostty versions that parse the sequence but do not yet apply its sizing metadata.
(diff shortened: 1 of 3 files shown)


simonklee:
diagnoses the wrong cause + regression is fabricaed/impossible

PR #1423, opened 2026-08-24
fix(core): normalize embedded terminal keys

jlongster:
## Summary
- reject parser escape codes such as `[A` and `[27u` as physical key identifiers
- infer Web-style physical identities (`ArrowUp`, `KeyC`, `Digit1`, `Escape`) before calling Ghostty
- derive the unshifted codepoint from that normalized physical identity
- cover raw arrow input and nested Kitty character/escape re-encoding

## Why this is needed
OpenTUI key events use `code` for two different kinds of values. Kitty and raw escape parsing can produce parser codes such as `[A` for Up or `[27u` for Kitty Escape, while other events provide Web-style physical codes such as `KeyA`. Ghostty's key encoder accepts the latter physical identities, not parser escape fragments. Passing `[A` through as the physical key leaves Ghostty unable to encode an arrow key.

Embedded terminals also need to re-encode input when the program inside them enables the Kitty keyboard protocol. In that mode, text alone is insufficient for modified and disambiguated keys: Ghostty needs a normalized physical identity and its unshifted codepoint. Raw character events may not include `code`, so this change infers `KeyA`/`Digit1`; Kitty escape events whose parser code is `[27u` fall back to `Escape`. The regression verifies that Ctrl+C and Escape are re-emitted as Kitty sequences for a nested terminal.

## Testing
- `bun test src/renderables/EmbeddedTerminal.test.ts` (16 passed)
- `oxlint packages/core/src/renderables/EmbeddedTerminal.ts packages/core/src/renderables/EmbeddedTerminal.test.ts`
- `oxfmt --check packages/core/src/renderables/EmbeddedTerminal.ts packages/core/src/renderables/EmbeddedTerminal.test.ts`

### DIFF
--- a/packages/core/src/renderables/EmbeddedTerminal.test.ts
+++ b/packages/core/src/renderables/EmbeddedTerminal.test.ts
@@ -98,6 +98,12 @@ describe("EmbeddedTerminalRenderable", () => {
     )
     expect(new TextDecoder().decode(terminal.encodeKey(keyEvent({ name: "😀", sequence: "😀" })))).toBe("😀")
     expect(new TextDecoder().decode(terminal.encodeKey(keyEvent({ name: "space", sequence: " " })))).toBe(" ")
+    expect(new TextDecoder().decode(terminal.encodeKey(keyEvent({ name: "up", sequence: "\x1b[A", code: "[A" })))).toBe(
+      "\x1b[A",
+    )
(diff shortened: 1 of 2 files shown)


simonklee (approved):

clarkarch:
thank you for this fix

PR #1426, opened 2026-08-24
feat(native): filter Ghostty logs

jlongster:
## Summary
- recognize the selected Ghostty `std.log` scopes in OpenTUI's root logger
- keep all standard-library logs silent by default
- read `OTUI_GHOSTTY_LOG_LEVEL` inside the native `logFn` and forward enabled messages through the existing `logger.logMessage` path
- support `error`/`err`, `warn`/`warning`, `info`, and `debug` thresholds
- document the environment variable

## Behavior
Unset or invalid `OTUI_GHOSTTY_LOG_LEVEL` values emit no `std.log` messages. Recognized Ghostty scopes are forwarded up to the configured severity. Logs from every other scope are discarded. Explicit OpenTUI logger calls are unchanged.

## Testing
- release `bun run build` from `packages/core`
- `bun run test:native` from `packages/core` (2066 passed, 8 skipped)
- `zig fmt --check packages/native/src/lib.zig packages/native/src/logger.zig`
- `oxfmt --check packages/web/src/content/docs/reference/env-vars.mdx`

### DIFF
--- a/packages/native/src/lib.zig
+++ b/packages/native/src/lib.zig
@@ -4,19 +4,58 @@ const build_options = @import("build_options");
 const Allocator = std.mem.Allocator;
 
 pub const std_options: std.Options = .{
-    .logFn = discardStdLog,
+    .log_level = .debug,
+    .logFn = handleStdLog,
 };
 
-fn discardStdLog(
+fn handleStdLog(
     comptime message_level: std.log.Level,
     comptime scope: @EnumLiteral(),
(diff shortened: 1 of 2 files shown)


simonklee (approved):
was about to fix the docs but you beat me to it. lgtm

jlongster:
Thank you! Talked with @kommander a bunch too on this approach on slack and it's the best thing we could come up with

Issue #1408, opened 2026-08-22
core: word forward considers chunks


simonklee:
`getNextWordBoundary` only inspects wrap breaks inside the current rope chunk. A CJK ASCII split is emitted only when both scripts sit in the same chunk. Typing or a second `insertText` puts `世界` and `abc` in different chunks, so word-forward, word-back, `ctrl+w`, and option+arrow treat the run as one word. `setText("世界abc")` still splits. The tests only use `setText`, so they never see it.

**Manual (editor-demo)**
1. Open editor-demo.
2. Type `世界abc` (do not paste the whole string).
3. Go to the start of that run.
4. `Ctrl+Right` jumps to the end (column 7). One word.
Paste `世界abc` as one shot, or start from a `setText` of that string: first `Ctrl+Right` lands on `a` (column 4). Two words.

**Simple repro**

```
cd packages/core
bun -e '
import { EditBuffer } from "./src/edit-buffer.ts"

function first(buf) {
  buf.setCursorToLineCol(0, 0)
  return buf.getNextWordBoundary().col
}

const a = EditBuffer.create("wcwidth")
a.setText("世界abc")
console.log("setText", first(a)) // 4  (on a)
a.destroy()

const b = EditBuffer.create("wcwidth")
b.insertText("世界")
b.insertText("abc")
console.log("two inserts", first(b)) // 7  (EOL)
b.destroy()
'
```
Same bytes. `setText` splits. Two inserts do not. Word-forward only sees wrap breaks inside the current chunk.

simonklee:
Commit d2089a81ae2c88e8b817867582a81682fbb3194f fixed this issue. Forward and
backward word movement now detect CJK/ASCII boundaries between separate rope
chunks.

Issue #799, opened 2026-03-10
[Bug] Markdown CJK wrap duplicates glyph at specific widths


stephanoskomnenos:
## Summary
Rendering Markdown with CJK content can duplicate a fullwidth glyph at specific terminal widths where the line wraps.

## Repro
Repo: https://github.com/stephanoskomnenos/opentui-md-wrap-demo

Commands:
- `scripts/capture.sh --tmux-columns 38 --out-png ./tmp/demo-38.png`
- `scripts/capture.sh --tmux-columns 39 --out-png ./tmp/demo-39.png`
- `scripts/capture.sh --tmux-columns 40 --out-png ./tmp/demo-40.png`

## Expected
No duplicated glyph on wrap.

## Actual
At width 39, the wrap boundary duplicates a CJK glyph ("确" appears both at end of line and again on next line).

## Screenshots (38/39/40)
![38](https://raw.githubusercontent.com/stephanoskomnenos/opentui-md-wrap-demo/main/tmp/demo-38.png)
![39](https://raw.githubusercontent.com/stephanoskomnenos/opentui-md-wrap-demo/main/tmp/demo-39.png)
![40](https://raw.githubusercontent.com/stephanoskomnenos/opentui-md-wrap-demo/main/tmp/demo-40.png)

ANSI/HTML/PNG are included in the repo under `tmp/`.

## Environment (devcontainer)
- @opentui/core: 0.1.87
- @opentui/react: 0.1.87
- Bun runtime (Linux)
- Rendering via `<markdown>` component

kommander:
@simonklee didn't we just fix that recently, is that a regression?

simonklee:
> [@simonklee](https://github.com/simonklee) didn't we just fix that recently, is that a regression?

Not that i remember. i have a branch where i've fixed it — but that is massive (all my scan layout wrapping things) https://github.com/anomalyco/opentui/compare/main...simonklee:scan-wrap

I'll have a look at this.

simonklee:
Commit d2089a81ae2c88e8b817867582a81682fbb3194f fixes the CJK wrap duplication.
The renderer now uses exact byte ranges for wrapped text. A native regression
test uses the text from this report and checks the rendered rows for duplicate
glyphs.

PR #1428, opened 2026-08-25
refactor(core): make wrapped layout byte-accurate

kommander:
- Makes wrapped layout byte/column authoritative; fixes CJK/emoji splits and leading wrap whitespace.
- Removes legacy wrap paths and specializes render/measure hot paths.
- Uses `u32` chunk widths, streamed large-chunk layout, and lazy cold caches.
- Fixes a pre-existing `setTabWidth` bug by refreshing cached rope metrics, including undo/redo roots.

| ReleaseFast benchmark | `main` median | PR median | Median paired result |
|---|---:|---:|---:|
| **Validated equivalent output:** char wrap, 62.5 KiB, width 40 | 0.264 ms | 0.191 ms | **27.8% less time** |
| **Validated equivalent output:** char wrap, 62.5 KiB, width 80 | 0.150 ms | 0.111 ms | **26.2% less time** |
| **Validated equivalent output:** char wrap, 62.5 KiB, width 120 | 0.103 ms | 0.075 ms | **26.9% less time** |
| **Validated equivalent output:** word wrap, 62.5 KiB, width 80 | 0.298 ms | 0.287 ms | **3.7% less time** |
| Word wrap, 1 MiB multiline, width 40 | 14.816 ms | 12.123 ms | **18.4% less time** |
| Word wrap, 1 MiB multiline, width 80 | 12.508 ms | 10.845 ms | **12.7% less time** |
| Word wrap, 1 MiB multiline, width 120 | 11.816 ms | 10.100 ms | **14.5% less time** |
| Incremental appends + wrapped measurement | 276.111 ms | 50.699 ms | **5.42× faster** |
| Incremental appends + intrinsic measurement | 63.090 ms | 62.931 ms | **neutral** |
| Repeated cached wrapped measurement | 12.584 ms | 7.851 ms | **38.1% less time** |
| First char wrap + render, 500 lines | 0.382 ms | 0.337 ms | **10.4% less time** |

- Alternating CPU-pinned pairs against current `main` `8046c846`; absolute columns are marginal medians, comparisons use median paired `main / PR` ratios. Current head is `6c02af05`; product implementation is `98709be8`, followed only by comments/benchmark corrections.
- The validated cases use one exact 64,000-column chunk and assert complete output outside timing. The old ~2 MiB rows were removed because `main` clipped layout at 65,535 columns while the PR processed all 1.86M columns.
- Uncached ASCII render chunks stream instead of retaining rich break metadata; existing caches and medium non-ASCII caching remain.
- Multiline and measurement cases use the same requests and inputs but intentionally corrected wrapping output, so they are end-to-end comparisons rather than isolated scanner claims.
- The direct scanner benchmark records 0.105 ms per 62.5 KiB scan on the PR; `main` has no equivalent API, so no direct scanner speedup is claimed.
- The three-line `setText` microbenchmark is neutral over 20 paired processes: 0.04% difference, paired CI −3.2% to +2.9%.
- Internal unit-name cleanup is split into stacked PR #1429.

### DIFF
--- a/packages/native/src/edit-buffer.zig
+++ b/packages/native/src/edit-buffer.zig
@@ -751,7 +751,7 @@ pub const EditBuffer = struct {
                     continue;
                 };
 
-                if (utf8.isCjkAsciiTransition(previous_word_class, layout.first_word_class) and
+                if (utf8.isCjkAsciiTransition(previous_word_class, layout.word_classes.first) and
                     cols_before > cursor.col and cols_before <= line_width)
                 {
                     const offset = iter_mod.coordsToOffset(self.tb.rope(), cursor.row, cols_before) orelse cursor.offset;
@@ -799,7 +799,7 @@ pub const EditBuffer = struct {
                     // Mark that we've processed/passed the cursor position
                     passed_cursor = true;
                 }
-                previous_word_class = layout.last_word_class;
+                previous_word_class = layout.word_classes.last;
                 cols_before = next_cols;
             }
         }
@@ -837,7 +837,7 @@ pub const EditBuffer = struct {
                     continue;
                 };
 
-                if (utf8.isCjkAsciiTransition(previous_word_class, layout.first_word_class) and cols_before < cursor.col) {
+                if (utf8.isCjkAsciiTransition(previous_word_class, layout.word_classes.first) and cols_before < cursor.col) {
                     last_boundary = cols_before;
                 }
 
@@ -848,7 +848,7 @@ pub const EditBuffer = struct {
                     }
                 }
 
-                previous_word_class = layout.last_word_class;
+                previous_word_class = layout.word_classes.last;
                 cols_before = next_cols;
                 if (cursor.col <= cols_before) break;
             }
--- a/packages/native/src/edit-buffer.zig
+++ b/packages/native/src/edit-buffer.zig
@@ -672,6 +672,7 @@ pub const EditBuffer = struct {
         var current_meta_buffer: [64]u8 = undefined;
         const current_meta = try self.encodeCurrentCursorMeta(current_meta_buffer[0..]);
         const prev_meta = try self.tb.rope().undo(current_meta);
+        self.tb.refreshTabWidthMetrics();
 
         const restored = try self.restoreCursorFromMeta(prev_meta);
 
@@ -689,6 +690,7 @@ pub const EditBuffer = struct {
 
     pub fn redo(self: *EditBuffer) ![]const u8 {
         const next_meta = try self.tb.rope().redo();
+        self.tb.refreshTabWidthMetrics();
 
         const restored = try self.restoreCursorFromMeta(next_meta);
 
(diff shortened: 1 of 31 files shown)


kommander:
Hey @simonklee — short map before the diff eats the scrollbar:

This is a current-tree port of Sebastian’s old `tbv-refactor` architecture (not your `scan-wrap` branch), but it converges on the invariant you pushed there: rendered slices carry exact UTF-8 byte windows **and** display-column widths together. Public `lineStartCols` stays column-based.

The main change is one comptime-specialized `TextBufferView` state machine for render/measure and char/word wrapping. The scanner emits byte+column break metadata, word/class state survives rope-chunk boundaries, and `VirtualChunk`s render their stored byte windows directly. Separator whitespace and truncation keep logical/grapheme boundaries; medium render chunks cache lazily, small/large chunks stream, measurement reuses but does not create that cache, and OOM cannot publish partial layout.

Supporting work widens chunk widths to `u32` (#1304 overlap), covers the exact #799 CJK duplication vector, carries CJK/ASCII classes across chunks (the #1408 root cause), and subsumes #658’s stale-tab-metrics fix through undo/redo roots. No public ABI or line-start semantic change; viewport-cache redesign and editor/wrap semantic separation remain out of scope. Unit-name churn is isolated in #1429.

Perf footnote: the validated equal-output, break-dense 62.5 KiB case initially exposed a **2.02× slowdown**. Profiling pinned that on cold ASCII cache materialization; the existing streaming path now handles uncached ASCII, bringing the final paired result to parity/slightly faster (0.284 → 0.276 ms) while cutting TextBuffer arena use from 682 KiB to 2.9 KiB. Multiline wrapping uses 13–18% less time, incremental wrapped measurement is 5.42× faster, and cached measurement uses 38.1% less time; corrected-output workloads remain qualified, and no direct scanner speedup versus `main` is claimed.

Main review question: does the single shared state machine now look like the right correctness/complexity trade? No performance hall pass required. 🙂

simonklee:
I love the look of those snapshot tests ♥

packages/native/src/utf8.zig
```diff
@@ -174,33 +176,33 @@ pub inline fn decodeUtf8Unchecked(text: []const u8, pos: usize) struct { cp: u21
...
-        0x2010, // HYPHEN
-        0x3001, // IDEOGRAPHIC COMMA
-        0x3002, // IDEOGRAPHIC FULL STOP
-        0xFF01, // FULLWIDTH EXCLAMATION MARK
-        0xFF0C, // FULLWIDTH COMMA
-        0xFF1A, // FULLWIDTH COLON
-        0xFF1F, // FULLWIDTH QUESTION MARK
-        => true,
-        else => false,
+        0x00A0,
```
simonklee: Lets keep the comments here

packages/native/src/text-buffer-iterators.zig
```diff
@@ -362,43 +361,6 @@ pub fn getPrevGraphemeWidth(rope: *UnifiedRope, mem_registry: *const MemRegistry
...
 }
 
-pub const CharOffsetColumnInfo = struct {
-    col: u32,
-    width: u32,
-};
-
-/// char_offset is grapheme-count based (not raw codepoint)
-/// grapheme_idx and col_delta carry incremental state across calls
-pub fn charOffsetToColumn(
```
simonklee, line 372: 🥳

packages/native/src/edit-buffer.zig
```diff
@@ -735,42 +735,42 @@ pub const EditBuffer = struct {
...
+                    cols_before = next_cols;
+                    passed_cursor = passed_cursor or cursor.col < next_cols;
+                    previous_word_class = .other;
+                    continue;
+                };
+
+                if (utf8.isCjkAsciiTransition(previous_word_class, layout.word_classes.first) and
+                    cols_before > cursor.col and cols_before <= line_width)
+                {
+                    const offset = iter_mod.coordsToOffset(self.tb.rope(), cursor.row, cols_before) orelse cursor.offset;
```
simonklee, line 755: I hate that we still have to convert between coordinates, i think i got rid of the coordinates to offset kind of thing when i looked at it once, but that required quiet a few changes. Also, maybe looks okay here since its for the fn return.

_Just a note for myself as i look through the code, so i remember what to ask the agent afterwards._

packages/native/src/utf8.zig
kommander: Restored the Unicode codepoint names in place in `42fee150`—the switch still classifies whitespace vs punctuation, but the useful annotations are back.

simonklee:
https://opentui-bench.fly.dev/compare?benchmark_kind=zig&js_runtime=bun&base=653&curr=655&category=TextBuffer+Wrapping
| Benchmark | Baseline P50 | Current P50 | P50 delta (compared / baseline) |
|---|---|---|---|
| TextBufferView wrap (char, width=120, single-line) | 638.05µs | 18.18ms | +2749.5% |
| TextBufferView wrap (char, width=80, single-line) | 727.71µs | 19.58ms | +2590.1% |
| TextBufferView wrap (char, width=40, single-line) | 980.17µs | 23.72ms | +2319.6% |
| TextBuffer setText small (3 lines, 40 bytes) | 4.83µs | 5.92µs | +22.6% |
| TextBufferView wrap (char, width=120, multi-line) | 7.41ms | 7.33ms | -1.1% |
| TextBuffer setText large (4590 lines, 1.00 MiB) | 6.39ms | 6.30ms | -1.5% |
| TextBufferView wrap (char, width=80, multi-line) | 8.71ms | 8.10ms | -7.0% |
| TextBufferView wrap (word, width=40, single-line) | 30.07ms | 25.05ms | -16.7% |
| TextBufferView wrap (char, width=40, multi-line) | 12.50ms | 10.37ms | -17.0% |
| TextBufferView wrap (word, width=120, multi-line) | 13.88ms | 10.56ms | -23.9% |
| TextBufferView wrap (word, width=80, multi-line) | 15.67ms | 11.64ms | -25.7% |
| TextBufferView wrap (word, width=80, single-line) | 30.12ms | 20.26ms | -32.7% |
| TextBufferView wrap (word, width=40, multi-line) | 19.45ms | 12.98ms | -33.3% |
| TextBufferView wrap (word, width=120, single-line) | 29.90ms | 18.71ms | -37.4% |

kommander:
@simonklee the scary char single-line rows are not workload-equivalent: run 653 on `main` clamps `TextChunk.width` to the `u16` ceiling (65,535), and char wrapping stops there; run 655 uses `u32` and processes the full ~2 MiB line. So that row is effectively ~65K columns vs ~2M columns—about 32× more work, not a 28× algorithm regression. The View arena numbers tell the same story: 477,088 bytes vs 9,634,232 bytes. Word mode still scans the full byte input on `main`, which is why its single-line rows do not show that pattern.

I excluded those rows from the PR claims, but leaving them in the dashboard suite is still misleading. I’m removing/replacing them with a ≤65,535-column matched workload. The validated equivalent-output 64,000-column case currently measures 0.284 ms on `main` vs 0.276 ms here.

The small `setText` row is separate: +1.09 µs over only three samples. I’m running alternating paired processes before calling that real or changing code.

kommander:
Follow-up: fixed in `6c02af05`. The six old ~2 MiB single-line rows are gone; they are replaced by newly named, validated 64,000-column cases, so the dashboard cannot pair them with the incompatible clipped history.

Fresh alternating pairs on equal work:
- char width 40/80/120: **27.8% / 26.2% / 26.9% faster**
- View arena: **32–34% smaller**
- small `setText`: **0.04% difference**, paired CI **−3.2% to +2.9%** (neutral)

The PR table and report now use those numbers.

PR #1429, opened 2026-08-25
refactor(core): clarify text layout units

kommander:
- Stacked on #1428; no behavior changes.
- Renames internal byte, render-cluster, display-column, and document-cell units explicitly.
- Keeps public TypeScript and C ABI names unchanged.


simonklee:
For some reason i thought i had already merged a pr that did something similar once. Maybe that got stuck in one of my wrap branches.

simonklee (dismissed):
Anyway, this looks like just a rename. Also, like width_cols over cols_width 👍🏼

PR #1430, opened 2026-08-25
core: gate capability mutation on an outstanding probe, not response shape

mrrsavila:
## Problem

`CliRenderer.processCapabilitySequence()` accepted any stdin sequence matching the capability-response *shape* at *any* time — an unconditional, lifetime-open listener.

xterm encodes modified F1–F4 as `CSI 1;<mod> P/Q/R/S`. The `R` slot (e.g. **Ctrl+F3** = `ESC [ 1 ; 5 R`) is byte-identical to a **row-1 cursor position report**. The native layer reads `col >= 2` as proof of the explicit-width/scaled-text protocol and latches both capabilities — **set-only, never cleared**. One modified-F-key chord arriving at the wrong moment permanently corrupts rendering until restart.

Shape is not provenance: bytes that merely *look like* a capability response must never be able to latch capabilities.

## Fix

Introduce `OutstandingCapabilityProbes` (`lib/outstanding-capability-probes.ts`): the startup probe window armed in `setupTerminal()` is now the sole provenance of capability mutations.

- `consume()` is the **only producer** of `CapabilityUpdate`; `CliRenderer.applyCapabilities()` is the **only mutation sink**. With no probe outstanding, mutation is unrepresentable at the API level.
- The renderer's `capabilityTimeoutId` remains the window timer; all window-state reads (the capability gate, `waitForXtVersion`, `destroy` teardown) now go through the registry.
- **Recognition stays shape-based**: with no probe outstanding, capability-shaped bytes (including malformed XTGETTCAP replies) are still swallowed rather than forwarded as application input — they simply cannot latch state. In-window behavior for genuine probe replies is unchanged.

Residual, documented in the module header: bytes arriving *while* a probe window is open still pass — within the window a colliding chord is indistinguishable from a genuine reply at this layer. That is a far smaller surface than a lifetime-open listener and is not solvable here.

## Tests

- New `renderer.capability-provenance.test.ts` (**written red-first** per AGENTS.md — fails on the pre-fix gate, passes after):
  - a modified-F3-shaped chord with no probe outstanding mutates nothing
  - a DA1-shaped response with no probe outstanding mutates nothing
  - a width-probe reply *inside* the window still applies
  - window expiry closes provenance
- Existing tests that drive capability responses without `setupTerminal()` (palette, console-startup) now arm the registry explicitly via `startCapabilityDetectionWindow`.

## Verification

- `bun test` in `packages/core`: **5467 pass / 23 skip** (1 failure in `node-assets.test.ts` reproduces identically on clean `main` in this environment — host artifact, unrelated)
- `bun run fmt:check` and `bun run lint` clean; `bun run build:lib` clean

### DIFF
--- a/packages/core/src/lib/outstanding-capability-probes.ts
+++ b/packages/core/src/lib/outstanding-capability-probes.ts
@@ -1,0 +1,67 @@
+import { isCapabilityResponse } from "./terminal-capability-detection.js"
+
+/**
+ * A terminal-capability mutation request minted by
+ * OutstandingCapabilityProbes.consume() — the only producer. The renderer
+ * applies capability changes exclusively through this value, so a mutation
+ * without an outstanding probe is unrepresentable at the API level.
+ */
+export interface CapabilityUpdate {
+  readonly sequence: string
+  readonly hasCursorReport: boolean
+  readonly hasStandardCapabilitySignature: boolean
+}
+
+/**
+ * Provenance gate for terminal capability mutations.
+ *
+ * The renderer issues its startup capability probes when it arms the
+ * capability window in setupTerminal(); the window expires with the probe
+ * timeout (or renderer teardown). While no probe is outstanding, consume()
+ * returns null for every byte sequence — including bytes that merely SHARE
+ * THE SHAPE of a capability response, such as modified F-key chords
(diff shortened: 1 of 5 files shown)


simonklee:
This may be a problem — but i'm not a fan of adding another system that maintains this. I think we can solve this with the existing design.

PR #1433, opened 2026-08-25
fix(react): wrap test renderer teardown in act

### LINKED ISSUE #1315: testRender emits React act warning for static components
## Reproduction

With `@opentui/react@0.4.5`, this Bun test renders a static component with no effects, timers, context, or asynchronous work:

```tsx
import React from \"react\";
import { describe, expect, it } from \"bun:test\";
import { testRender } from \"@opentui/react/test-utils\";

describe(\"static render\", () => {
    it(\"renders\", async () => {
        const setup = await testRender(<text>hello</text>, { width: 80, height: 24 });
        await setup.renderOnce();
        expect(setup.captureCharFrame()).toContain(\"hello\");
        setup.renderer.destroy();
    });
});
```

The test passes but prints:

```
An update to Root inside a test was not wrapped in act(...).
```

`testRender` already wraps `root.render(node)` and unmounting in React `act()`. Wrapping `setup.renderOnce()` in `act()`, forcing the initial mount through `flushSync`, and using a legacy root did not remove the warning.

## Expected

A static `testRender` test should complete without an `act()` warning. The test utility likely needs to coordinate its frame scheduler with React's test lifecycle.

manrods:
## Summary

- unmount the React test root inside `act()` when the CLI renderer emits `DESTROY`
- register the test cleanup before `createRoot` so its automatic cleanup becomes a no-op
- add a regression test for static `testRender` usage

## Root cause

`CliRenderer` emits `DESTROY` before invoking its configured `onDestroy` callback. `createRoot` therefore unmounted the React container outside `act()` before `testRender` reached its wrapped cleanup, producing the warning.

## Verification

- `cd packages/react && bun test tests/test-utils.test.tsx`
- `cd packages/react && bun test`
- `cd packages/react && bun run build --ci`
- formatting and lint checks

Closes #1315

### DIFF
--- a/packages/react/src/test-utils.ts
+++ b/packages/react/src/test-utils.ts
@@ -1,3 +1,4 @@
+import { CliRenderEvents } from "@opentui/core"
 import { createTestRenderer, type TestRendererOptions } from "@opentui/core/testing"
 import { act, type ReactNode } from "react"
 import { createRoot, type Root } from "./reconciler/renderer.js"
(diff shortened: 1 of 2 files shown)


simonklee:
Thanks, but prefer to fix these our self instead of the generated pr.

manrods:
Sure no problem!

PR #1436, opened 2026-08-25
fix(core): release the stdin handle on destroy

### LINKED ISSUE #1405: renderer.destroy() leaves the process.stdin handle it created registered on the event loop, wedging the legacy Windows console host
### Summary

`createCliRenderer()` touches `process.stdin`, which lazily instantiates the TTY read stream and registers its handle on the event loop. `destroy()` removes the data listener, clears raw mode and calls `pause()` — but never releases the handle. On the **legacy Windows console host** (`conhost.exe`), if a keypress was delivered during the session **and** the event loop turns between `destroy()` and process exit, the console host wedges and then dies, taking the window with it.

`pause()` stops the read; it does not remove the handle from the loop. That distinction is the whole bug.

### Environment

- `@opentui/core` **0.4.5** (measured) and **0.5.6** (confirmed), Windows 11, `conhost.exe powershell.exe`
- Reproduced on both Node 26.7.0 and Bun 1.3.14 — it is not runtime-specific
- Windows Terminal is **not** affected; this is legacy conhost only

The relevant code is **unchanged in 0.5.6**. `cleanupBeforeDestroy()` still does exactly `stdin.removeListener("data", …)` → `setRawMode(false)` → `stdin.pause()`, and neither `unref()` nor `destroy()` is ever called on the stdin stream on either version (`unref()` appears in 0.5.6 only on an internal poll timer). 0.5.6 additionally calls `stopTerminalKeepAlive()` and drains buffered input, but neither touches the handle registration.

### Reproduction

Run in a real `conhost.exe powershell.exe` window (not Windows Terminal):

```js
// repro.mjs

const renderer = await createCliRenderer({});

process.stderr.write("press a key, then wait\n");
await new Promise((resolve) => process.stdin.once("data", resolve));

renderer.destroy();

// let the event loop turn. no console I/O of any kind happens here.
await new Promise((resolve) => setTimeout(resolve, 3000));
process.exit(0);
```

The console window wedges roughly 0.6–0.8s after `destroy()` and closes itself shortly after. Remove either condition and it survives:

- don't press a key → survives
- replace the 3s wait with an immediate `process.exit(0)` → survives

Two notes for anyone re-running this on 0.5.6. On Node the native library loads through `node:ffi`, so it needs `--experimental-ffi` or `createCliRenderer` throws `OpenTUI native FFI is not available for this runtime yet`. And the `once("data")` above resolves on OpenTUI's own startup capability replies (CPR, XTVERSION, OSC 10/11), which conhost answers within ~50ms — my instrumented version ignores chunks beginning with `ESC` so that the keypress arm is genuinely exercised. Notably the capability reply alone is enough to arm the defect, which suggests the trigger is that a read completed at all, rather than that a human was involved.

### What the conditions are, precisely

Both are necessary, neither is sufficient. Twenty-plus controlled runs in a real conhost window:

| Condition                                                       | Result       |
| --------------------------------------------------------------- | ------------ |
| no keypress, loop turns 3s                                      | survives     |
| no keypress, renderer held open 10s                             | survives     |
| keypress, main thread blocked 3s with `Atomics.wait`, then exit | **survives** |
| keypress, loop allowed to turn                                  | **dies**     |
| keypress, loop turns, no console I/O at all after `destroy()`   | **dies**     |

The blocked-thread case is the informative one: 3 seconds of wall clock pass and the console lives, so the trigger is JS-scheduled work on the main loop, not elapsed time, not a worker, and not anything native.

### The handle

`process.stdin` does not exist before the OpenTUI import; `createCliRenderer` brings it into being.

| Phase                     | stdin                                        |
| ------------------------- | -------------------------------------------- |
| before import             | absent                                       |
| after `createCliRenderer` | present, `isRaw=true`                        |
| after `destroy()`         | present, `isRaw=false`, **still registered** |

Right after `destroy()`, active handles are `{"TTYWrap":3,"Timeout":1}` on 0.4.5, and `{"WriteStream":2,"TTY":3,"ReadStream":1}` on 0.5.6 under Node 26.7.0.

### Deciding control

Releasing the handle **before** `destroy()` fixes it; releasing it after does not.

| stdin handling                                     | On the loop at `destroy()` | Console      |
| -------------------------------------------------- | -------------------------- | ------------ |
| `setRawMode(false)` + `pause()` before `destroy()` | yes — `TTYWrap:3`          | dies         |
| full release **after** `destroy()`                 | yes                        | dies         |
| full release **before** `destroy()`                | no — `TTYWrap:2`           | **survives** |

Where "full release" is: remove listeners → `setRawMode(false)` → `pause()` → `unref()` → `destroy()`.

The same control reproduces on 0.5.6 **under Node**, with the census reading `{"WriteStream":2,"TTY":3,"ReadStream":1}` in the runs that die and `{"WriteStream":2,"TTY":2}` — `ReadStream` gone — in the run that survives.

**The workaround does not hold under Bun**, and that is the part I would most like your eyes on. On Bun 1.3.14 with 0.5.6, a run applying the full release before `destroy()` and then turning the loop for 3s still wedges the console, indistinguishably from the run that does nothing. Both runs took a genuine keypress (a 1-byte `0x71` chunk ~1.7s in, distinct from the ~50ms capability reply).

The release is not being ignored. Sampling `process.stdin` immediately before and after it:

| | before release | after release |
| --- | --- | --- |
| `isRaw` | `true` | `false` |
| `paused` | `false` | `true` |
| `destroyed` | `false` | **`true`** |
| `readable` | `true` | `false` |
| `data` listeners | 1 | 0 |

The stream is fully destroyed at the JS level and the console still dies. So on Bun, destroying `process.stdin` does not appear to detach whatever is actually reading the console — `process._getActiveHandles()` is stubbed on Bun and returns `{}`, so I can't see the handle directly to say more than that. Since Bun is the primary runtime here, this is probably the more consequential half of this report: the Node-side fix below is straightforward, and the Bun side may not be.

With the handle off the loop, the surviving run turns the loop for a full 3 seconds after `destroy()` — the exact configuration that kills the console in every other run — exits cleanly, and holds console mode `503` and code page `850` through a 15-second watch.

By the time `destroy()` has returned, intervening is already too late: the console is unresponsive ~0.1s in, and a `setRawMode(false)` issued there blocks for 0.7s. A parent process's `GetConsoleMode` on the same window blocks and then fails with **233 `ERROR_PIPE_NOT_CONNECTED`**.

### Things this is _not_

Ruled out by direct measurement, in case they look like candidates:

- **Not the console output code page.** A live sampler saw `outputCP` go 850 → 65001 → 850 across `destroy()` in runs where the console **survived**.
- **Not the shutdown VT blob.** Every surviving keypress run executes the same `performShutdownSequence()`.
- **Not raw mode.** A script that loads no OpenTUI at all reproduces console input mode `520` and survives.
- **Not a timer leak.** The one 5000ms one-shot OpenTUI creates is cleared by `destroy()`; a control clearing every remaining timer afterwards reported `cleared 0 timer(s)` and died anyway.
- **Not `FreeConsole` or a closed handle.** The shipped `@opentui/core-win32-x64@0.4.5` `opentui.dll` imports no `FreeConsole`, `AllocConsole`, `AttachConsole`, `GetStdHandle` or `SetStdHandle`, never opens `CONOUT$`/`CONIN$`, and has no native stdin path.

### Suggested fix

`AGENTS.md` already states the invariant this breaks — _"make native ownership explicit; clean up handles, callbacks, buffers, and listeners on every exit path."_ `destroy()` cleans up the listener but not the handle, and `process.stdin` is a handle the renderer itself brought into existence.

So: `destroy()` should release it rather than only pausing it — `unref()` at minimum, and only when the renderer instantiated it (i.e. `this.stdin === process.stdin` and no caller-supplied `config.stdin`). Releasing a handle the caller passed in would be wrong.

That is sufficient on Node. It is evidently **not** sufficient on Bun — a fully `destroyed` stdin stream still leaves the console to die there — so the Bun path likely needs something below the stream abstraction. You will know Bun's stdin internals far better than I do.

### Workaround for anyone hitting this

**On Node**, release stdin yourself, before calling `destroy()`:

```js
process.stdin.removeAllListeners("data");
if (process.stdin.setRawMode) process.stdin.setRawMode(false);
process.stdin.pause();
process.stdin.unref();
process.stdin.destroy();

renderer.destroy();
```

Ordering is load-bearing: the same five lines after `destroy()` do not help. I have no working workaround for Bun.

### On process, since CONTRIBUTING asks for PRs on bug fixes

I have a fix I am confident in for the Node path and none for Bun, so this is a report rather than a half-finished PR. If the Node-side change is worth having on its own I am glad to open it — say the word and I will.

`AGENTS.md` also asks for a focused regression test first. I do not think this one is automatable: it needs a real `conhost.exe` window, and both a pty master and ConPTY misreport it — the earlier investigation got false readings in **both** directions from ConPTY before switching to file-logged evidence inside a real console, which is why every number above comes from a log file rather than from scraping a terminal. Per the same guide's allowance for platforms that cannot be automated, the evidence and the manual verification steps are recorded above instead.

Both runtimes measured are the supported ones: Bun 1.3.14, and Node 26 with `--experimental-ffi` matching `scripts/node26.mjs`.

### Possibly the same root cause

Several downstream reports describe a terminal or parent shell dying on TUI exit on Windows. They pin `@opentui/core@0.4.5`:

- https://github.com/anomalyco/opencode/issues/30495 — Event Viewer shows `conhost.exe` faulting `0xc0000005`, then pwsh failing on `GetConsoleScreenBufferInfo` with `0xE9` (decimal **233**), matching the error above
- https://github.com/anomalyco/opencode/issues/28673 — user bisect landing on an `@opentui/core` version bump

Happy to run further controls on the same Windows setup if any of this would be more useful in a different form.

rohit-jsfreaky:
Fixes #1405.

## Summary

`createCliRenderer()` touches `process.stdin`, which lazily instantiates the TTY read stream and registers its handle on the event loop. `cleanupBeforeDestroy()` removes the data listener, clears raw mode and calls `pause()` — but `pause()` stops the read without releasing that registration, so the loop keeps holding it after `destroy()`. As the report puts it: pause() stops the read; it does not remove the handle from the loop.

`stdin.unref()` / `stdin.destroy()` appear nowhere in `packages/core/src` on current main.

## Measured

Node 26.7.0, Windows 11 (build 26200.9168), a real `conhost.exe` window, same script both times — create renderer, wait for a keypress, `destroy()`, let the loop turn 3s. `process.stdin` checked against `process._getActiveHandles()` immediately after `destroy()`:

| | stdin still on the event loop | outcome |
|---|---|---|
| main | `true` | survived |
| this branch | `false` | survived |

So the leaked registration is real and this removes it.

**What I did not reproduce:** the console wedge itself. Both runs survived the 3s window on this Windows build, so the conhost crash remains the reporter's observation rather than mine — what is fixed here is the leaked registration their report identifies as the cause. If someone can still reproduce the wedge, a run against this branch would be worth having.

One gotcha for anyone re-measuring: `process._getActiveHandles()` is a stub under Bun — it reports zero handles even with an active stdin listener — so this can only be measured on Node.

## Changes

`releaseStdinHandle()`, called from `cleanupBeforeDestroy()` right after `pause()`. Deliberately not called from `suspend()`, since `resume()` follows that.

Three conditions, all required, each for a case where releasing would be wrong:

- **`_ownsStdinHandle`** — captured in the constructor before anything touches stdin, true only when the stream is `process.stdin` *and* nothing was already consuming it. A caller-supplied `config.stdin`, or a stdin the host app was already reading, is not ours to unref. This follows the existing `this.stdin === process.stdin` pattern already used by `startTerminalKeepAlive`, `disableMouse` and `destroyRenderer`.
- **no listeners at destroy time** — so a listener the host attached during our lifetime and kept afterwards still holds the loop open.
- **`typeof unref === "function"`** — a redirected, file-backed stdin is a plain `ReadStream` with no `unref`; calling it there throws `TypeError`.

Worth stating explicitly: `unref()` persists. A consumer that attaches to stdin *after* `destroy()` will not be kept alive by it, because `resume()` does not re-ref. That is the state stdin was in before the renderer ran, which is why the ownership guard is scoped the way it is.

## Tests

Six in `packages/core/src/tests/renderer.stdin-handle.test.ts`. The two positive ones fail on main and pass here; the other four are guards that pass either way and exist so the release can never be widened by accident:

- releases the handle the renderer itself registered
- leaves a caller-supplied `config.stdin` alone
- leaves a stdin the host app was already consuming
- leaves stdin alone when a listener attached during the session survives
- does not throw when stdin has no `unref`
- `suspend()` does not release, because `resume()` follows it

`bun test` in `packages/core`: 5469 pass. The 37 failures are all pre-existing on a clean checkout of the same commit — 36 are "Unhandled error between tests" in the vendored Yoga suites under `src/zig/zig-pkg/`, and `getNodeAssets` fails on a missing `parser.worker.js` build artifact. I verified both by stashing this change and re-running them.

`bun run lint` clean; `tsc --noEmit` reports no errors in `renderer.ts`.

### DIFF
--- a/packages/core/src/renderer.ts
+++ b/packages/core/src/renderer.ts
@@ -997,6 +997,14 @@ export class CliRenderer extends EventEmitter implements RenderContext {
   // or duck-typed capability checks (e.g. `stdin.setRawMode?.()`).
   private readonly _usesProcessStdout: boolean
 
+  // Whether this renderer is what put process.stdin's handle on the event loop.
+  // Captured before any listener is attached: touching process.stdin lazily
+  // instantiates the TTY read stream and registers its handle, and `pause()` on
+  // destroy stops the read without releasing that registration (#1405). If the
+  // host app was already consuming stdin the handle is theirs, and destroy()
+  // must leave it exactly as it found it.
+  private readonly _ownsStdinHandle: boolean
+
   // Feed wiring. Non-null when the given stdout is not process.stdout and native
   // output is not explicitly redirected to a buffered memory destination.
   private _feed: NativeSpanFeed | null = null
@@ -1045,6 +1053,8 @@ export class CliRenderer extends EventEmitter implements RenderContext {
 
     this.stdin = stdin
     this.stdout = stdout
+    this._ownsStdinHandle =
+      stdin === process.stdin && stdin.listenerCount("data") === 0 && stdin.listenerCount("readable") === 0
     this._usesProcessStdout = stdout === process.stdout
     this.realStdoutWrite = stdout.write
 
@@ -1318,6 +1328,36 @@ export class CliRenderer extends EventEmitter implements RenderContext {
   private startTerminalKeepAlive(): void {
     if (this.stdin !== process.stdin || this.terminalKeepAliveTimer !== null) return
     this.terminalKeepAliveTimer = setInterval(() => {}, 60_000)
+  }
+
+  /**
+   * Put stdin's event-loop registration back the way this renderer found it.
+   *
+   * `pause()` stops the read but leaves the libuv handle registered, so after
+   * `destroy()` the loop still holds it. On the legacy Windows console host that
+   * wedges conhost once any read has completed and then takes the window with it
+   * (#1405); `suspend()` deliberately does not do this, because `resume()`
+   * follows it.
+   *
+   * Three conditions, all required:
+   *   - we are what registered the handle (`_ownsStdinHandle`) — a caller-supplied
+   *     `config.stdin`, or a stdin the host app was already consuming, is not ours
+   *     to unref;
+   *   - nothing else is consuming it now, so a listener attached during our
+   *     lifetime and kept afterwards still keeps the process alive;
+   *   - the stream actually implements `unref` — a redirected, file-backed stdin
+   *     is a plain ReadStream and does not.
+   */
+  private releaseStdinHandle(): void {
+    if (!this._ownsStdinHandle) return
+    if (this.stdin.listenerCount("data") > 0 || this.stdin.listenerCount("readable") > 0) return
+    const unref = (this.stdin as Partial<{ unref: () => void }>).unref
+    if (typeof unref !== "function") return
+    try {
+      unref.call(this.stdin)
+    } catch (e) {
+      console.error("Error releasing stdin during destroy:", e)
+    }
   }
 
   private stopTerminalKeepAlive(): void {
@@ -4353,6 +4393,7 @@ export class CliRenderer extends EventEmitter implements RenderContext {
     } catch (e) {
       console.error("Error pausing stdin during destroy:", e)
     }
+    this.releaseStdinHandle()
 
     if (this._feed !== null && this._splitHeight > 0 && !this._terminalIsSetup) {
       this.flushPendingSplitOutputBeforeTransition(false, { allowSuspended: true, allowUnsetup: true })
--- a/packages/core/src/tests/renderer.stdin-handle.test.ts
+++ b/packages/core/src/tests/renderer.stdin-handle.test.ts
@@ -1,0 +1,128 @@
+import { test, expect, afterEach } from "bun:test"
+import { Readable } from "stream"
+import { createCliRenderer, type CliRenderer } from "../renderer.js"
+import { createTestStdout } from "../testing/test-streams.js"
+
+// #1405: createCliRenderer touches process.stdin, which lazily instantiates the
+// TTY read stream and registers its handle on the event loop. destroy() removed
+// the listener, cleared raw mode and paused — but pause() stops the read without
+// releasing the registration, so the loop kept holding it. On the legacy Windows
+// console host that wedges conhost shortly after destroy() and takes the window
+// with it. These cover the release and, just as importantly, every case where
+// releasing would be wrong.
+
+class FakeStdin extends Readable {
+  public unrefCalls = 0
+  public rawMode: boolean | null = null
+  constructor() {
+    super({ read() {} })
+  }
+  unref(): void {
+    this.unrefCalls += 1
+  }
+  setRawMode(mode: boolean): this {
+    this.rawMode = mode
+    return this
+  }
+}
+
+/** A stdin with no `unref` at all — what a redirected, file-backed stdin looks like. */
+class UnrefLessStdin extends Readable {
+  constructor() {
+    super({ read() {} })
+  }
+  setRawMode(mode: boolean): this {
+    return this
+  }
+}
+
+const destroyFns: Array<() => void> = []
+let restoreStdin: (() => void) | null = null
+
+afterEach(() => {
+  while (destroyFns.length) destroyFns.pop()!()
+  restoreStdin?.()
+  restoreStdin = null
+})
+
+/** Install `stream` as process.stdin so the renderer's default path picks it up. */
+function useAsProcessStdin(stream: Readable): void {
+  const original = Object.getOwnPropertyDescriptor(process, "stdin")!
+  Object.defineProperty(process, "stdin", { value: stream, configurable: true })
+  restoreStdin = () => Object.defineProperty(process, "stdin", original)
+}
+
+async function makeRenderer(config: Record<string, unknown> = {}): Promise<CliRenderer> {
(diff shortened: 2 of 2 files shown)


simonklee:
Closing since you couldn't repro the original issue.

PR #1437, opened 2026-08-25
fix(native): ship stripped libraries with separate release symbols

kommander:
Build optimized native libraries once with debug information, then detach matching symbols before packaging the stripped npm copies. Keep Debug builds unchanged.

Archive Linux `.debug`, macOS `.dSYM`, and Windows `.pdb` files as separate GitHub Release assets with version, target/ABI, commit, binary identity, and checksums. CI verifies symbol matching for all eight targets and tests the distributed libraries.

Linux locally verified: approximately 26 MB → 6 MB, preserved exports, matching build ID/debuglink, and source-line resolution. See `packages/web/src/content/docs/test-and-debug/native-crashes.mdx` for usage and retention details.


simonklee:
The standard name is **separate debug symbols**, or **detached debug information**. I’d use that for OpenTUI: **ship stripped binaries through npm, retain the matching symbols as separate release assets.**

**How It Works**
1. Build the optimized release **once, with debug information**.
2. Extract or collect its debug symbols.
3. Strip the distributed binary, preserving required exports and runtime unwinding information.
4. Publish the small binary to npm and archive the symbols for later debugging.

**Debug symbols are not the same as a `Debug` build.** An optimized `ReleaseFast` build can have symbols. A separately compiled `Debug` binary has different machine code and addresses, so its symbols cannot explain a crash from the release binary. It can also behave differently enough to hide the bug.

With matching symbols, a debugger can turn crash addresses into function names and source locations. Optimized-out variables will still be unavailable.

**Platform Formats**
| Platform | Separate Symbols | How They Match |
|---|---|---|
| Linux | `.debug` file, extracted with `objcopy --only-keep-debug` | Build ID and/or `.gnu_debuglink` |
| macOS | `.dSYM` bundle, normally generated with `dsymutil` | Mach-O UUID |
| Windows | `.pdb` file from the compiler/linker | PDB signature, typically GUID and age |

These are established workflows: [GDB separate debug files](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Separate-Debug-Files.html), [Apple debug symbols](https://developer.apple.com/documentation/xcode/building-your-app-to-include-debugging-information), and [Microsoft symbols](https://learn.microsoft.com/en-us/windows/win32/dxtecharts/debugging-with-symbols).

**For OpenTUI**
I would adjust [PR #1437](https://github.com/anomalyco/opentui/pull/1437) rather than abandon stripping:

- **Preserve symbols before stripping.** The PR sets `.strip = optimize != .Debug` during the Zig build, without retaining release symbols. Instead, build with symbols and strip the distribution copy afterward.
- **Do this in the native build job, not the final npm publish job.** All targets are already built on macOS in `.github/workflows/build-native.yml:57`. That provides a suitable place to collect symbols before packaging and Windows signing. Generate macOS dSYMs while the object files are still available.
- **Keep symbols outside npm.** Separate GitHub Release downloads are sufficient initially. No debug npm packages or symbol server are necessary.
- **Retain them for the lifetime of the release.** Current CI artifacts expire after 30 days (`.github/workflows/build-native.yml:335`). Record the version, target/ABI, commit, and binary identity alongside the symbols. The local Linux library I inspected has no build ID, so I would enable one.
- **Leave local `Debug` builds unchanged.** They remain useful for development, independently of release-symbol retention.

Symbols do not capture a crash by themselves: an issue still needs a core/minidump or a native trace containing addresses and enough module information to resolve them.

**Checked Locally**
I tested separation on a temporary copy of the existing Linux library:

- Original: **26.3 MB**
- Stripped: **6.2 MB**
- Separate symbols: **20.0 MB**

The exported dynamic symbols were unchanged, and GDB automatically loaded the separate file and resolved `getBuildOptions` to `lib.zig:1117`.

PR #1442, opened 2026-08-26
markdown: fix hyperlink rendering and detection

### LINKED ISSUE #1439: Make Markdown hyperlinks clickable, compact, and usable with mouse capture
OpenTUI already supports OSC 8 hyperlinks. The remaining problems are:

1. Markdown link labels do not consistently retain their URL.
2. Concealed links show unnecessary or duplicated URLs.
3. Bare HTTP(S) URLs are not consistently clickable.
4. Mouse capture can prevent terminal-native link activation, and applications
   cannot look up the URL under the pointer.

For this Markdown:

```markdown
[OpenTUI](https://github.com/anomalyco/opentui)
https://example.com/path
```

A terminal with hyperlink support should display:

```text
OpenTUI
https://example.com/path
```

Both lines should carry the correct hyperlink. Without hyperlink support, the
named link should remain readable as:

```text
OpenTUI (https://github.com/anomalyco/opentui)
https://example.com/path
```

The bare URL must appear exactly once in both cases.

## Implementation

- [ ] Preserve the destination on every visible Markdown link-label cell in both
      the tree-sitter/conceal path and the inline Markdown path.
- [ ] Handle single-character labels, formatted labels, table cells, and links
      wrapped across multiple rows.
- [ ] Detect bare `http://` and `https://` URLs in Markdown prose and tables. Do
      not link URLs inside inline or fenced code.
- [ ] When concealment and OSC 8 are enabled, render `[label](url)` as `label`
      with the URL attached as hyperlink metadata.
- [ ] When OSC 8 is unavailable, render named links as `label (url)` so the
      destination remains visible.
- [ ] Render bare URLs once regardless of terminal capability.
- [ ] Refresh affected content when hyperlink capability is detected after the
      initial render.
- [ ] Expose a small public lookup such as `renderer.getLinkAt(x, y): string | null`
      so applications can handle clicks while `useMouse` remains enabled.
- [ ] Reuse the link ID already stored in rendered cell attributes and the
      existing native URL lookup instead of introducing another link representation.
- [ ] Add tests that verify link metadata or emitted OSC 8 sequences, not just
      displayed text.
- [ ] Verify terminal-native activation in Windows Terminal and
      application-handled activation with mouse capture in Ghostty on macOS.

## Existing Code

- `packages/core/src/renderables/Markdown.ts`: inline Markdown rendering and
  concealed link output.
- `packages/core/src/lib/detect-links.ts`: tree-sitter link metadata. Currently
  skips single-character chunks.
- `packages/core/src/renderer.ts`: existing renderable hit testing.
- `packages/core/src/utils.ts`: extracts hyperlink IDs from cell attributes.
- `packages/core/src/zig.ts`: existing `linkGetUrl` binding.

## Out Of Scope

OpenTUI should not open URLs, add scheme allowlists, implement `file://` reveal
behavior, or define application security policy. URL handling belongs to the
terminal, operating system, or application.

Applications decide which links to create and how to handle a URL returned by
cell lookup. OpenTUI cannot intercept a link click handled directly by the
terminal; application handling is only possible when the terminal forwards the
mouse event.

## Related Issues

- #1435: concealed named links display their URL even when OSC 8 is available.
- #1380: mouse capture prevents native link activation without application-level link lookup.
- #1196: bare URLs are not consistently clickable.
- #1050: bare URLs are duplicated as `url (url)`.
- #869: concealed Markdown labels are not clickable.

simonklee:
Make named and bare links clickable without exposing their destinations
when the terminal supports OSC 8. Preserve visible URL fallbacks
elsewhere and keep links intact across formatting, streaming, wrapping,
and late terminal capability detection.

Fix #1439

### DIFF
--- a/packages/core/src/lib/detect-links.test.ts
+++ b/packages/core/src/lib/detect-links.test.ts
@@ -95,4 +95,341 @@ describe("detectLinks", () => {
     // The label chunk should also get the link
     expect(result.find((c) => c.text === "Click here")!.link).toEqual({ url: "https://example.com" })
   })
+
+  test("links formatted and source-replaced label chunks across intervening conceal captures", () => {
+    const content = "[a **A&amp;B** z](https://example.com)"
+    const labelEnd = content.indexOf("]")
+    const entityStart = content.indexOf("&amp;")
+    const urlStart = content.indexOf("https://")
+    const highlights: SimpleHighlight[] = [
+      [1, labelEnd, "markup.link.label"],
+      [3, labelEnd - 2, "markup.strong"],
+      [3, 5, "conceal", { conceal: "" }],
+      [entityStart, entityStart + 5, "character.special", { conceal: "&" }],
+      [labelEnd - 4, labelEnd - 2, "conceal", { conceal: "" }],
+      [labelEnd, labelEnd + 1, "conceal", { conceal: " " }],
+      [urlStart, content.length - 1, "markup.link.url"],
+    ]
+    const chunks = ["a ", "A", "&", "B", " z", " ", "(", "https://example.com"].map(chunk)
+    const sourceRanges = [
+      { start: 1, end: 3 },
+      { start: 5, end: entityStart },
+      { start: entityStart, end: entityStart + 5 },
+      { start: entityStart + 5, end: entityStart + 6 },
+      { start: labelEnd - 2, end: labelEnd },
+      { start: labelEnd, end: labelEnd + 1 },
+      { start: labelEnd + 1, end: labelEnd + 2 },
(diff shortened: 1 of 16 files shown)


kommander:
Blocker: links lose OSC 8 metadata after preceding multibyte text because Tree-sitter byte offsets are treated as JS string offsets. Please add named + bare Unicode regression coverage.

simonklee:
> Blocker: links lose OSC 8 metadata after preceding multibyte text because Tree-sitter byte offsets are treated as JS string offsets. Please add named + bare Unicode regression coverage.

👍🏻

PR #1445, opened 2026-08-26
feat: FreeBSD support for the native library

apbodrov:
## What this does

Makes `@opentui/core` build and load its native library on FreeBSD. Four small changes, none of them in the Zig or C/C++ sources — the renderer itself already compiles cleanly for this target.

## Why it does not work today

`zig build` refuses to build for the host at all:

```
No matching supported target for native platform (x86_64-freebsd)
error: UnsupportedNativeTarget
```

`buildNativeTarget()` matches the host against `SUPPORTED_TARGETS` and errors out when nothing matches. Even with a library in hand, `getNativeAssetDescriptor()` then throws `Unsupported OpenTUI Node asset target: freebsd-x64` before any lookup happens — including before `OTUI_ASSET_ROOT` is consulted, so there is no escape hatch for a locally built `.so` either.

## The changes

1. **`packages/native/build.zig`** — FreeBSD entries in `SUPPORTED_TARGETS`, plus a `.freebsd` arm in `linkSystemLibraries()`. It deliberately omits the `dl` that the Linux arm links: FreeBSD keeps `dlopen(3)` and friends in libc and has no separate `libdl` to link against. `m` and `pthread` are linked as on Linux.

2. **`packages/core/src/node-asset-target.ts`** — `freebsd: "libopentui.so"` in `NATIVE_FILE_NAMES`, and `"freebsd"` in the `NodeAssetTarget["platform"]` union.

3. **`packages/core/src/platform/runtime-assets.bun.ts`** — a `freebsd` arm in `resolveNativeLibraryPath()`. The **node** resolver needs no change: it imports `asset.packageName` dynamically, so it works as soon as the descriptor resolves. Only the bun path hardcodes one branch per platform.

`ghosttyVtAvailable()` is intentionally untouched — FreeBSD falls through its `else` arm, so ghostty-vt is compiled out. I did not investigate what, if anything, that costs at runtime; the TUI I tested rendered correctly without it.

## Verification

FreeBSD 15.1-RC2, amd64, zig 0.16.0:

- `sh scripts/prepare-zig-deps.sh` then `zig build -Doptimize=ReleaseFast` exits 0 and produces `lib/x86_64-freebsd/libopentui.so` (20 MB, 1829 exported text symbols).
- Importing `@opentui/core` resolves to that library through the new arm.
- End-to-end rendering was verified against the **0.4.5** tree rather than main, because that is what the application I was porting pins. The same two `build.zig` edits apply there (the sources live at `packages/core/src/zig`, and that tree needs zig 0.15.2 — 0.16 fails on a vendored uucode `build.zig` with `std.Build` API errors). With that library in place a full-screen TUI renders correctly under a real tty: logo, input box, status bar, the lot.

`aarch64-freebsd` is listed in the table but was **not** built or tested — I included it so an arm64 FreeBSD user does not need a second patch, but treat it as unverified and drop that line if you would rather not carry an untested target.

## What this PR does not do

Publishing `@opentui/core-freebsd-x64` to npm and adding a FreeBSD job to `build-native.yml` are yours to decide — I have not touched the workflows. Without a published platform package, FreeBSD users can still point `OTUI_ASSET_ROOT` at a locally built library, which now works because change 2 stops the descriptor from throwing first.

### DIFF
--- a/packages/core/src/node-asset-target.ts
+++ b/packages/core/src/node-asset-target.ts
@@ -1,5 +1,5 @@
 export type NodeAssetTarget = {
-  readonly platform: "darwin" | "linux" | "win32"
+  readonly platform: "darwin" | "linux" | "win32" | "freebsd"
   readonly arch: "arm64" | "x64"
   readonly libc?: "glibc" | "musl"
 }
@@ -14,6 +14,7 @@ const NATIVE_FILE_NAMES = {
   darwin: "libopentui.dylib",
   linux: "libopentui.so",
   win32: "opentui.dll",
+  freebsd: "libopentui.so",
 } as const
 
 export function getNativeAssetDescriptor(target: NodeAssetTarget): NativeAssetDescriptor {
--- a/packages/core/src/platform/runtime-assets.bun.ts
+++ b/packages/core/src/platform/runtime-assets.bun.ts
@@ -68,6 +68,14 @@ export async function resolveNativeLibraryPath(): Promise<string> {
     }
   }
 
+  if (process.platform === "freebsd") {
+    if (process.arch === "x64") {
+      return (await import("@opentui/core-freebsd-x64")).default
+    }
+    if (process.arch === "arm64") {
+      return (await import("@opentui/core-freebsd-arm64")).default
+    }
+  }
   if (process.platform === "win32") {
     if (process.arch === "x64") {
       return ((await import("@opentui/core-win32-x64" as string)) as NativePackageModule).default
--- a/packages/core/src/platform/runtime-assets.bun.ts
+++ b/packages/core/src/platform/runtime-assets.bun.ts
@@ -69,13 +69,15 @@ export async function resolveNativeLibraryPath(): Promise<string> {
   }
 
   if (process.platform === "freebsd") {
-    if (process.arch === "x64") {
-      return (await import("@opentui/core-freebsd-x64")).default
-    }
-    if (process.arch === "arm64") {
-      return (await import("@opentui/core-freebsd-arm64")).default
-    }
+    // Built at runtime rather than written as a literal: no @opentui/core-freebsd-*
+    // package is published yet, and a literal specifier would make the bundler try to
+    // resolve it at build time, failing `bun run build:lib` on every platform. The node
+    // resolver already imports asset.packageName the same way. Set OTUI_ASSET_ROOT to
+    // point at a locally built libopentui.so instead (handled above).
+    const specifier: string = `@opentui/core-freebsd-${process.arch}`
+    return ((await import(specifier)) as NativePackageModule).default
   }
+
   if (process.platform === "win32") {
     if (process.arch === "x64") {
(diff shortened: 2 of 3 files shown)


apbodrov:
Self-correcting the first push: it broke CI, and it deserved to.

I had written the FreeBSD arm with a literal specifier, the same shape as the existing arms:

```ts
return (await import("@opentui/core-freebsd-x64")).default
```

The neighbouring literals resolve because those packages are published — mine does not exist, so `bun run build:lib` failed to resolve it at bundle time, on every platform:

```
error: Could not resolve: "@opentui/core-freebsd-x64".
    at packages/core/src/platform/runtime-assets.bun.ts:73:28
```

Everything else that went red followed from `Build core` failing.

Now fixed by building the specifier from `process.arch` so it stays a runtime import — which is exactly why the node resolver never had this problem, since it imports `asset.packageName` through a variable. The published-platform arms keep their literals.

Verified by running the step that failed, `bun scripts/build.ts --lib`: it bundles clean, no resolution error. (My run stops a step later on `Executable not found in $PATH: "bunx"`, which is my box lacking bunx, not something this diff touches.)

Worth noting the practical consequence for FreeBSD users either way: with no published platform package, the import throws at runtime and `OTUI_ASSET_ROOT` is the supported route to a locally built `libopentui.so`. That path works only because of the `node-asset-target.ts` change in this PR — without it the descriptor throws before `OTUI_ASSET_ROOT` is ever read.

apbodrov:
aarch64 update — the PR text said `aarch64-freebsd` was listed but untested. It is tested now, so here is what that line is worth.

**Cross-compiled from amd64** (this tree, main, zig 0.16.0):

```
zig build -Dlibrary-target=aarch64-freebsd -Doptimize=ReleaseFast   →  exit 0
lib/aarch64-freebsd/libopentui.so: ELF 64-bit LSB shared object, ARM aarch64
```

**Built natively on FreeBSD 14.4-RELEASE-p5 arm64**, against the 0.4.5 tree with the same two `build.zig` edits (zig 0.15.2, as that tree requires):

```
zig build -Doptimize=ReleaseFast   →  exit 0
lib/aarch64-freebsd/libopentui.so  →  ELF 64-bit LSB shared object, ARM aarch64, 14 MB
nm -D | grep -c ' T '              →  1756 exported text symbols
```

And it loads — `dlopen(..., RTLD_NOW)` succeeds and `dlsym` resolves `createRenderer`:

```
DLOPEN_OK sym=0x841234a4
```

So both FreeBSD targets in `SUPPORTED_TARGETS` now have a build behind them, on both architectures, and the arm64 one is a native build rather than only a cross.

One environment note in case you add a FreeBSD CI job: a stock FreeBSD image has no CA bundle, and `zig fetch` fails on the `build.zig.zon` dependencies with `CertificateBundleLoadFailure` until `pkg install ca_root_nss`. Nothing to do with this patch, but it is the first thing that bites on a fresh box.

Still unverified on arm64: end-to-end TUI rendering. That was checked on amd64 (full screen under a real tty); on arm64 I only got as far as loading the library and resolving symbols.

PR #1451, opened 2026-08-26
fix(core): limit embedded terminal selection highlights to text

jlongster:
## Summary

- Clip embedded-terminal selection painting to each row's occupied text extent instead of highlighting unused trailing cells and empty rows.
- Preserve written spaces and interior cursor-created gaps; highlight both display cells of wide characters when either cell is selected.
- Delay cell-drag selection in EmbeddedTerminalRenderable until the pointer leaves its starting cell. Once activated, keep normal behavior for the rest of the drag, including returning to the anchor or selecting just one text cell.
- Leave shared renderer selection policy, word/line gesture handling, mouse-up clearing, and clipboard extraction unchanged. A click that never crosses the threshold exposes no selected terminal text. Written trailing spaces remain highlightable even though copying trims them.
- Add native coverage for text-only painting and mouse-event regression tests for same-cell presses/movement, release without dragging, horizontal/vertical activation, returning to the anchor, and resetting on a new gesture.

## Verification

- Reproduced both regressions with failing tests before their respective fixes.
- Full native suite: 2,067 passed, 8 skipped.
- Latest core JavaScript suite after rebuilding: 5,464 passed, 23 skipped.
- Focused embedded-terminal and shared-renderer selection tests: 28 passed.
- Full OpenTUI build for the native change, core library/declaration rebuild for the gesture change, formatting checks, lint, and Zig formatting passed. Native verification used Zig 0.16.0.
- Checked wide/narrow text-only painting and press/drag/return-to-anchor rendering with the production embedded-terminal component on Linux.
- Linked the local build into a fresh OpenCode v2 worktree and smoke-tested the TUI. Source-link typechecking reports existing errors in unchanged OpenTUI TypeScript files; configured external plugins also report failures. macOS was not tested.

### DIFF
--- a/packages/core/src/renderables/EmbeddedTerminal.test.ts
+++ b/packages/core/src/renderables/EmbeddedTerminal.test.ts
@@ -87,6 +87,88 @@ describe("EmbeddedTerminalRenderable", () => {
     expect(terminal.getSelectedText()).toBe("")
   })
 
+  test("waits for a drag to leave its starting cell before selecting", async () => {
+    const terminal = new EmbeddedTerminalRenderable(setup.renderer, { width: 20, height: 4 })
+    setup.renderer.root.add(terminal)
+    terminal.write("hello")
+    await setup.renderOnce()
+    const original = setup.captureSpans().lines
+
+    await setup.mockMouse.pressDown(1, 0)
+    await setup.renderOnce()
+    expect(terminal.hasSelection()).toBe(false)
+    expect(setup.captureSpans().lines).toEqual(original)
+
+    await setup.mockMouse.moveTo(1, 0)
+    await setup.renderOnce()
+    expect(terminal.hasSelection()).toBe(false)
+    expect(setup.captureSpans().lines).toEqual(original)
+    await setup.mockMouse.release(1, 0)
+    await setup.renderOnce()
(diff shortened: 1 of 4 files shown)


kommander:
Reviewed `92030599`: no behavior regression reproduced in differential mouse/painting tests. A modest native repaint slowdown is measurable:

| 120×40 workload | Before → PR | Paired slowdown |
|---|---:|---:|
| Dense selected redraw | 115.08 → 126.91 µs | +10.2% |
| Full-selection movement | 114.58 → 127.42 µs | +11.0% |
| Dense unselected redraw | 114.10 → 122.18 µs | +7.1% |

Evidence: same main (`7a0dca8d`) with/without the exact compositor patch; Linux/EPYC, Zig 0.16 ReleaseFast, 16 alternating CPU-pinned pairs using public `write/setSelection/invalidate/compose`. Every output cell validated outside timing; these cases have identical output. Selected-redraw 95% CI: **+8.2–12.2%**; excluding compiler-overlap pairs retains +10.1%.

Cached compose shows no proven slowdown. Unselected redraw also slows, so the reverse scan alone does not explain it. This is ~12 µs extra per full selected repaint, not demonstrated application-level latency.

PR #1455, opened 2026-08-27
perf(native): reduce word rewrap work without expanding caches

kommander:
Short Unicode chunks rescan on resize, and word layout also freed/rebuilt its arena each time. Reuse word-layout capacity while text is unchanged, process ASCII prefixes/tails in the existing scanner, and append fitting words/separators together; break-cache policy and output stay unchanged.

| 945 KB Unicode word-wrap | v0.5.8 | Main | Branch |
|---|---:|---:|---:|
| First layout | 11.63 ms | 7.55 ms | 5.29 ms |
| First measurement | 11.11 ms | 5.51 ms | 3.69 ms |
| Width 79/80 rewrap | 2.62 ms | 7.38 ms | 4.18 ms |
| Native retained | 22.77 MiB | 14.52 MiB | 14.52 MiB |

Six balanced fresh-process triples, CPU-pinned Zig 0.16 ReleaseFast/page allocator, identical input and verified layout/rendered output. Retained bytes include TextBuffer/View arena capacity; 40 resizes reduce allocation traffic from 376 to 24 MiB, with zero after destruction. Rewrap is 44.5% faster than main but still 1.6x slower than v0.5.8; expanding metadata caching was rejected because it increased retained/edit-history memory.

### DIFF
--- a/packages/native/src/tests/text-buffer-segment_test.zig
+++ b/packages/native/src/tests/text-buffer-segment_test.zig
@@ -96,6 +96,23 @@ test "walkChunkLayoutInfo keeps Prepend joined to an ASCII vector" {
     try testing.expectEqual(utf8.LayoutWrapBreakKind.preserved_whitespace, breaks.items[0].kind);
 }
 
+test "walkChunkLayoutInfo ASCII prefixes keep their final grapheme intact" {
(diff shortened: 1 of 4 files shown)


kommander:
Superseded by https://github.com/anomalyco/opentui/pull/1456, which now targets main and includes both this optimization and view-owned word metadata reuse. The combined PR remains a draft.

PR #1460, opened 2026-08-28
fix(native): keep capability replies in tmux pane

### LINKED ISSUE #1459: tmux capability replies leak into another pane after focus switch
## Description

OpenTUI sends response-generating terminal capability probes through tmux DCS passthrough during startup. If the user switches panes before the outer terminal replies, tmux can deliver those replies to the newly active pane instead of the pane that sent the query.

The receiving shell then displays or executes fragments such as:

```text
1+r4D73=...
2027;1$y
2031;1$y
1004;1$y
2004;1$y
2026;1$y
Gi=31337;OK
62;22;52c
```

These correspond to OpenTUI's XTGETTCAP `Ms`, DECRQM, Kitty graphics, and primary device attributes startup probes.

This is different from #334: #415 fixed the framing that caused the Kitty graphics query itself to affect the pane title, but DCS passthrough does not associate a later terminal reply with the pane that originated the query.

## Reproduction

Environment used:

- Ghostty 1.3.1
- tmux 3.7c
- OpenCode 1.18.23
- macOS

1. Enable tmux passthrough with `tmux set -g allow-passthrough on`.
2. Create two tmux panes, A and B.
3. Start OpenCode, or another OpenTUI application, in pane A.
4. Immediately switch to pane B while startup capability detection is running.
5. Press a key in pane B.
6. Observe terminal response fragments in pane B's shell input or output.

The race is timing-sensitive with a local terminal. It can be made deterministic by delaying the outer terminal's replies until after `select-pane`.

## Root cause

`Terminal.queryTerminalSend()` emits `capabilityQueriesTmux`, `kittyGraphicsQueryTmux`, and `primaryDeviceAttrsTmux`. The wrapper transports each query from pane A to the outer terminal, but the corresponding response has no tmux pane identity:

```text
pane A -> tmux passthrough -> outer terminal
                                  |
pane B <- tmux client input <-----+
```

OpenTUI already documents this constraint in `queryThemeColors()`:

> DCS passthrough replies are not routed back to the pane that asked.

The same constraint applies to the other response-generating startup probes.

## Proposed behavior

When running inside tmux:

- Do not send response-generating capability probes through DCS passthrough.
- Send probes that tmux can answer directly without wrapping.
- Skip the outer Kitty graphics probe, falling back conservatively when the complete OpenTUI -> tmux -> terminal path cannot be detected safely.
- Do not retry the probes through passthrough after late XTVERSION-based tmux detection.

One-way passthrough operations are outside this issue.

I can submit a focused PR with native regression tests if this direction is acceptable.

## Expected behavior

Starting an OpenTUI application in one tmux pane must not inject terminal response bytes into another pane after focus changes.

clang-engineer:
Closes #1459

## Summary

- Keep response-generating capability probes inside tmux's virtual terminal instead of sending them through DCS passthrough.
- Skip automatic Kitty graphics probing under tmux while retaining tmux-local primary device attributes detection.
- Do not retry startup probes through passthrough after late XTVERSION-based tmux detection.
- Ignore unowned Kitty graphics replies once tmux is known.

## Problem

#415 fixed the framing of the Kitty graphics query from #334, but DCS passthrough only transports output to the outer terminal. A later reply carries no identity for the pane that sent the query.

If a user switches panes before Ghostty replies, tmux may deliver XTGETTCAP, DECRQM, Kitty graphics, or DA response bytes to the newly active shell pane.

OpenTUI already applies the same constraint to theme color queries: DCS passthrough replies are not routed back to the pane that asked.

## Approach

tmux directly handles the capability and DA queries used here and writes replies to the originating pane's PTY. This change sends those queries unwrapped and treats unsupported capabilities conservatively.

The Kitty graphics query is omitted because tmux cannot safely associate its outer-terminal response with the requesting pane. Automatic image protocol selection already falls back under tmux, while an explicit image protocol override remains available.

## Verification

- `bun run test:native`: 2088 passed, 32 skipped
- `bun run build`
- `bun run fmt:check`
- `bun run lint`

The regression tests verify known and late-detected tmux paths, absence of DCS passthrough and Kitty probes, retention of direct tmux-local queries, and rejection of unowned Kitty replies.

### DIFF
--- a/packages/native/src/terminal.zig
+++ b/packages/native/src/terminal.zig
@@ -296,9 +296,10 @@ pub fn exitAltScreen(self: *Terminal, tty: anytype) !void {
 
 pub fn queryTerminalSend(self: *Terminal, tty: anytype) !void {
     self.checkEnvironmentOverrides();
+    const is_tmux = self.isInTmux();
     self.unicode_wide_locked = self.caps.unicode == .unicode_wide;
-    self.graphics_query_pending = !self.skip_graphics_query;
-    self.sixel_query_pending = !self.skip_graphics_query;
+    self.graphics_query_pending = !self.skip_graphics_query and !is_tmux;
+    self.sixel_query_pending = !self.skip_graphics_query and !is_tmux;
     self.capability_queries_pending = false;
     self.startup_cursor_query_pending = true;
     self.startup_cursor_query_captured = false;
@@ -319,11 +320,11 @@ pub fn queryTerminalSend(self: *Terminal, tty: anytype) !void {
     // Capture the current cursor position before temporary home-position queries.
     try tty.writeAll(ansi.ANSI.cursorPositionRequest);
 
-    if (self.isInTmux()) {
+    if (is_tmux) {
         if (self.is_foot) {
-            try tty.writeAll(ansi.ANSI.capabilityQueriesFootIsBrokenTmux);
+            try tty.writeAll(ansi.ANSI.capabilityQueriesFootIsBroken);
         } else {
-            try tty.writeAll(ansi.ANSI.capabilityQueriesTmux);
+            try tty.writeAll(ansi.ANSI.capabilityQueries);
         }
     } else {
         if (self.is_foot) {
@@ -335,9 +336,10 @@ pub fn queryTerminalSend(self: *Terminal, tty: anytype) !void {
     }
 
     if (!self.skip_graphics_query) {
-        if (self.isInTmux()) {
-            try tty.writeAll(ansi.ANSI.kittyGraphicsQueryTmux);
-            try tty.writeAll(ansi.ANSI.primaryDeviceAttrsTmux);
+        if (is_tmux) {
+            // tmux can answer DA for its virtual terminal. A passthrough Kitty
+            // reply has no pane ownership and may reach a different active pane.
+            try tty.writeAll(ansi.ANSI.primaryDeviceAttrs);
         } else {
             try tty.writeAll(ansi.ANSI.kittyGraphicsQuery);
             try tty.writeAll(ansi.ANSI.primaryDeviceAttrs);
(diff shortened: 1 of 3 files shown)


simonklee:
I had gpt look into this.

The reply-routing problem is real, but the new query changes what the Sixel capability describes. The unwrapped device-attributes query asks tmux about its own Sixel support, not the outer terminal’s support. Tmux answers according to its build configuration.
However, `parseSixelDeviceAttributes()` still sets `caps.sixel` from that answer. `refusesForcedSixel()` then treats it as permission to use Sixel. The image output still uses tmux passthrough, so it bypasses the Sixel implementation that answered the query.
Consider startup with tmux already detected, passthrough enabled, and usable pixel dimensions. For an explicit Sixel override:
- A tmux build with Sixel support can authorize Sixel output to an outer terminal that does not support it.
- A tmux build without Sixel support can prevent Sixel output to an outer terminal that does support it.
The automatic block fallback does not protect explicit overrides.
A direct probe against tmux 3.7b returned Sixel support without any outer-terminal reply. I checked that response and the selection logic, but did not test complete image output before and after this patch.
Please reconcile the new query source with `refusesForcedSixel()` and the passthrough output path. Add tests that vary tmux and outer-terminal Sixel support independently. Cover explicit overrides and both known and late tmux detection. The current query-byte assertions do not catch this mismatch.

clang-engineer:
@simonklee 

Thanks for catching this. Updated in 57bd9b3.

The tmux-local DA response no longer sets caps.sixel, including during late tmux detection, so it cannot select Sixel automatically for the outer passthrough endpoint.

Explicit Sixel overrides remain authoritative under tmux and no longer depend on tmux's own Sixel build support.

Added coverage for:
- known tmux with and without Sixel in its DA response
- late tmux detection after a Sixel DA response
- explicit Sixel output through tmux passthrough

Verification:
- tmux-filtered native tests: 26 passed
- bun run build
- bun run fmt:check
- bun run lint

The full native suite reached 2112 passed / 8 skipped. Two unrelated X11 clipboard listener tests could not create sockets in the sandbox (EPERM).

GrimOutlook (approved):
Any ETA on when this might be merged?

PR #1462, opened 2026-08-28
fix(core): bound large diff rendering work

kitlangton:
## Why

A large diff can occupy tens of thousands of layout rows while only a few rows reach the destination buffer. Line-number gutters currently allocate and inspect document-height data during paint, and applying syntax chunks scans every logical line for every chunk. The first cost makes scrolling depend on offscreen content; the second makes multiline syntax application quadratic and can block input after the initial plaintext frame.

## What Changes

Keep the complete text, syntax, native layout, wrapping, selection owners, and scroll extent while bounding work at the two expensive boundaries.

| Boundary | Before | After |
| --- | --- | --- |
| Gutter raster and background paint | Allocate and visit document-height rows | Allocate and visit destination-intersecting rows |
| Visual-row source metadata | Copy complete mappings | Read only the requested native source-ID range |
| Styled highlight placement | Visit every logical line for every chunk | Binary-search the first overlapping line and visit only the affected interval |

```mermaid
flowchart LR
  D[Complete document and syntax chunks] --> L[Native layout remains complete]
  L --> V[Destination-row paint window]
  D --> S[Seek first overlapping logical line]
  S --> H[Apply styles only to intersecting lines]
```

The bounded source read reuses the existing logical-line-info ABI. Bun paint consumers separately truncate native destination coordinates and target scroll offsets before deriving source rows, matching native integer-cell behavior while leaving public scrolling values and strict range validation unchanged.

Styled chunk offsets continue to use newline-excluded display-cell coordinates. The native lookup preserves empty-line boundaries, all four width methods, tabs, Unicode, multiline ranges, links, attributes, priorities, and highlight transactions. The FFI surface and exported native symbols are unchanged.

## Performance

| Fixture | Before | After |
| --- | ---: | ---: |
| Native styled text, 5,000 lines / 25,000 chunks | 3,686 ms | 15 ms |
| Packed Bun `CodeRenderable`, 5,000 JSON fields | 5,973 ms | 164 ms |
| Packed Bun `CodeRenderable`, 50,000 JSON fields | Not run against the quadratic baseline | 1,271 ms |
| Packed Node `CodeRenderable`, 50,000 JSON fields | Not run against the quadratic baseline | 1,406 ms |

The original production DiffViewer reproduction timed out processing input after its plaintext frame. The 50,000-field packed consumers assert 400,001 actual native highlights, exact syntax colors on early and late rows, complete text, scrolling, and narrow resizing. They do not use a plaintext fallback. Conversion, FFI packing, and native styling can still occupy the main thread for roughly one second at this size; this PR removes the quadratic traversal rather than claiming zero event-loop delay.

A follow-up autoresearch pass measured one warmup and seven runs against this exact fixture. Reusing the previous line position across ordered styled chunks reduced Bun highlight completion from a fresh 1,408 ms median to 1,271 ms, native packing/application from 774 ms to 619 ms, and the maximum heartbeat gap from 1,057 ms to 903 ms. A TypeScript style-combination cache was discarded because its apparent improvement was inside the noise floor and did not justify the complexity.

## Demo

https://github.com/user-attachments/assets/4c8bbb7f-b471-4fc0-a6ae-2055ddd462c8

Matched production OpenCode V2 `/diff` runs at 160×30 with 5,000 changed JSON fields. **Before** uses OpenCode base `0d42e76006` with published OpenTUI 0.5.9; **after** uses OpenCode integration `b2bd2150bf` with the SHA-addressed preview built from this stack (`dd9fe495`, parent `5df06ff6`). Both receive the same Ctrl+F page request two seconds after the plaintext frame appears. The request blocks for 9,962 ms before and 4 ms after.

Playback is unaccelerated and sequential to keep 160-column text legible. The faster segment holds its completed state so both sections have equal duration. Providers are simulated, but this flow makes no model request. A separate 50,000-field after run completes 20 page requests in 230 ms total, reaches the real tail before and after narrow resizing, reviews and reopens the file, and finalizes its recording successfully.

## Scope

This PR owns core gutter/background paint windowing, bounded source-row access, native styled-highlight line lookup, and their Bun, Node, and native regressions. It does not virtualize document content or change the native ABI. The production implementation is approximately 160 additions and 118 deletions; most of the 1,294 additions are regression tests, benchmarks, licenses, and a pinned 5.6 KB JSON parser fixture used only by tests.

Split-diff pane-width reflow is intentionally isolated in stacked child PR #1463.

## Verification

Checks below ran sequentially on final revision `5df06ff6` with Bun 1.3.14, Node 26.5.0, and Zig 0.16.0 on macOS arm64.

```sh
# packages/core
bun run test:js
bun run test:js:node
bun run test:native
bun run typecheck
bun run build:lib
bun run test:dist --skip-build --keep-temp

# repository root
bun run build
bun run fmt:check
bun run lint

# final metadata
git diff --check upstream/main...HEAD
```

| Check | Result |
| --- | --- |
| Full native suite | 2,092 passed, 31 skipped |
| Full Core Bun suite | 5,499 passed, 23 skipped |
| Full configured Node suite | 4,756 passed, 6 skipped |
| Core typecheck and library build | Passed |
| Clean packed Bun and Node consumers | Passed |
| Root build, format, and lint | Passed |
| Native exports | All 1,852 symbols identical before and after |
| All-platform preview | Eight native targets built and published from `dd9fe495` |
| OpenCode TUI consumer suite | 905 passed, 4 skipped; TUI and CLI typechecks passed |
| Production OpenCode Drive | Matched 5,000-field A/B and 50,000-field stress run passed |
| Final whitespace check | Passed |


kommander:
Why are there even layout lines?

PR #1463, opened 2026-08-28
fix(core): realign split diffs after resizing

kitlangton:
## Why

After a wrapped diff switches from split to unified and back, scrolling to the end can show blank rows instead of the removed and added line tails. Split alignment can use the previous unified pane width, while the outer-width tracker skips reflow when the viewer returns to its original split size.

## What Changes

**Before:** In the 20,008-character regression fixture, alignment sees stale Code pane widths of 69 and 51 cells. That produces 290 and 393 wrapped rows and inserts 103 unnecessary padding rows. Once layout restores both panes to 51 cells, those padding rows remain, extending the scrollable area past the actual text.

**After:** Alignment records the actual left/right Code width pair and schedules a coalesced rebuild when either pane reports changed line information at a different width. The final split view has 393 wrapped rows on each side, with both `_OLD_END` and `_NEW_END` visible at the bottom.

```mermaid
flowchart LR
  A[Pane layout changes] --> B[Code line-info-change]
  B --> C{Actual pane width changed?}
  C -- yes --> D[Coalesced alignment rebuild]
  D --> E[Equal wrapped-row geometry]
  C -- no --> F[Keep current alignment]
```

The viewer's width-only gutter reservation at line-sign key `-1` survives alignment rebuilds. Regressions cover asymmetric character and word wrapping, view switches without an outer-width change, line-number visibility changes, hunk offsets, complete source text, selection, and real syntax highlighting.

## Demo

The deterministic OpenCode Drive comparison is complete locally: the base loses the real end after narrowing and restoring the viewer, while this branch keeps it accessible. The recording will be attached with the final consumer demo after the stack's all-platform preview is available.

## Dependency

This PR is stacked on parent PR #1462 (`diff-paint-window`), not directly on `main`. The parent repairs offscreen painting that otherwise sends negative scrolled Y coordinates to Node's unsigned FFI binding. That repair is necessary for these production-default scroll-to-bottom fixtures to run under Node without weakening or replacing the fixture.

Review this PR against `diff-paint-window`. After the parent merges, this PR can be retargeted to `main` with no source changes.

## Scope

The child diff is exactly four files under `packages/core`: 17 additions and 13 deletions in `Diff.ts`, plus 265 lines of regression registration and coverage. It changes when wrapped-row alignment is recomputed, not source text, syntax, native wrapping, or the renderer fixes inherited from its parent.

## Verification

Checks ran sequentially on the patch-identical source revision before the final rebase. `git range-diff` confirms all three child commits are unchanged.

```sh
# packages/core
bun run test:js src/renderables/Diff.resize-wrap.test.ts
bun run test:js src/renderables/Diff
bun typecheck
bunx tsc --noEmit -p tsconfig.node-test.json
bun run build:lib
bun run test:js:node

# final metadata
git range-diff 0766bc18..15f93eec diff-paint-window..HEAD
git diff --check diff-paint-window...HEAD
```

| Check | Result |
| --- | --- |
| Focused resize suite | 3 passed, 258 assertions |
| Source Diff suites | 87 passed, 31 snapshots, 605 assertions |
| Production and Node-test typechecks | Passed |
| Library build | Passed |
| Standard Node runner | 4,758 passed, 6 skipped, 0 failed |
| Final child range-diff and whitespace check | Patch-identical; passed |

### DIFF
--- a/packages/core/src/renderables/Diff.ts
+++ b/packages/core/src/renderables/Diff.ts
@@ -99,7 +99,7 @@ export class DiffRenderable extends Renderable {
   private rightCodeRenderable: CodeRenderable | null = null
 
   private pendingRebuild: boolean = false
-  private _lastWidth: number = 0
+  private _lastCodeWidths: [number, number] = [0, 0]
 
   private errorTextRenderable: TextRenderable | null = null
   private errorCodeRenderable: CodeRenderable | null = null
@@ -218,17 +218,6 @@ export class DiffRenderable extends Renderable {
     return false
   }
 
-  protected override onResize(width: number, height: number): void {
-    super.onResize(width, height)
-
-    if (this._view === "split" && this._wrapMode !== "none" && this._wrapMode !== undefined) {
-      if (this._lastWidth !== width) {
-        this._lastWidth = width
-        this.requestRebuild()
-      }
-    }
-  }
-
   private requestRebuild(): void {
     if (this.pendingRebuild) {
       return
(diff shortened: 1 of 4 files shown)


packages/core/src/renderables/Diff.ts
```diff
@@ -747,6 +750,7 @@ export class DiffRenderable extends Renderable {
     }
 
     const shouldDoAlignment = canDoWrapAlignment && !highlightingInProgress
+    this._lastCodeWidths = [leftCodeRenderable.width, rightCodeRenderable.width]
```
kommander: **Performance:** A width-driven rebuild can enqueue a second, redundant full rebuild here. Before this assignment, `createOrUpdateCodeRenderable()` replaces aligned/padded content with the unpadded content; `CodeRenderable.content` synchronously emits `line-info-change`. The handler therefore still compares the new pane widths with the old `_lastCodeWidths` and calls `requestRebuild()` again. I reproduced this against the exact parent with an asymmetric 20k/12k-character split diff while alternating 120/160 columns: the parent ran `buildView()` once per resize, while this branch ran it twice on every resize across repeated alternating runs (the extra build took ~0.23–0.80 ms here). Please avoid scheduling from the rebuild’s own intermediate content update.

PR #1505, opened 2026-09-15
fix(native): render OpenCode V2 images in cmux

### LINKED ISSUE #1506: OpenCode V2 image previews show gray boxes in cmux
## Symptom

OpenCode V2 image previews render as solid gray boxes in cmux. The same dragged screenshot renders correctly when OpenCode runs in standalone Ghostty.

This affects both prompt attachments and submitted-message previews, so enabling `prompt.image_preview` and `session.image_preview` does not fix it.

## Environment

- OpenCode V2 2.0.3
- OpenTUI 0.5.10 (also present on current 0.5.11)
- cmux 0.64.23
- no tmux or zellij layer

## Reproduction

1. Run OpenCode V2 in cmux.
2. Enable prompt and session image previews.
3. Drag a PNG screenshot into the prompt and submit it.
4. Observe a correctly sized gray rectangle instead of the image.
5. Run the same OpenCode build in Ghostty; the image renders.

A raw Kitty protocol probe narrows the failure:

- combined transmit-and-display (`a=T`) renders
- separate transmit and placement (`a=t`, then `a=p`) at OpenTUI's current z-index does not render
- the same separate placement at `z=-1` renders

## Cause

OpenTUI emits Kitty placements at approximately `z=-1500000000 + placement_id`. The Kitty graphics protocol reserves z-indices below `INT32_MIN / 2` (-1,073,741,824) for images below cells with non-default background colors.

OpenCode's image preview container paints such a background, so the image is correctly placed underneath it and only the gray container remains visible.

## Expected behavior

Inline images should remain below terminal text but above non-default cell backgrounds. A z-index in Kitty's normal negative range preserves text overlay without hiding the image behind its container.

kunickiaj:
Fixes #1506.

## Symptom

OpenCode V2 image previews show gray boxes in cmux. The boxes have the correct size, but the image is hidden. The same dragged screenshot renders in standalone Ghostty, and cmux renders Kitty images emitted by `chafa`.

## Cause

OpenTUI places inline images at approximately `z=-1500000000 + placement_id`. Kitty reserves values below `INT32_MIN / 2` for images below cells with non-default backgrounds. OpenCode's preview container paints such a background, which covers the image.

A protocol probe confirms the boundary:

- combined `a=T` renders
- separate `a=t` + `a=p` at OpenTUI's current z-index is hidden
- the same separate placement at `z=-1` renders

## Fix

Move live and scrollback Kitty placements into the normal negative z-index range. Images remain below text but above non-default cell backgrounds, and placement ordering is preserved across OpenTUI's complete placement-ID range.

## Test coverage

The first commit updates renderer-output assertions for both live and scrollback image placements. The second commit implements the corrected z-index policy.

The focused native tests pass, including a red-before/green-after regression check. The full native suite passes with 2,146 tests passed and 32 skipped. The root build, formatting check, and lint also pass. OpenCode 2.0.3 built against the patched native library renders an image preview correctly in cmux.

## Before / After
before-after


simonklee:
Thanks for the pr. Superseded by https://github.com/anomalyco/opentui/commit/4b1474e009bf78365a7a4866e83a9e522533ef42

